PR #2, opened 2019-12-12
UI.Importance for output-relevant value list parameters

ralfhandl:


BerSie:
Hi Ralf,
there is nothing wrong with it. But was the requirement to add the dependency in a Common vocabulary to a UI vocabulary? If it's about table rendering there was another proposal to use the presentationVariantQualifier which point to a LineItem annotations, which can have Importance on DataField level.

ralfhandl:
Hi Bernhard,

This is about type-ahead/autocomplete (https://experience.sap.com/fiori-design-web/input-field/#responsiveness-and-adaptiveness), to only show the highly important fields in the autocomplete dropdown. The presentationVariantQualifier would then be used in the value help dialog.

BerSie:
OK - got it.

PR #3, opened 2019-12-12
Recommended length of text field

ralfhandl:


erlethor:
This one was replaced by https://github.com/SAP/odata-vocabularies/pull/28
@BerSie : It would be nice if you could confirm?

BerSie:
Yes, exactly - we created the HTML5 voc with CssDefaults term.

PR #6, opened 2019-12-16
Improved descriptions

ralfhandl:


ralfhandl:
We define it in [7.2.7 Default Value](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_DefaultValue), referencing the [ABNF rule `primitiveValue`](https://github.com/oasis-tcs/odata-abnf/blob/master/abnf/odata-abnf-construction-rules.txt#L902).

Would it suffice to add these references to the description?

BerSie:
> We define it in [7.2.7 Default Value](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_DefaultValue), referencing the [ABNF rule `primitiveValue`](https://github.com/oasis-tcs/odata-abnf/blob/master/abnf/odata-abnf-construction-rules.txt#L902).
> 
> Would it suffice to add these references to the description?

Yes, that would be good.

PR #17, opened 2020-03-05
Common: new term WhyNotAvailable 

ralfhandl:
that can be applied to annotations with Core.OperationAvailable


ralfhandl:
Is this PR still relevant?

HeikoTheissen (commented):
Cannot say whether this is still relevant.

RaizoC:
No, we should close this one unless we have a valid prototype. I'd rather transport this information similar to any other messages bound to the related control property.

PR #18, opened 2020-03-05
ODM: terms for One Domain Model

ralfhandl:


vocabularies/ODM.xml
```diff
@@ -0,0 +1,43 @@
...
+            <PropertyValue Property="href" String="https://sap.github.io/odata-vocabularies/vocabularies/ODM.json" />
+          </Record>
+          <Record>
+            <PropertyValue Property="rel" String="describedby" />
+            <PropertyValue Property="href" String="https://github.com/sap/odata-vocabularies/blob/master/vocabularies/ODM.md" />
+          </Record>
+        </Collection>
+      </Annotation>
+
+      <Term Name="codeList" Type="Core.Tag" AppliesTo="EntityType EntitySet">
```
GeraldKrause: In other vocabs we apply UpperCamelCase. Why not here?

ralfhandl: Data Model Alignment is using lowerCamelCase for its annotations. I'd like to use identical annotation names in CDS sources and in OData $metadata.

PR #27, opened 2020-04-24
UI.Placeholder to support user input

BerSie:
new term "UI.Placeholder" to support user input by giving an example or hint.

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1253,7 +1253,7 @@ The trend is
         <Annotation Term="Core.RequiresType" String="Edm.String" />
       </Term>
 
-      <Term Name="Placeholder" Type="Edm.String" AppliesTo="Property">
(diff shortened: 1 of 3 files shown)


vocabularies/UI.xml
```diff
@@ -1253,6 +1253,11 @@ The trend is
         <Annotation Term="Core.RequiresType" String="Edm.String" />
       </Term>
 
+      <Term Name="Placeholder" Type="Edm.String" AppliesTo="Property">
```
ralfhandl, line 1256:
```suggestion
      <Term Name="Placeholder" Type="Edm.String" Nullable="false" AppliesTo="Property">
```

BerSie:
I took the change, but I had to run build locally to update UI.json as well. Now the representations are in sync.

ralfhandl:
> I had to run build locally to update UI.json as well

Yes, the automatic build step only checks whether there are differences, it doesn't commit differing JSON files to the repo. Basically because I don't know how to do this 😄

PR #36, opened 2020-06-29
Soft Link to SAP Object Node Type

ralfhandl:
- mark an entity type as a representation of an SAP Object Node Type
- mark a property as a reference to an SAP Object Node Type

To discuss
- [x] new vocabulary or part of Common?
- [x] two terms - one for each purpose?
- [x] Rename back to "Node" Type because node types also will become entity types


RaizoC (commented):
I doubt that we should introduce a VDM-specific vocabulary, but rather align with existing or similar concepts, e.g. from ODM.

vocabularies/Common.xml
```diff
@@ -553,34 +553,6 @@ The minimum structure is
         </Annotation>
       </Term>
 
-      <Term Name="WeakReferentialConstraint" Type="Common.WeakReferentialConstraintType" Nullable="false" AppliesTo="EntitySet">
```
HeikoTheissen, line 556: Why is this term deleted in this PR?

examples/VDM.SAPObjectNodeType-sample.xml
```diff
@@ -0,0 +1,105 @@
...
+
+      <Annotations Target="this.A_AnyEntity/Region">
+        <Annotation Term="VDM.SAPObjectNodeType">
+          <Record>
+            <PropertyValue Property="Name" String="Region" />
+          </Record>
+        </Annotation>
+      </Annotations>
+
+      <Annotations Target="this.A_AnyEntity/CountryThreeLetterISOCode">
```
HeikoTheissen, line 95: Is there a deeper reason why the analogous property in the other entity types is not annotated in the same way?

vocabularies/VDM.xml
```diff
@@ -0,0 +1,55 @@
...
+        </Collection>
+      </Annotation>
+
+      <Term Name="SAPObjectNodeType" Type="VDM.SAPObjectNodeTypeType" Nullable="false" AppliesTo="EntityType Property">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The SAP Object Node Type represented by the annotated entity type, or referenced by the annotated property." />
+        <Annotation Term="Core.Description">
+          <String>SAP Object Node Types describe the structure of SAP Object Types.
+            
+SAP Object Types are a generaliation of Business Object, Technical Object, Configuration Object, and Analytical Object.</String>
```
HeikoTheissen, line 41:
```suggestion
SAP Object Types are a generalization of Business Object, Technical Object, Configuration Object, and Analytical Object.</String>
```

HeikoTheissen (changes requested):

vocabularies/Common.xml
ralfhandl: It was an early attempt at solving the soft-link problem, which is now replaced with the SONT-reference

vocabularies/VDM.xml
```diff
@@ -0,0 +1,55 @@
...
+            <PropertyValue Property="href" String="https://sap.github.io/odata-vocabularies/vocabularies/VDM.json" />
+          </Record>
+          <Record>
+            <PropertyValue Property="rel" String="describedby" />
+            <PropertyValue Property="href" String="https://github.com/sap/odata-vocabularies/blob/master/vocabularies/VDM.md" />
+          </Record>
+        </Collection>
+      </Annotation>
+
+      <Term Name="SAPObjectNodeType" Type="VDM.SAPObjectNodeTypeType" Nullable="false" AppliesTo="EntityType Property">
```
HeikoTheissen, line 35: I find the term "object node type" misleading if this is applied to a property.

HeikoTheissen:
> I doubt that we should introduce a VDM-specific vocabulary, but rather align with existing or similar concepts, e.g. from ODM.

What about re-using `Core.Links` with a new `rel="SAPObjectNodeType"`?

ralfhandl:
> What about re-using `Core.Links` with a new `rel="SAPObjectNodeType"`

That would require having URLs pointing to a SOT/SONT.

HeikoTheissen:
> That would require having URLs pointing to a SOT/SONT.

e.g., `urn:com.sap.objecttype:SalesOrder`

ralfhandl:
> e.g., `urn:com.sap.objecttype:SalesOrder`

Not easily done according to https://datatracker.ietf.org/doc/html/rfc8141#section-1

> While other URI schemes may allow resource identifiers to be freely
   chosen and assigned, such is not the case for URNs.  The syntactical
   correctness of a name starting with "urn:" is not sufficient to make
   it a URN.  In order for the name to be a valid URN, the namespace
   identifier (NID) needs to be registered in accordance with the rules
   defined here, and the remaining parts of the assigned-name portion of
   the URN need to be generated in accordance with the rules for the
   registered URN namespace.

Plus we would need to generalize Core.Links to allow URIs in addition to URLs - which is easier than getting a URN namespace 😄

HeikoTheissen:
I withdraw my idea to use Core.Links

ralfhandl:
> I withdraw my idea to use Core.Links

I really liked the idea, which is why I looked into the URN stuff again.

HeikoTheissen:
> I really liked the idea, which is why I looked into the URN stuff again.

But if Core.Links demands a URL today, isn't the expectation that you can navigate there using an HTTP client? That would be impossible with `urn:com.sap.objecttype:SalesOrder`, because there is no such thing as a URN client.

ralfhandl:
> isn't the expectation that you can navigate there using an HTTP client?

Correct, navigable hyper-link.

The "soft links" are more like tags that you can use to search on e.g. api.sap.com for an entity with this tag.

vocabularies/VDM.xml
```diff
@@ -0,0 +1,55 @@
...
+          <String>SAP Object Node Types describe the structure of SAP Object Types.
+            
+SAP Object Types are a generalization of Business Object, Technical Object, Configuration Object, and Analytical Object.</String>
+        </Annotation>
+      </Term>
+
+      <ComplexType Name="SAPObjectNodeTypeType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Information about an SAP Object Node Type" />
+        <Property Name="Name" Type="Edm.String" Nullable="false">
```
HeikoTheissen, line 48: Is there a namespace concept for SAP object node types? For example, if a partner introduces new types, must their have their own namespace? Perhaps this should be a QualifiedName rather than just an Edm.String.

vocabularies/VDM.xml
ralfhandl: Done

vocabularies/VDM.xml
ralfhandl: This question goes to the SOT/SONT experts

vocabularies/Common.xml
```diff
@@ -553,6 +553,27 @@ The minimum structure is
...
 
+      <Term Name="SAPObjectType" Type="Common.SAPObjectTypeType" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The SAP Object Type represented by the annotated entity type" />
+        <Annotation Term="Core.LongDescription">
+          <String>SAP Object Types are a generalization of Business Object, Technical Object, Configuration Object, and Analytical Object.</String>
+        </Annotation>
+      </Term>
+
+      <ComplexType Name="SAPObjectTypeType">
```
HeikoTheissen, line 564: Why is this a complex type? Do we envisage more properties in the future?

vocabularies/Common.xml
```diff
@@ -553,34 +553,27 @@ The minimum structure is
...
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Information about an SAP Object Type" />
+        <Property Name="Name" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The name of the SAP Object Type" />
         </Property>
       </ComplexType>
 
+      <Term Name="SAPObjectTypeReference" Type="Edm.String" Nullable="false" AppliesTo="Property">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The SAP Object Type referenced by the annotated property" />
```
HeikoTheissen, line 574:
```suggestion
        <Annotation Term="Core.Description" String="The name of the SAP Object Type referenced by the annotated property" />
```
since `SAPObjectType` is a complex type, not just a name.

vocabularies/Common.xml
ralfhandl: Yes, whatever is useful

HeikoTheissen (dismissed):

RaizoC (approved):
the sample still has SAPObjectNodeType in, further a typo "SAPOBjecType" with missing "t".

HeikoTheissen (dismissed):

RaizoC (dismissed):

NilsObermiller-SAP:
I like the idea of having a navigable link. Initially (years back) the idea was to navigate to an OData service which represents the SO(N)T. This changed meanwhile and the "soft link" is rather a tag. But what about a link to the API Hub, example https://api.sap.com/sap-object-type/SalesOrderType. There all OData Services, CDS Views etc. which **represent** this SO(N)T should be listed.

HeikoTheissen:
Do we need a new `rel` for that? Must its value come from https://html.spec.whatwg.org/#linkTypes?
```json
"@Core.Links": [{"rel": "What would this be?", "href": "https://api.sap.com/sap-object-type/SalesOrderType"}]
```

ralfhandl:
Let's worry about that once this kind of link returns something other than `404 Not Found` 😄

vocabularies/Common.xml
```diff
@@ -553,34 +553,27 @@ The minimum structure is
...
+        <Annotation Term="Core.LongDescription">
+          <String>SAP Object Types are a generalization of Business Object, Technical Object, Configuration Object, and Analytical Object.</String>
+        </Annotation>
+      </Term>
+
+      <ComplexType Name="SAPObjectTypeType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Information about an SAP Object Type" />
+        <Property Name="Name" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The name of the SAP Object Type" />
```
HeikoTheissen, line 568:
```suggestion
        <Property Name="Name" Type="Edm.String" Nullable="false">
          <Annotation Term="Core.Description" String="The name of the SAP Object Type" />
        </Property>
        <Property Name="IsSubnode" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
          <Annotation Term="Core.Description" String="Whether the name describes a subnode of an SAP Object Type" />
```
Would this suffice to support SAPObjectNodeTypes?

vocabularies/Common.xml
HeikoTheissen: Do you envisage properties of the SAPObjectType? Then you would have to specify them consistently in every occurrence of `Common.SAPObjectType`. Or what kind of property?

ralfhandl: The `SAPObjectTypeReference` only names the object type referenced by a property.

The `SAPObjectType` marks an entity type as a representation of an SAP Object Type, and this may have additional useful information, for example the category.

Once we have a publicly available repository of SONTs, we would just add a link to it instead of copying a snapshot of the publicly available information.

HeikoTheissen: What if one entity type is annotated
```json
"@Common.SAPObjectType": {"Name": "SalesOrder", "Category": "internal"}
```
and another entity type (in a different service, say) as
```json
"@Common.SAPObjectType": {"Name": "SalesOrder", "Category": "external"}
```
? What's then the category of that SAPObjectType?

ralfhandl: Since this information is to be pulled automatically from the non-public SONT repository, the newer file has the newer content.

HeikoTheissen: We could solve this problem if we define SAPObjectTypes in a separate CSDL document.
```json
{"com.sap.objecttypes.SalesAndDistribution": {
 "$Alias": "SD_SOT",
 "@Common.SAPObjectType#SalesOrder": {"Name": "SalesOrder", "Category": "internal"}
}}
```
Here, the annotation applies to the schema. Then we annotate the entity type by referencing that annotation:
```json
"@Common.SAPObjectType": {"$Path": "SD_SOT@Common.SAPObjectType#SalesOrder"}
```

ralfhandl: Let's have a meeting, I think this discussion is heading in the wrong direction, far away from where we wanted to go.

NilsObermiller-SAP:
"Node" required for UBM (User Behaviour Mining) use case.

HeikoTheissen:
> "Node" required for UBM (User Behaviour Mining) use case.

See my suggestion in #119.

HeikoTheissen (approved):

HeikoTheissen (dismissed):

HeikoTheissen (dismissed):
No plans to enrich an `SAPObjectNodeTypeReference` with additional attributes, except for a key binding, which will be expressed in an additional annotation `SAPObjectNodeTypeBinding` or similar.

vocabularies/Common.xml
```diff
@@ -553,34 +553,27 @@ The minimum structure is
...
-        <Property Name="Property" Type="Edm.PropertyPath" Nullable="false">
-          <Annotation Term="Core.Description" String="Path to property of the annotated entity set" />
-        </Property>
-        <Property Name="ReferencedProperty" Type="Edm.String" Nullable="false">
-          <Annotation Term="Core.Description" String="Path to property in reference set, format is identical to PropertyPath annotations" />
+      <Term Name="SAPObjectNodeType" Type="Common.SAPObjectNodeTypeType" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The SAP Object Node Type represented by the annotated entity type" />
+        <Annotation Term="Core.LongDescription">
+          <String>SAP Object Node Types are a generalization of Business Object, Technical Object, Configuration Object, and Analytical Object.</String>
```
ralfhandl, line 560:
```suggestion
          <String>SAP Object Node Types define the structure of SAP Object Types, which are a generalization of Business Object, Technical Object, Configuration Object, and Analytical Object.</String>
```

HeikoTheissen (approved):

HeikoTheissen (approved):

BerSie:
Is this a more generalized SemanticObject approach?

RaizoC (dismissed):

RaizoC:
Title of pull request does not fit content. I wonder if we also need the SAPObjectType in addition or where we get the link from.

ralfhandl:
> Title of pull request does not fit content.

Fixed

> I wonder if we also need the SAPObjectType in addition or where we get the link from.

We "know" that for now we only get root node types, and that the root node type has the same name as the object type.

ralfhandl:
> Is this a more generalized SemanticObject approach?

Apparently Semantic Objects are a UI concept, and SAP Object (Node) Types are a domain-modeling concept, with only accidental overlap.

HeikoTheissen (approved):

BerSie (approved):

PR #38, opened 2020-07-08
Update nodejs.yml

ralfhandl:
Auto-refresh *.md and *.json files on push.

With this change it is no longer necessary to locally `npm run build` to refresh the derived files, this happens on push to a branch, including `master` when a pull request is merged.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -24,9 +24,10 @@ jobs:
         npm install
         npm run build --if-present
         git config user.name ${GITHUB_ACTOR}
-        git config user.email ${GITHUB_ACTOR}
+        git config user.email ${PUSHER_EMAIL}
         git add vocabularies/*.md vocabularies/*.json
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -21,6 +23,10 @@ jobs:
       run: |
         npm install
         npm run build --if-present
-        git diff --exit-code
+        git config user.name ${GITHUB_ACTOR}
+        git config user.email ${GITHUB_ACTOR}
```
HeikoTheissen, line 27: Can't you give the email address here?

ralfhandl: I first have to figure out how to get at it. Which I will, and we can use it as is for now.

ralfhandl: And as I have a "how to create a vocab PR" session with Renzo in an hour, I'd rather have this merged before that.

ralfhandl: Fixed

PR #40, opened 2020-07-10
Link to annotation cheat sheet

ralfhandl:


ralfhandl:
> Why not writing "Edm.EnumType"?

Because there is no such type, enumeration types are top-level "user-defined" types.

PR #41, opened 2020-07-10
Direct Edit vocabulary

ralfhandl:
Terms for direct-edit UIs, currently only `SideEffects`.


vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,73 @@
...
+        <Annotation Term="Core.Description" String="Determine side effects of client-side data modification" />
+      </Term>
+
+      <ComplexType Name="SideEffectsType">
+        <Annotation Term="Core.LongDescription">
+          <String>A direct-edit side-effect consists of:
+
+- a list of possible triggers, _preliminary_ syntax:
+  - `Path:` followed by a path to the property whose change triggered the side-effects determination 
+  - `Action:` followed by the name of the (UI) action that triggered the side-effects determination
```
bernhardthimmel, line 46:
```suggestion
  - `Path:` followed by a path to the property whose change triggers the side-effects determination 
  - `Action:` followed by the name of the action that is called for the side-effects determination
```

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,73 @@
...
+
+- an action bound to the entity set with the following non-binding parameters:
+  - `Qualifier` of type [`Core.SimpleIdentifier`](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#SimpleIdentifier)
+    or cast-compatible: the qualifier of the `SideEffects` annotation
+  - `Trigger` of type `Edm.String`: the trigger of side-effects determination, see `Triggers` property
+  - `Data` of either the entity type of the annotated entity set or a complex type that is structure-compatible with it, 
+  i.e.
+    - all primitive properties are cast-compatible with corresponding properties of the entity type and are nullable,
```
bernhardthimmel, line 54:
```suggestion
    - each primitive property that has the same name as a corresponding primitive property of the entity type of the annotated entity set is cast-compatible with the corresponding property and is nullable,
```

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,73 @@
...
+    - all complex properties are structure-compatible with corresponding complex or navigation properties
+      (for related entities necessary for side-effect determination)
```
bernhardthimmel, line 56:
```suggestion
    - each complex property that has the same name as a corresponding complex or navigation property of the entity type of the annotated entity set is structure-compatible with the corresponding property
```
Is the recursive definition of 'structure-compatible' here ok?

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,73 @@
...
+            <PropertyValue Property="href" String="https://sap.github.io/odata-vocabularies/vocabularies/DirectEdit.json" />
+          </Record>
+          <Record>
+            <PropertyValue Property="rel" String="describedby" />
+            <PropertyValue Property="href" String="https://github.com/sap/odata-vocabularies/blob/master/vocabularies/DirectEdit.md" />
+          </Record>
+        </Collection>
+      </Annotation>
+
+      <Term Name="SideEffects" Type="DirectEdit.SideEffectsType" Nullable="false" AppliesTo="EntitySet">
```
bernhardthimmel, line 35: Since Common.SideEffects can be applied to an entity type as well, should we do this here, too?

bernhardthimmel (changes requested):

vocabularies/DirectEdit.xml
ralfhandl: No, not without urgent need. We prefer annotations on resources (entity sets, singletons) over annotations on types.

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,72 @@
...
+  i.e.
+    - each primitive property that has the same name as a corresponding primitive property of the entity type of the annotated entity set is cast-compatible with the corresponding property and is nullable,
+    - each complex property that has the same name as a corresponding complex or navigation property of the entity type of the annotated entity set is structure-compatible with the corresponding property
+    - it may contain properties without a corresponding property in the entity type
+    - it may omit properties of the entity type
+  
+  The return type of the action also needs to be either the entity type of the annotated entity set or structure-compatible with it,
+  it can be the same type as for `Data`.</String>
+        </Annotation>
+      <Property Name="Triggers" Type="Collection(Edm.String)" Nullable="false">
```
bernhardthimmel, line 62: Is there a way to say that the collection must not be empty?

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,72 @@
...
+          <Annotation Term="Core.Description" String="List of possible triggers for determining side effects" />
```
ralfhandl, line 63:
```suggestion
          <Annotation Term="Core.Description" String="List of possible triggers for determining side effects" />
          <Annotation Term="Validation.MinItems" Int="1" />
```

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.01">
+  <edmx:Reference Uri="https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.xml">
+    <edmx:Include Alias="Core" Namespace="Org.OData.Core.V1" />
+  </edmx:Reference>
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/Common.xml">
```
ralfhandl, line 6:
```suggestion
  <edmx:Reference Uri="https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Validation.V1.xml">
    <edmx:Include Alias="Validation" Namespace="Org.OData.Validation.V1" />
  </edmx:Reference>
  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/Common.xml">
```

bernhardthimmel (approved):

HeikoTheissen (commented):
How does this relate to `Common.SideEffects`?

bernhardthimmel:
Relationship to Common.SideEffects:
Common.SideEffects relies on a backend draft to calculate side effects. This means the backend either automatically calculates side effects after patch requests and the frontend only needs to know which properties and entities to reload after a patch or the frontend needs to call an action to trigger these calculations. In all cases, the backend already has all data of the involved entities.
In our case without a draft, the state is kept in the client. This means, in order to calculate side effects, the frontend needs to send all (potentially changed) data of this client state to the backend.
Since the backend then is able to return exactly the changed information, the frontend does not necessarily need to know upfront (through annotations) which properties might change.

HeikoTheissen:
> In our case without a draft, the state is kept in the client. This means, in order to calculate side effects, the frontend needs to send all (potentially changed) data of this client state to the backend.

How does the `DirectEdit.SideEffects` annotation help with that? (Sorry, but I miss the big picture.)

bernhardthimmel:
The annotation tells the client after changes of which fields, which backend action has to be called.
So if one of the fields in _Triggers_ is changed, the UI will call the _CalculationAction_ with the current state and it will get the new state back.
So e.g. if the customer has configured an onChange rule on field Position in JobInformation, the backend would generate the annotation in the metadata with Position as the _Trigger_.
I can also send you a slide deck.

HeikoTheissen:
> The annotation tells the client after changes of which fields, which backend action has to be called.
So e.g. if the customer has configured an onChange rule on field Position in JobInformation, the backend would generate the annotation in the metadata with Position as the _Trigger_.

But if the backend already knows what action to perform _after_ patching `JobInformation/Position`, it could also perform that action _as part of_ the PATCH request, right? The `DirectEdit.SideEffects` annotation would be useful only if it is contributed by a party different from the backend.

bernhardthimmel:
But the crucial point is that the side effects should be executed before the SAVE. The patch request will only come with the SAVE. The user is meant to see the side effects (e.g. an attachment upload becomes available after setting the time type to 'sickness' or the job title is copied over from the position) and react to them before saving.
There can be additional side effects after saving but this feature here focuses on the support of onChange rules during editing.

HeikoTheissen:
> But the crucial point is that the side effects should be executed before the SAVE. The patch request will only come with the SAVE.

But then the `CalculationAction` is really a `CalculationFunction`, because it simulates only and changes nothing on the server, right?

ralfhandl:
Technically it is an action because it needs to send a lot of data to the backend, and this is better done in the POST request body of an action call than squeezed into the URL of a function call - which would then have to be wrapped in a batch request to avoid technical limits for URL length.

bernhardthimmel:
> But then the `CalculationAction` is really a `CalculationFunction`,

Yes, this is correct; well-spotted. We need to define it as action so we can pass the state of the object but it does not change the backend state and should therefore be called `CalculationFunction`.

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,77 @@
...
+      <ComplexType Name="SideEffectsType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.LongDescription">
+          <String>A direct-edit side-effect consists of:
+
+- a list of possible triggers, _preliminary_ syntax:
+  - `Path:` followed by a path to the property whose change triggers the side-effects determination 
+  - `Action:` followed by the name of the action that is called for the side-effects determination
+
+- an action bound to the entity set with the following non-binding parameters:
```
HeikoTheissen, line 52: Isn't it sufficient if every entity set has _one_ action `SideEffects` whose single parameter `Data` and return type both have the given entity type (or something structure-compatible)? For example,
```
POST ~/TimeSheet/SideEffects
Content-Type: application/json

{"Data": {"TimeType": "sickness"}}
```
would return
```json
{"@odata.context": "$metadata#TimeSheet",
 "value": {"TimeType": "sickness", "UploadEnabled": true}}
```
Then `DirectEdit.SideEffects` would be unnecessary and the long list of side effects need not be passed to the client, only the server must implement it. (Or, if the client wants to avoid unnecessary requests, it only needs the list of triggers, not the calculation actions. But I think that in the end, _every_ change on the client will trigger some side effect.)

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,77 @@
...
+  - `Qualifier` of type [`Core.SimpleIdentifier`](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#SimpleIdentifier)
+    or cast-compatible: the qualifier of the `SideEffects` annotation
+  - `Trigger` of type `Edm.String`: the trigger of side-effects determination, see `Triggers` property
+  - `Data` of either the entity type of the annotated entity set or a complex type that is structure-compatible with it,
```
HeikoTheissen, line 56: Is it sufficient to pass only one entity at a time? Can't changes to _other_ entities which the client has accumulated also affect the outcome of the determination? In other words: Must `Data` not have the structure of the entire business object?

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,77 @@
...
+      </Term>
+
+      <ComplexType Name="SideEffectsType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.LongDescription">
+          <String>A direct-edit side-effect consists of:
+
+- a list of possible triggers, _preliminary_ syntax:
+  - `Path:` followed by a path to the property whose change triggers the side-effects determination 
+  - `Action:` followed by the name of the action that is called for the side-effects determination
```
bernhardthimmel, line 50:
```suggestion
  - `CalculationAction:` followed by the name of the action that is called for the side-effects determination
```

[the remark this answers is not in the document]
bernhardthimmel: I've just added a suggestion to change the name 'Action' to 'CalculationAction' (this is still pending our discussion if it should actually be called 'CalculationFunction'.
In short it is the same action.
Yes, the client holds the state and must thus accumulate all changes entered by the user and calculated by the side effects.

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,77 @@
...
+      </Term>
+
+      <ComplexType Name="SideEffectsType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.LongDescription">
+          <String>A direct-edit side-effect consists of:
+
+- a list of possible triggers, _preliminary_ syntax:
+  - `Path:` followed by a path to the property whose change triggers the side-effects determination 
+  - `CalculationAction:` followed by the name of the action that is called for the side-effects determination
```
ralfhandl, line 50: @bernhardthimmel This text is mismatched with the current definition of `Triggers` as a collection of strings. Which one needs to be fixed?

vocabularies/DirectEdit.xml
bernhardthimmel: The list of actions is not completely necessary but it gives more flexibility. 
The triggers are important. Every field change with a side effect has a performance impact so we want to avoid unnecessary round trips.
We do need to know which field triggered the change because this is not possible to determine just from the backend state (e.g. if you have a side effect when changing the position and another one when changing the division, you do not want to recalculate the position change after a subsequent division change).

vocabularies/DirectEdit.xml
bernhardthimmel: We allow passing trees. Either directly as nested data of the entity type or via a complex type with nested complex types.
We do not currently have the use case of multiple top level entities although this could in theory happen. In that case we could still extend the action signature.

vocabularies/DirectEdit.xml
bernhardthimmel: You're correct, there is a mismatch. Why did we not notice this earlier?
The actual annotations are correct and we need to fix the description here.
We are mixing the setup that the service needs to fulfill and the metadata describing it a bit.
I'll have to look at it again.

vocabularies/DirectEdit.xml
HeikoTheissen: I do not understand the semantics of a side effect
```json
{"Triggers": ["CalculationAction:self.A"], "CalculationAction": "self.A"}
```

bernhardthimmel: Let's start from the end result and see how we can tidy the explanation.
A typical side effect would be:
{"Triggers": ["position", "division"], "CalculationAction": "self.A"}

One idea to clean this up:
We could move everything that we have under "an action bound to the entity set..." into the description of the property CalculationAction.
Similarly we could enhance the description of Triggers slightly with the information that is in the bullet point `Path`.
Then we could start the description of the type similarly to Common.SideEffects by saying that "While editing a document, changes of certain fields can have side effects on other properties or entities that can only be calculated by the service. The side effects type allows to determine the consumer when such side effects are triggered and which calculation function of the service (modeled as an action) needs to be called to evaluate the side effects."
 
Would this change make sense?

vocabularies/DirectEdit.xml
HeikoTheissen: > We do need to know which field triggered the change

But that could be achieved by having _two_ parameters `Data` and `BeforeData`, which would not only show which property/ies has/ve changed, but also the previous value(s), which could also be important.

vocabularies/DirectEdit.xml
HeikoTheissen: Suggestion: After a change to a property whose path is contained in `Triggers`, the client should pass the entity with the changed property to the `CalculationFunction` and receive the entity with all changes that happen as side effects of the given change.

vocabularies/DirectEdit.xml
bernhardthimmel: At least for our current non-OData based implementation we do not really need to know the complete before image but simply send the information which field changed.
The before image would also need more work in the client and would roughly double the payload for most cases.
I do not want to discount other use cases, though. Do you see a way of adding additional variants later?

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,77 @@
...
+  - `Trigger` of type `Edm.String`: the trigger of side-effects determination, see `Triggers` property
+  - `Data` of either the entity type of the annotated entity set or a complex type that is structure-compatible with it, 
+  i.e.
+    - each primitive property that has the same name as a corresponding primitive property of the entity type of the annotated entity set is cast-compatible with the corresponding property and is nullable,
+    - each complex property that has the same name as a corresponding complex or navigation property of the entity type of the annotated entity set is structure-compatible with the corresponding property
+    - it may contain properties without a corresponding property in the entity type
+    - it may omit properties of the entity type
+  
+  The return type of the action also needs to be either the entity type of the annotated entity set or structure-compatible with it,
+  it can be the same type as for `Data`.</String>
```
bernhardthimmel, line 64:
```suggestion
          <String>After a change to a property whose path is contained in `Triggers`, the client should pass the entity with the changed property to the `CalculationFunction` and receive the entity with all changes that happen as side effects of the given change.</String>
```

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,77 @@
...
+        </Annotation>
+        <Property Name="Triggers" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.Description" String="List of possible triggers for determining side effects" />
+          <Annotation Term="Validation.MinItems" Int="1" />
+        </Property>
+        <Property Name="CalculationAction" Type="Common.QualifiedName" Nullable="false">
+          <Annotation Term="Core.Description" String="Name of the action for determining side effects" />
```
bernhardthimmel, line 71:
```suggestion
        <Property Name="CalculationFunction" Type="Common.QualifiedName" Nullable="false">
          <Annotation Term="Core.Description" String="The operation calculating the side effects. While non-changing for the service, this technically is an action bound to the entity set so that the parameters can be sent in the POST request body. The action has the following non-binding parameters:
  - `Qualifier` of type [`Core.SimpleIdentifier`](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#SimpleIdentifier)
    or cast-compatible: the qualifier of the `SideEffects` annotation
  - `Trigger` of type `Edm.String`: the trigger of side-effects determination, see `Triggers` property
  - `Data` of either the entity type of the annotated entity set or a complex type that is structure-compatible with it, 
  i.e.
    - each primitive property that has the same name as a corresponding primitive property of the entity type of the annotated entity set is cast-compatible with the corresponding property and is nullable,
    - each complex property that has the same name as a corresponding complex or navigation property of the entity type of the annotated entity set is structure-compatible with the corresponding property
    - it may contain properties without a corresponding property in the entity type
    - it may omit properties of the entity type
  
  The return type of the action also needs to be either the entity type of the annotated entity set or structure-compatible with it,
  it can be the same type as for `Data`." />
```

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,77 @@
...
+    - each primitive property that has the same name as a corresponding primitive property of the entity type of the annotated entity set is cast-compatible with the corresponding property and is nullable,
+    - each complex property that has the same name as a corresponding complex or navigation property of the entity type of the annotated entity set is structure-compatible with the corresponding property
+    - it may contain properties without a corresponding property in the entity type
+    - it may omit properties of the entity type
+  
+  The return type of the action also needs to be either the entity type of the annotated entity set or structure-compatible with it,
+  it can be the same type as for `Data`.</String>
+        </Annotation>
+        <Property Name="Triggers" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.Description" String="List of possible triggers for determining side effects" />
```
bernhardthimmel, line 67:
```suggestion
          <Annotation Term="Core.Description" String="List of paths to the properties whose changes should trigger the side effects calculation." />
```

vocabularies/DirectEdit.xml
HeikoTheissen: I want to keep it simple, not just for now but also long-term. Based on my experience with `Common.SideEffects`, I am afraid that this annotation might grow into a monster, therefore I would like to avoid it altogether.

Better to keep the complete knowledge on the server and have the client invoke an action with `Data` and `BeforeData` after every change. If your current use case doesn't care about the before-value, you can even set it equal to the after-value, the mere presence of `BeforeData/SomeProperty` would then indicate that this property triggered the action call. And if `BeforeData` contains only the `SomeProperty`, its effect on the payload is negligible.

HeikoTheissen (commented):
I am still not convinced of the necessity for this term, see https://github.com/SAP/odata-vocabularies/pull/41#discussion_r622949306.

vocabularies/DirectEdit.xml
HeikoTheissen: My proposal was to send `BeforeData` = before-image and `Data` = after-image. The `BeforeData` could be restricted to the properties that have actually changed, thus implying the trigger.

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,71 @@
...
+            <PropertyValue Property="rel" String="describedby" />
+            <PropertyValue Property="href" String="https://github.com/sap/odata-vocabularies/blob/master/vocabularies/DirectEdit.md" />
+          </Record>
+        </Collection>
+      </Annotation>
+
+      <Term Name="SideEffects" Type="DirectEdit.SideEffectsType" Nullable="false" AppliesTo="EntitySet">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Determine side effects of client-side data modification" />
+      </Term>
```
HeikoTheissen, line 41:
```suggestion
      <Term Name="SideEffects" Type="Edm.Boolean" Nullable="false" AppliesTo="Action Function">
        <Annotation Term="Common.Experimental" />
        <Annotation Term="Core.Description" String="After a change to an entity, the client should pass the before and after images to the annotated action or function and receive the entity with all changes that happen as side effects of the given change. But if this annotation evaluates to false, there are no side effects, so the action/function call can be avoided." />
      </Term>
```
Cf. the idea from my mail.

bernhardthimmel: I'll set up a call. I do not understand the suggestion, I must admit.

HeikoTheissen (commented):
The description is better formatted now.

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,79 @@
...
+          <Annotation Term="Core.Description" String="List of paths to the properties whose changes should trigger the side effects calculation." />
+          <Annotation Term="Validation.MinItems" Int="1" />
+        </Property>
+        <Property Name="CalculationFunction" Type="Common.QualifiedName" Nullable="false">
+          <Annotation Term="Core.Description" String="The operation calculating the side effects. While non-changing for the service, this technically is an action bound to the entity set so that the parameters can be sent in the POST request body. The action has the following non-binding parameters:
+
+- `Qualifier` of type [`Core.SimpleIdentifier`](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#SimpleIdentifier)
+    or cast-compatible: the qualifier of the `SideEffects` annotation
+
+- `Trigger` of type `Edm.String`: the trigger of side-effects determination, see `Triggers` property
```
bernhardthimmel, line 58:
```suggestion
- `Trigger` of type `Edm.String`: the trigger of the side-effects determination, see `Triggers` property
```

vocabularies/DirectEdit.xml
```diff
@@ -0,0 +1,79 @@
...
+
+- `Data` of either the entity type of the annotated entity set or a complex type that is structure-compatible with it
+
+The return type of the action also needs to be either the entity type of the annotated entity set or structure-compatible with it,
+it can be the same type as for `Data`.
```
bernhardthimmel, line 63: In the md file this is right in the middle of the action description since the description of 'structure compatible' is right behind it.
Is there a way to mark 'return type' more obviously so that you can immediately detect where the return signature starts (the 3 input parameters are clearly visible)? Would it make sense to move this to the end (after defining 'structure compatible') or could the definition of 'structure compatible' be separated/indented somehow to show that this is just additional information?

HeikoTheissen:
```
You can use **bold** or *italic*.
```
You can use **bold** or *italic*.

bernhardthimmel:
```suggestion
The **return type** of the action also needs to be either the entity type of the annotated entity set or structure-compatible with it,
it can be the same type as for `Data`.
```

bernhardthimmel (approved):

vocabularies/DirectEdit.json
```diff
@@ -0,0 +1,71 @@
...
+            "@Common.Experimental": true,
+            "@Core.Description": "Determine side effects of client-side data modification"
+        },
+        "SideEffectsType": {
+            "$Kind": "ComplexType",
+            "@Common.Experimental": true,
+            "@Core.LongDescription": "After a change to a property whose path is contained in `Triggers`, the client should pass the entity with the changed property to the `CalculationFunction` and receive the entity with all changes that happen as side effects of the given change.",
+            "Triggers": {
+                "$Collection": true,
+                "@Core.Description": "List of paths to the properties whose changes should trigger the side effects calculation.",
```
uhlmannm, line 62: To be on the safe side: This path can include navigation properties?

vocabularies/DirectEdit.json
```diff
@@ -0,0 +1,71 @@
...
+                "@Validation.MinItems": 1
+            },
+            "CalculationFunction": {
```
uhlmannm, line 65: In a more complex UI, the function might not only return changes to the entity itself but also for dependent entities. E.g. when called on a Sales Order also changes to items might be returned. 
The function call would then need to be done with $expand. How would that work out?

vocabularies/DirectEdit.json
bernhardthimmel: Yes, correct. It could be that a change of a property on a subnode has a side effect.

uhlmannm (approved):

PR #42, opened 2020-07-13
Define new enum member OnlyIcon for CriticalityRepresentationType

BerSie:
could be used to define UI.DataFields with icon only


vocabularies/UI.xml
```diff
@@ -1377,6 +1377,9 @@ The trend is
         <Member Name="WithoutIcon">
           <Annotation Term="Core.Description" String="Criticality is represented without icon, e.g. only via text color" />
         </Member>
+        <Member Name="OnlyIcon">
```
ralfhandl, line 1380: Initially `Experimental`?

PR #46, opened 2020-07-28
UI.ExpressionParameter

HeikoTheissen:
Cf. internal discussion on 2015-09-07

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1120,6 +1120,15 @@ The trend is
           <Annotation Term="Core.Description" String="Value for the key property" />
         </Property>
       </ComplexType>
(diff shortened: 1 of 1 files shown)


vocabularies/UI.json
```diff
@@ -1327,6 +1327,18 @@
                 "@Core.Description": "Value for the key property"
             }
         },
+        "ExpressionParameter": {
```
ralfhandl, line 1330: You don't need this, just use the concrete type [`Parameter`](https://github.com/SAP/odata-vocabularies/blob/master/vocabularies/UI.md#Parameter) and use an expression for `PropertyValue`.

That way it is clear what the expression language is: EDM annotation expressions.

PR #50, opened 2020-08-03
New terms IsISOCurrency & IsUnit

BerSie:
Using these terms it's possible to identify currencies and unit of measures if no amount or price is within the entity. And in addition it eases the determination of currency and unit properties without looking into all ISOCurrency and Unit annotations and check the path there.


vocabularies/Common.xml
```diff
@@ -155,6 +155,14 @@
         <Annotation Term="Core.RequiresType" String="Edm.String" />
       </Term>
 
+      <Term Name="IsISOCurrency" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property Parameter">
```
ralfhandl, line 158: Given the lengthy discussions around "internal" currency codes, do we want to have the infix `ISO` here?

BerSie: Done

vocabularies/Common.xml
```diff
@@ -155,6 +155,14 @@
         <Annotation Term="Core.RequiresType" String="Edm.String" />
       </Term>
 
+      <Term Name="IsCurrency" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property Parameter">
+        <Annotation Term="Core.Description" String="Annotated property or parameter is a currency" />
```
ralfhandl, line 159:
```suggestion
        <Annotation Term="Core.Description" String="Annotated property or parameter is a currency code" />

```

BerSie:
New term names are Common.IsCurrency and Common.IsUnit

PR #52, opened 2020-08-05
UI.DataFieldWithValueHelp

HeikoTheissen:


vocabularies/UI.json
```diff
@@ -1750,6 +1750,12 @@
                 "@Core.Description": "Qualified name of an Action, Function, ActionImport or FunctionImport in scope"
             }
         },
+        "DataFieldWithValueHelp": {
```
erlethor, line 1753: Since I was not involved so far I would like to give at least some feedback that I don't have a clue what this should be based on description and long description.

Typically a property has a ValueList annotation. But this seems to be a DataField
WithAction. Again I can't understand the use case based on what I see here. 

_A piece of data_ doesn't sound very nice to me

HeikoTheissen: Thinking about renaming `DataFieldWithValueHelp` to `DataFieldWithFunction`.

examples/com.sap.vocabularies.UI.v1.DataFieldDefault.xml
```diff
@@ -0,0 +1,65 @@
...
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/UI.xml">
+    <edmx:Include Namespace="com.sap.vocabularies.UI.v1" Alias="UI" />
+  </edmx:Reference>
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/Common.xml">
+    <edmx:Include Namespace="com.sap.vocabularies.Common.v1" Alias="Common" />
+  </edmx:Reference>
+  <edmx:DataServices>
+    <Schema Namespace="datafielddefault.sample" xmlns="http://docs.oasis-open.org/odata/ns/edm">
+      <Annotations Target="my.container/QueryResult/PostingDate">
+        <!-- Based on these annotations, the filter facet for the PostingDate property shall, in addition to the usual date picker,
```
erlethor: I hope I am not completely off but in Filter Facet means the Filter Bar:
Filters are based on the SelectionField annotation which doesn't allow DataFields but only PropertyPath

vocabularies/UI.json
erlethor: 👍🏻 This will help avoiding confusion with ValueList that many people call ValueHelp.

erlethor: `DataFieldWithValueFunction` In case you need another candidate

examples/com.sap.vocabularies.UI.v1.DataFieldDefault.xml
HeikoTheissen: With this pull request, I want to introduce a new type of filter: Besides the usual filters like `PostingDate eq '2020-08-05'`, which you cannot usefully bookmark, you can now set functional filters like `PostingDate eq SAP.GregorianDateFunction(P_DateFunctionID='TODAY')`, which you *can* bookmark.

examples/com.sap.vocabularies.UI.v1.DataFieldDefault-sample.xml
```diff
@@ -0,0 +1,65 @@
...
+        </Annotation>
+        <Annotation Term="UI.DataFieldDefault" Qualifier="GregorianDateFunction">
+          <Record Type="UI.DataFieldWithValueFunction">
+            <PropertyValue Property="Label" String="Gregorian date function" />
+            <PropertyValue Property="Value" Path="PostingDate" />
+            <PropertyValue Property="Action" String="SAP.GregorianDateFunction" />
+          </Record>
+        </Annotation>
+      </Annotations>
+      <Annotations Target="SAP.FiscalDateFunction/P_DateFunctionID">
```
BerSie, line 33: what is SAP. - if it's a namespace it has to be defined somewhere

examples/com.sap.vocabularies.UI.v1.DataFieldDefault-sample.xml
```diff
@@ -0,0 +1,65 @@
...
+  <edmx:DataServices>
+    <Schema Namespace="datafielddefault.sample" xmlns="http://docs.oasis-open.org/odata/ns/edm">
+      <Annotations Target="my.container/QueryResult/PostingDate">
+        <!-- Based on these annotations, the filter facet for the PostingDate property shall, in addition to the usual date picker,
+             offer two input helps
+             * one computes the PostingDate based on a fiscal date function that is chosen from the FiscalDateFunctions entity set, e.g., "current period"
+             * one computes the PostingDate based on a Gregorian date function that is chosen from the GregorianDateFunctions entity set, e.g., "today".
+             In both cases the computation happens on the server, when an OData request like in com.sap.vocabularies.UI.v1.SelectionVariant-sample.xml is executed.
+        -->
+        <Annotation Term="UI.DataFieldDefault" Qualifier="FiscalDateFunction">
```
BerSie, line 18: why do you use DataFieldDefault here? Does the user have the choice which one he wants to see in the UI, or do you want offer all in one?

HeikoTheissen: Both fiscal and Gregorian date functions shall be offered to the user (in addition to the usual date picker).

PR #53, opened 2020-08-06
Validation using value helps

BerSie:
Using the new term Common.ValueListForValidation developers might point out a specific Common.ValueList or Common.ValueListMapping annotation that can be used for (immediate) validation. The current proposal is just using the qualifier, it might be better to use the complete annotation path (as there exists valuelist and valuelistmapping).

### DIFF
--- a/vocabularies/Common.xml
+++ b/vocabularies/Common.xml
@@ -642,6 +642,11 @@ The minimum structure is
         <Annotation Term="Core.Description" String="If specified as true, there's only one value list mapping and its value list consists of a small number of fixed values" />
       </Term>
 
+      <Term Name="ValueListForValidation" Type="Edm.String" Nullable="false" DefaultValue="" AppliesTo="Property Parameter">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Contains the qualifier of the ValueList or ValueListMapping that should be used for validation" />
+      </Term>
+
       <Term Name="ValueListReferences" Type="Collection(Edm.String)" Nullable="false" AppliesTo="Property Parameter">
         <Annotation Term="Core.IsURL" />
         <Annotation Term="Core.Description" String="A list of URLs of CSDL documents containing value list mappings for this parameter or property" />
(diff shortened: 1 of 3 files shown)


vocabularies/Common.xml
```diff
@@ -642,6 +642,11 @@ The minimum structure is
         <Annotation Term="Core.Description" String="If specified as true, there's only one value list mapping and its value list consists of a small number of fixed values" />
       </Term>
 
+      <Term Name="ValueListForValidation" Type="Edm.String" Nullable="false" DefaultValue="" AppliesTo="Property Parameter">
```
ralfhandl, line 645:
```suggestion
      <Term Name="ValueListForValidation" Type="Edm.String" Nullable="false" AppliesTo="Property Parameter">
```
I would not provide a default value, let's specify the empty string explicitly

BerSie: that's a good point - I will change that. And would you tend to use the pure qualifier or would you prefer the complete annotationpath like @com.sap.vocabularies.v1.Common.ValueListMapping#mySpecialQualifier?

RaizoC:
Hi, I wonder why this annotation is not in the UI vocabulary instead of common, but I assume it is just an indication of the server that this value list is "complete" and thus reasonable to validate against. Nevertheless I see a mismatch between this value help for validation and the value help that is taken for type ahead. Do you think there are scenarios where it makes sense to differ here as I do not? Currently the value help without qualifier is used here. Therefore I would propose a more generic term instead of "ForValidation" (but miss a good proposal though ;-)) that can then also be evaluated for the type ahead feature.

BerSie:
> Hi, I wonder why this annotation is not in the UI vocabulary instead of common, but I assume it is just an indication of the server that this value list is "complete" and thus reasonable to validate against. Nevertheless I see a mismatch between this value help for validation and the value help that is taken for type ahead. Do you think there are scenarios where it makes sense to differ here as I do not? Currently the value help without qualifier is used here. Therefore I would propose a more generic term instead of "ForValidation" (but miss a good proposal though ;-)) that can then also be evaluated for the type ahead feature.

In my opinion the ValueListForValidation should be used for type ahead if one is defined - if not the value list without qualifier is used (as you described). And that's the main difference the type ahead should be available for all fields - the (immediate) validation definitely not. And that's why I wouldn't define it more generically - it's for a specific use case (validation), but it might makes sense for the client to "re-use" it also for other use cases where it makes sense (type ahead).
We could put this term into the UI vocabulary - no problem. I thought to put it where all the other values list terms are defined, but if both of you think it makes more sense to have it in UI, I don't object.

PR #54, opened 2020-09-08
Emphasized

BerSie:
The ability to highlight properties and/or operations (actions, functions) is more and more required. From a visualization perspective these properties might become bold and the actions might become a color (that is not (!) expressing criticality - as error, warning or success).

### DIFF
--- a/vocabularies/UI.json
+++ b/vocabularies/UI.json
@@ -1811,6 +1811,18 @@
             ],
             "@Core.Description": "Parameters for client-calculated criticality, alternative to UI.Criticality"
         },
+        "Emphasized": {
+            "$Kind": "Term",
+            "$Type": "Edm.Boolean",
+            "$Nullable": true,
+            "$DefaultValue": true,
+            "$AppliesTo": [
+                "Record"
+            ],
+            "@Common.Experimental": true,
+            "@Core.Description": "Highlight something that is of special interest",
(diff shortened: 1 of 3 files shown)


vocabularies/UI.json
```diff
@@ -1811,6 +1811,18 @@
             ],
             "@Core.Description": "Parameters for client-calculated criticality, alternative to UI.Criticality"
         },
+        "Emphasized": {
+            "$Kind": "Term",
+            "$Type": "Edm.Boolean",
+            "$Nullable": true,
```
ralfhandl, line 1817: Boolean terms should not be nullable

vocabularies/UI.xml
```diff
@@ -1516,6 +1516,12 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
         <Annotation Term="Core.Description" String="Parameters for client-calculated criticality, alternative to UI.Criticality" />
       </Term>
 
+      <Term Name="Emphasized" Type="Edm.Boolean" DefaultValue="true" AppliesTo="Record">
```
ralfhandl, line 1519:
```suggestion
      <Term Name="Emphasized" Type="Edm.Boolean" DefaultValue="true" Nullable="false" AppliesTo="Record">

```

ralfhandl (changes requested):

ralfhandl (approved):

PR #56, opened 2020-09-21
Tag initial values of in parameters of value lists as significant

BerSie:
Introducing a new property InitialValueIsSignificant on ValueListParameterIn and ValueListParameterInOut to expose the information that the initial value, as empty string, is a valid value and should be used e.g. for filtering the value list.

### DIFF
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -870,7 +870,6 @@
             },
             "InitialValueIsSignificant": {
                 "$Type": "Edm.Bool",
-                "$Nullable": true,
                 "@Common.Experimental": true,
                 "@Core.Description": "Initial value, e.g. empty string, is a valid and significant value"
             }
@@ -893,7 +892,6 @@
             },
             "InitialValueIsSignificant": {
                 "$Type": "Edm.Bool",
-                "$Nullable": true,
                 "@Common.Experimental": true,
                 "@Core.Description": "Initial value, e.g. empty string, is a valid and significant value"
             },
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -867,6 +867,12 @@
             "LocalDataProperty": {
                 "$Type": "Edm.PropertyPath",
                 "@Core.Description": "Path to property that is used to filter the value list with `eq` comparison"
+            },
+            "InitialValueIsSignificant": {
+                "$Type": "Edm.Bool",
+                "$Nullable": true,
+                "@Common.Experimental": true,
+                "@Core.Description": "Initial value, e.g. empty string, is a valid and significant value"
             }
         },
         "ValueListParameterConstant": {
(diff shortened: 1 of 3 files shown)


RaizoC (dismissed):

vocabularies/Common.json
```diff
@@ -867,6 +867,12 @@
             "LocalDataProperty": {
                 "$Type": "Edm.PropertyPath",
                 "@Core.Description": "Path to property that is used to filter the value list with `eq` comparison"
+            },
+            "InitialValueIsSignificant": {
+                "$Type": "Edm.Bool",
+                "$Nullable": true,
```
ralfhandl, line 873: @BerSie nullable Boolean? Better Nullable="false" and DefaultValue="false".

ralfhandl (changes requested):

vocabularies/Common.json
BerSie: added Nullable="false" and changed to DefaultValue="false"

vocabularies/Common.xml
```diff
@@ -689,6 +689,10 @@ The minimum structure is
         <Property Name="LocalDataProperty" Type="Edm.PropertyPath" Nullable="false">
           <Annotation Term="Core.Description" String="Path to property that is used to filter the value list with `eq` comparison" />
         </Property>
+        <Property Name="InitialValueIsSignificant" Type="Edm.Bool" Default="false" Nullable="false">
```
ralfhandl, line 692:
```suggestion
        <Property Name="InitialValueIsSignificant" Type="Edm.Boolean" Default="false" Nullable="false">
```

vocabularies/Common.xml
```diff
@@ -700,6 +704,10 @@ The minimum structure is
         <Property Name="LocalDataProperty" Type="Edm.PropertyPath" Nullable="false">
           <Annotation Term="Core.Description" String="Path to property that is used to filter the value list with `startswith` comparison and filled from the picked value list item" />
         </Property>
+        <Property Name="InitialValueIsSignificant" Type="Edm.Bool" Default="false" Nullable="false">
```
ralfhandl, line 707:
```suggestion
        <Property Name="InitialValueIsSignificant" Type="Edm.Boolean" Default="false" Nullable="false">
```

ralfhandl (changes requested):
Unfortunately we don't have a CSDL validator yet ☹️

RaizoC (approved):

ralfhandl (approved):

Vinod-SAP (commented):
Do we really need InitialValueIsSignificant in ValueListParameterConstant? I mean if someone has annotated ValueListParameterConstant & set the value as empty, it already means this is important isn't it?

ralfhandl:
@Vinod-SAP I only see it in `ValueListParameterIn` and `ValueListParameterInOut`, and we need it in both.

Vinod-SAP:
@ralfhandl, You are correct. My bad :(

ralfhandl:
@Vinod-SAP No worries, the Github preview of the changes was really misleading 😄

PR #57, opened 2020-09-23
Added term ODM.deprecated

ralfhandl:
This annotation is used by Domain Model Alignment (DMA)


vocabularies/ODM.xml
```diff
@@ -33,11 +33,17 @@
         </Collection>
       </Annotation>
 
-      <Term Name="codeList" Type="Core.Tag" DefaultValue="true" AppliesTo="EntityType EntitySet">
+      <Term Name="codeList" Type="Core.Tag" DefaultValue="true" AppliesTo="EntitySet EntityType">
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="The annotated entity set represents a list of code values, the annotated entity type represents an entry in a code list" />
       </Term>
 
+      <Term Name="deprecated" Type="Edm.String" Nullable="false" AppliesTo="EntitySet EntityType Property">
```
HeikoTheissen, line 41: Why can `Core.Revisions` not be used instead?

HeikoTheissen (changes requested):

vocabularies/ODM.xml
ralfhandl: Good question, will ask the DMA team whether they want to use

```swift
    @Core.Revisions : [{
      Kind : #Deprecated,
      Description : 'Duplicate value. Use ''newStuff'' instead.'
    }]
    oldStuff: String;
```

instead of 

```swift
    @ODM.deprecated : 'Duplicate value. Use ''newStuff'' instead.'
    oldStuff: String;
```

HeikoTheissen: Even if they want to stick to their CDS annotation `@ODM.deprecated`, I would still expect this to be translated into an OData annotaion `@Core.Revisions`.

ralfhandl: CAP does not translate annotations from "CDS" to "OData" flavor, instead it allows direct use of OData vocabularies.

See [Annotation Cheat-Sheet for CAP CDS](https://sap.github.io/odata-vocabularies/docs/annotation-cheat-sheet-cap.html).

PR #60, opened 2020-10-02
Example for new "external code" annotation

ralfhandl:


examples/Common.ExternalCode.xml
```diff
@@ -0,0 +1,60 @@
...
+      </Annotations>
+
+      <Annotations Target="my.type/internalCodeProperty" Qualifier="two">
+        <!-- 
+          Pro: flexible text representation, also works for multi-part keys
+          Con: in edit mode it is not clear what has to be edited, and how
+
+          TODO: clarify interaction between ConnectedFields and Common.Text+UI.TextArrangement on the underlying fields
+         -->
+        <Annotation Term="UI.CodeRepresentation" AnnotationPath="related/@UI.ConnectedFields#three" />
```
bernhardthimmel, line 43: I am not sure if the term CodeRepresentation really hits the nail here.
The source property is some kind of (partial) key of the referenced entity and might not think of itself as a code.
In addition, the representation would usually also contain the 'Text'.
What about TextRepresentation?

examples/Common.ExternalCode.xml
```diff
@@ -0,0 +1,60 @@
...
+      </Annotations>
+
+      <!--
+        Not preferred, too much overlap with UI.ConnectedFields
+      -->
+      <Annotations Target="my.type/internalCodeProperty" Qualifier="three">
+        <Annotation Term="Common.ExternalCodes">
```
bernhardthimmel, line 50: Comment from meeting:
This could be accompanied by a ConnectedField annotation that describes how the external code parts should be combined for display.

PR #61, opened 2020-10-07
Update UI.xml

HeikoTheissen:


vocabularies/UI.xml
```diff
@@ -1601,6 +1601,31 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
...
-    </Schema>
+      <Term Name="UserInteraction" Type="UI.UserInteractionType" AppliesTo="NavigationProperty">
+        <Annotation Term="Core.Description" String="When the annotated navigation property occurs in a response, the UI shall use the corresponding collection to immediately interact with the user and then repeat the corresponding request with the additional information obtained from the user"/>
+        <Annotation Term="Core.LongDescription" String="A navigation property thus annotated may be interpreted as not belonging to the object model but containing auxiliary information that is not always needed."/>
+      </Term>
+      <Term Name="UserInteractionSimple" Type="Collection(Edm.PrimitiveType)" AppliesTo="Property">
+        <Annotation Term="Core.Description" String="When this instance annotation occurs in a response, the UI shall let the user choose from the collection and then repeat the corresponding request with the annotated property set to the chosen value"/>
+      </Term>
+      <ComplexType Name="UserInteractionType">
+        <Property Name="Method" Type="UI.UserInteractionMethod" Nullable="false">
```
ralfhandl, line 1612: Why not just have a Boolean "multi-select"?

HeikoTheissen: Because I can imagine more UserInteractionMethods in the future, beyond select single and select multiple. Perhaps the user must type in something, e.g., a second factor for certain critical actions.

vocabularies/UI.xml
```diff
@@ -1601,6 +1601,31 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
         <Annotation Term="Core.Description" String="The contents of this property must not be propagated to the app-to-app navigation context" />
       </Term>
 
-    </Schema>
+      <Term Name="UserInteraction" Type="UI.UserInteractionType" AppliesTo="NavigationProperty">
+        <Annotation Term="Core.Description" String="When the annotated navigation property occurs in a response, the UI shall use the corresponding collection to immediately interact with the user and then repeat the corresponding request with the additional information obtained from the user"/>
+        <Annotation Term="Core.LongDescription" String="A navigation property thus annotated may be interpreted as not belonging to the object model but containing auxiliary information that is not always needed."/>
```
ralfhandl, line 1606: Could you please describe the purpose / use case for this annotation?

vocabularies/UI.xml
```diff
@@ -1601,6 +1601,31 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
         <Annotation Term="Core.Description" String="The contents of this property must not be propagated to the app-to-app navigation context" />
       </Term>
 
-    </Schema>
+      <Term Name="UserInteraction" Type="UI.UserInteractionType" AppliesTo="NavigationProperty">
+        <Annotation Term="Core.Description" String="When the annotated navigation property occurs in a response, the UI shall use the corresponding collection to immediately interact with the user and then repeat the corresponding request with the additional information obtained from the user"/>
+        <Annotation Term="Core.LongDescription" String="A navigation property thus annotated may be interpreted as not belonging to the object model but containing auxiliary information that is not always needed."/>
+      </Term>
+      <Term Name="UserInteractionSimple" Type="Collection(Edm.PrimitiveType)" AppliesTo="Property">
```
ralfhandl, line 1608: Why do you need a second term with almost identical name? When to use which?

vocabularies/UI.xml
```diff
@@ -1601,6 +1601,31 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
         <Annotation Term="Core.Description" String="The contents of this property must not be propagated to the app-to-app navigation context" />
       </Term>
 
-    </Schema>
+      <Term Name="UserInteraction" Type="UI.UserInteractionType" AppliesTo="NavigationProperty">
+        <Annotation Term="Core.Description" String="When the annotated navigation property occurs in a response, the UI shall use the corresponding collection to immediately interact with the user and then repeat the corresponding request with the additional information obtained from the user"/>
+        <Annotation Term="Core.LongDescription" String="A navigation property thus annotated may be interpreted as not belonging to the object model but containing auxiliary information that is not always needed."/>
+      </Term>
+      <Term Name="UserInteractionSimple" Type="Collection(Edm.PrimitiveType)" AppliesTo="Property">
+        <Annotation Term="Core.Description" String="When this instance annotation occurs in a response, the UI shall let the user choose from the collection and then repeat the corresponding request with the annotated property set to the chosen value"/>
```
ralfhandl, line 1609: Please annotate if the term is only intended for instance annotations:
```suggestion
        <Annotation Term="Common.IsInstanceAnnotation" />
        <Annotation Term="Core.Description" String="When this instance annotation occurs in a response, the UI shall let the user choose from the collection and then repeat the corresponding request with the annotated property set to the chosen value"/>
```

vocabularies/UI.xml
```diff
@@ -1601,6 +1601,31 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
...
+      <Term Name="UserInteractionSimple" Type="Collection(Edm.PrimitiveType)" AppliesTo="Property">
+        <Annotation Term="Core.Description" String="When this instance annotation occurs in a response, the UI shall let the user choose from the collection and then repeat the corresponding request with the annotated property set to the chosen value"/>
+      </Term>
+      <ComplexType Name="UserInteractionType">
+        <Property Name="Method" Type="UI.UserInteractionMethod" Nullable="false">
+          <Annotation Term="Core.Description" String="The method for interacting with the user"/>
+        </Property>
+        <Property Name="Target" Type="Collection(Edm.PropertyPath)">
+          <Annotation Term="Core.Description" String="The properties to be set to the additional information obtained from the user"/>
+          <Annotation Term="Core.LongDescription" String="If the annotated NavigationProperty has a ReferentialConstraint, a Target need not be specified"/>
```
ralfhandl, line 1617: Referential constraints are only allowed for single-valued navigation properties.

The description of the term says "the UI shall use the corresponding collection" - so how can there be a referential constraint?

ralfhandl (changes requested):
Please add an example file similar to https://github.com/oasis-tcs/odata-vocabularies/tree/master/examples.

vocabularies/UI.xml
HeikoTheissen: You yourself suggested such a simple instance annotation for "value helps". The other has a more complex structure, so I need two terms.

vocabularies/UI.xml
HeikoTheissen: See https://wiki.wdf.sap.corp/wiki/display/odata/Remedial+action

vocabularies/UI.xml
HeikoTheissen: I'll make Target non-nullable.

vocabularies/UI.xml
ralfhandl: The vocabulary is on public GitHub, so we need a publicly visible description, either as part of the term definition, or linked from the term definition.

HeikoTheissen:
Why are _all_ .md files now added to this branch?

vocabularies/UI.xml
HeikoTheissen: I have described a use case in the example file.

ralfhandl:
> Why are _all_ .md files now added to this branch?

Your branch was out-of-date with master. Merging master made the other .md files disappear

vocabularies/examples/com.sap.vocabularies.UI.v1.UserInteraction-sample.xml
```diff
@@ -0,0 +1,42 @@
...
+  <edmx:Reference Uri="/sap/opu/odata4/product.xml">
+    <edmx:Include Namespace="com.sap.gateway.entities.product" Alias="Product" />
+  </edmx:Reference>  
+  <edmx:DataServices>
+    <Schema Namespace="userinteraction.sample" xmlns="http://docs.oasis-open.org/odata/ns/edm">
+      <!-- Certain user interactions cannot be handled by one reqeust, because they follow the pattern "user triggers, server offers, user chooses".
+           For example, the action "exchange product in a sales order item" is triggered by the user on a sales order item
+           (POST ~/SalesOrderItems(...)/ExchangeProduct), the server returns the item with an inlined collection of alternative products,
+           from which the user must then choose, and the request must be repeated with the chosen key
+           (POST ~/SalesOrderItems(...)/ExchangeProduct?NewProduct='...').
```
SNairVinod, line 18: So how is it expected show in the UI. User performs an action the result is shown in a Dialog with selectable list is it? What if the use case support selecting more than one item

SNairVinod (commented):
If I understood your use case & today I have to model this, I will have to create action which has a parameter and an associated Value Help. User should be able to select entry from the Value help and execute the action. Isn't it? 

Did I get the use case correctly or I have understood this completely wrong?

vocabularies/examples/com.sap.vocabularies.UI.v1.UserInteraction-sample.xml
HeikoTheissen: Yes, but this still requires alignment with the UI frameworks. (That's why this PR is still a draft.)

HeikoTheissen:
@SNairVinod, in simple cases you could model this as an action parameter with value help. Users would then have to click twice: first to open the parameter popup, then to trigger the value help. User flow becomes even more confusing if
* the action has other parameters that users must fill before the value help can be triggered
* the action may sometimes succeed without additional user interaction (product exchange is not a good example for this, but think of order creation with quote reference: if there is more than one matching quote, the user must choose, if there is one matching quote, the user must confirm, if there is no matching quote, no user interaction is required).
Such cases are better dealt with if we don't let the user pick a value help but let the server decide which navigation property to return (in the product exchange example, this is navigation property `_AlternativeProducts`, but there could be others).

vocabularies/examples/com.sap.vocabularies.UI.v1.UserInteraction-sample.xml
SNairVinod: Ok

vocabularies/UI.xml
```diff
@@ -1687,6 +1687,51 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
...
-    </Schema>
+      <Term Name="UserInteraction" Type="UI.UserInteractionType" Nullable="false" AppliesTo="NavigationProperty">
+        <Annotation Term="Core.Description">
+          <String>When the annotated navigation property or its `odata.navigationLink` occurs in a response,
+          the UI shall use the addressed entity or collection to immediately interact with the user
+          and then repeat the corresponding request with the additional information obtained from the user</String>
+        </Annotation>
+        <Annotation Term="Core.LongDescription">
+          <String>If the response is an [error response](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_ErrorResponse),
+          the annotated navigation property MAY occur in the `innererror` property which has the format of a [message body](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_MessageBody).
```
HeikoTheissen, line 1698: Instead of `innererror`, we could use a new instance annotation of type `Edm.Untyped`, but this would have to come from a new vocabulary that uses version `Edmx Version="4.01"`.

ralfhandl: I'd prefer a dedicated instance annotation, the `innererror` is intended for debugging information.

vocabularies/ConflictResponse.json
```diff
@@ -0,0 +1,51 @@
...
+        "https://sap.github.io/odata-vocabularies/vocabularies/Common.json": {
+            "$Include": [
+                {
+                    "$Namespace": "com.sap.vocabularies.Common.v1",
+                    "$Alias": "Common"
+                }
+            ]
+        }
+    },
+    "com.sap.vocabularies.ConflictResponse.v1": {
```
uhlmannm, line 21: Tun wir uns mit einem so speziellen Namen einen Gefallen?

HeikoTheissen: Other suggestions are welcome.

HeikoTheissen: What about ErrorResponse?

ralfhandl: How about UserInteraction as the vocabulary name?

HeikoTheissen: ErrorResponse solves a more general problem: How can we embed structured data into an error response. Navigation properties annotated with UI.UserInteraction are just one example for this.

uhlmannm: UserInteraction sounds more generic to me. The question is really where we are heading here. Is this a vocabulary specifically for the user interaction purpose we are discussing today? Or will this rather be a Common-4.01 vocabulary for everything that does not fit into Common?

HeikoTheissen: It is a vocabulary that helps us embed structured data into error responses because we do not want to use 409 Conflict responses for that.

ralfhandl: Instance annotations for error responses are already in the Common vocabulary, so using another "Common" vocabulary seems appropriate:
- Common-4.01
- CommonPlus
- Common2
- ...

HeikoTheissen: Common_401 (hyphens and dots are not allowed in simple identifiers)?

PR #63, opened 2020-10-16
Refresh markdown: show nullable

ralfhandl:
Nullability is now indicated with a `?` suffixed to the type of a term or property.

This should help spotting accidental nullability, in addition to providing more information at one glance.


BerSie (approved):
Interesting :-) ... maybe that helps to avoid missing the Nullable option

PR #68, opened 2020-11-03
Common.Timezone

HeikoTheissen:


BerSie:
For me this looks already as a good starting point. I would avoid the comment with the browser (user's timezone is sufficient I believe). What is still needed to get that meaningful working is a kind of code list and the calculation rules.

ralfhandl (dismissed):

ralfhandl (approved):

vocabularies/UI.xml
```diff
@@ -1605,6 +1604,16 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
         <Annotation Term="Core.Description" String="The contents of this property must not be propagated to the app-to-app navigation context" />
       </Term>
 
+      <Term Name="DisplayTimeZone" Type="Edm.String" Nullable="false" AppliesTo="Property Parameter">
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property shall be presented in the context of the given time zone"/>
+        <Annotation Term="Core.LongDescription">
+          <String>The value list for time zones is not prescribed by this vocabulary.
+The `Edm.DateTimeOffset` in the annotated property SHOULD contain a [timezoneOffset](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#vp-dt-timezone) that corresponds to the time zone.
+In the absence of this annotation, points in time are typically presented in the current user's time zone.</String>
+        </Annotation>
```
ralfhandl, line 1613:
```suggestion
        <Annotation Term="Core.Description" String="The point in time represented by the annotated property shall be presented in the context of the given time zone"/>
        <Annotation Term="Core.LongDescription">
          <String>The value of this annotation is only used for display and not used for timezone calculation.
Instead the [timezoneOffset](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#vp-dt-timezone) in the `Edm.DateTimeOffset` value of the annotated property is removed, and the dateTime part is used for displaying the timestamp. It is the service's responsibility to keep the symbolic timezone and the numeric offset in sync.</String>
        </Annotation>

```

ralfhandl (changes requested):

vocabularies/UI.xml
HeikoTheissen: Simply removing the timezoneOffset may _change_ the value of the timestamp. This is justified only if it was the correct timezoneOffset to begin with. I'd rather keep my wording and possibly replace SHOULD by MUST.

HeikoTheissen:
This has been obsoleted by #84.

vocabularies/UI.xml
```diff
@@ -1609,6 +1608,17 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
         <Annotation Term="Core.Description" String="The contents of this property must not be propagated to the app-to-app navigation context" />
       </Term>
 
+      <Term Name="Timezone" Type="Edm.String" AppliesTo="EntityContainer Property Parameter">
```
ralfhandl, line 1611: How is this supposed to work when applied to the entity container? Static user-independent values don't make much sense, and user-dependent values would require an absolute path to a property of a "user-settings" singleton.

Remove this for now and just apply it to properties and parameters?

ralfhandl (changes requested):

vocabularies/UI.xml
```diff
@@ -1609,6 +1600,17 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
         <Annotation Term="Core.Description" String="The contents of this property must not be propagated to the app-to-app navigation context" />
       </Term>
 
+      <Term Name="Timezone" Type="Edm.String" AppliesTo="EntityContainer Property Parameter">
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or all points in time in this entity container shall be presented in the context of the given time zone"/>
+        <Annotation Term="Core.LongDescription">
+          <String>The value list for time zones is not prescribed by this vocabulary.
+            If a [IANA](https://www.iana.org/time-zones) time zone is specified, browsers can perform the necessary conversion.
+            Otherwise, the `Edm.DateTimeOffset` in the annotated property SHOULD contain a [timezoneOffset](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#vp-dt-timezone) that corresponds to the time zone.
```
ralfhandl, line 1608: I would remove this sentence unless we want to require that UI5 can do this.

vocabularies/UI.xml
```diff
@@ -1609,6 +1600,17 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
...
 
+      <Term Name="Timezone" Type="Edm.String" AppliesTo="EntityContainer Property Parameter">
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or all points in time in this entity container shall be presented in the context of the given time zone"/>
+        <Annotation Term="Core.LongDescription">
+          <String>The value list for time zones is not prescribed by this vocabulary.
+            If a [IANA](https://www.iana.org/time-zones) time zone is specified, browsers can perform the necessary conversion.
+            Otherwise, the `Edm.DateTimeOffset` in the annotated property SHOULD contain a [timezoneOffset](https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#vp-dt-timezone) that corresponds to the time zone.
+            In the absence of this annotation, points in time are typically presented in the current user's time zone.</String>
+        </Annotation>
+        <Annotation Term="Core.RequiresType" String="Edm.DateTimeOffset" />
```
ralfhandl, line 1611:
```suggestion
        <Annotation Term="Core.RequiresType" String="Edm.DateTimeOffset" />
        <Annotation Term="Common.Experimental" />
```

examples/UI.Timezone-sample.xml
```diff
@@ -0,0 +1,44 @@
...
+      <EntityType Name="WorkerTimeSheet">
+        <Property Name="ClockInDateTime" Type="Edm.DateTimeOffset">
+          <!-- This timestamp shall be presented in the context of its time zone, but the time zone shall
+               be represented simply by the timezoneOffset, e.g., 2020-12-09-T06:00:00+05:00 as Dec 09 2020, 6am, +05:00
+          -->
+          <Annotation Term="UI.Timezone">
+            <Apply Function="SAP.timezoneOffset">
+              <Path>ClockInDateTime</Path>
+            </Apply>
+          </Annotation>
```
ralfhandl, line 23: How about a simple example with a string property carrying the clear-text time zone?

I am not aware of this function and its semantics.

examples/UI.Timezone-sample.xml
```diff
@@ -0,0 +1,44 @@
...
+        <Property Name="ClockInTime" Type="Edm.TimeOfDay"/>
+        <Annotation Term="Common.Timestamp" Qualifier="ClockIn">
+          <Apply Function="SAP.UTCTimestamp">
+            <Path>ClockInDate</Path>
+            <Path>ClockInTime</Path>
+          </Apply>
+          <Annotation Term="UI.Timezone">
+            <Apply Function="SAP.timezoneOffset">
+              <Path>@Common.Timestamp#ClockIn</Path>
+            </Apply>
```
ralfhandl, line 38: Where are these functions defined? Please point to definition or use a self-describing example.

ralfhandl (changes requested):

vocabularies/UI.xml
HeikoTheissen: Which of the three sentences?

vocabularies/UI.xml
HeikoTheissen: Can't the value be user-dependent (just like the texts in the metadata depend on the user's language)? It could be the timezone from the user settings.

If not, then let's apply it only to properties and parameters.

ralfhandl: Neither RAP nor CAP support user-dependent $metadata.

vocabularies/UI.xml
ralfhandl, line 1609:
```suggestion
```

examples/UI.Timezone-sample.xml
```diff
@@ -0,0 +1,22 @@
...
+<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/UI.xml">
+    <edmx:Include Namespace="com.sap.vocabularies.UI.v1" Alias="UI" />
+  </edmx:Reference>
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/Common.xml">
+    <edmx:Include Namespace="com.sap.vocabularies.UI.v1" Alias="Common" />
+  </edmx:Reference>
+  <edmx:Reference Uri="/sap/opu/odata4/standard-functions.xml">
+    <edmx:Include Namespace="com.sap.gateway.functions.standard" Alias="SAP" />
+  </edmx:Reference>
```
ralfhandl, line 11:
```suggestion
```

ralfhandl (changes requested):

vocabularies/UI.xml
```diff
@@ -1609,6 +1600,17 @@ This is in contrast to [DataFieldForIntentBasedNavigation](#DataFieldForIntentBa
...
       </Term>
 
+      <Term Name="Timezone" Type="Edm.String" AppliesTo="Property Parameter">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or all points in time in this entity container shall be presented in the context of the given time zone"/>
+        <Annotation Term="Core.LongDescription">
+          <String>The value list for time zones is not prescribed by this vocabulary.
+            If a [IANA](https://www.iana.org/time-zones) time zone is specified, browsers can perform the necessary conversion.
+            In the absence of this annotation, points in time are typically presented in the current user's time zone.</String>
+        </Annotation>
```
ralfhandl, line 1610:
```suggestion
        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or parameter shall be presented in the given time zone"/>
        <Annotation Term="Core.LongDescription">
          <String>The value list for time zones is not prescribed by this vocabulary.
            If an [IANA](https://www.iana.org/time-zones) time zone is specified, browsers can perform the necessary conversion.
            In the absence of this annotation, points in time are typically presented in the current user's time zone.</String>
        </Annotation>
```

ralfhandl (changes requested):

ralfhandl (approved):

ralfhandl (approved):

ralfhandl (approved):

ralfhandl (approved):

HeikoTheissen:
@ralfhandl, should this be in the Common vocabulary instead of UI?

ralfhandl:
The original intent was exclusively for representation on a UI.

HeikoTheissen:
> The original intent was exclusively for representation on a UI.

But wouldn't that be valuable for other consumers as well?

ralfhandl:
Would we want to have this annotation in A2X services published in API Business Hub?

If yes, then let's move it to Common.

vocabularies/Common.xml
```diff
@@ -113,6 +113,17 @@
         </Member>
       </EnumType>
 
+      <Term Name="Timezone" Type="Edm.String" AppliesTo="Property Parameter">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or parameter shall be presented in the given time zone"/>
```
ralfhandl, line 118: "shall be presented" is UI terminology. Please replace with "common" terminology.

vocabularies/UI.xml
```diff
@@ -106,7 +106,6 @@ a collection of business object instances, e.g. as a list or table.</String>
         <Annotation Term="UI.ThingPerspective" />
         <Annotation Term="Core.Description" String="Collection of fields identifying the object" />
       </Term>
```
ralfhandl, line 108:
```suggestion
      </Term>

```

ralfhandl (changes requested):

ralfhandl (dismissed):

ralfhandl (approved):

ralfhandl (approved):

examples/Common.Timezone-sample.json
```diff
@@ -0,0 +1,29 @@
+{
+    "$Version": "4.0",
+    "$Reference": {
+        "https://sap.github.io/odata-vocabularies/vocabularies/UI.xml": {
```
BerSie, line 4: Common.xml

examples/Common.Timezone-sample.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/UI.xml">
```
BerSie, line 3: Common.xml

vocabularies/Common.xml
```diff
@@ -113,6 +113,17 @@
         </Member>
       </EnumType>
 
+      <Term Name="Timezone" Type="Edm.String" AppliesTo="Property Parameter">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or parameter shall be interpreted in the given time zone" />
+        <Annotation Term="Core.LongDescription">
+          <String>The value list for time zones is not prescribed by this vocabulary.
+            If an [IANA](https://www.iana.org/time-zones) time zone is specified, browsers can perform the necessary conversion.
```
uhlmannm, line 121: Would it not be possible to request that the time zone must be an IANA time zone? In the end, we want to interpret it in the browser. So what help would other time zones be?

HeikoTheissen: For example, browsers also understand CET. Or in the future, they may understand other standards.

BerSie: So, the gracious renderer has to understand it to be able to convert. Otherwise it's only for visualization purposes? And it's not the "browser" that performs the conversion - it has to be client-side code (that might run in the browser).

vocabularies/Common.xml
```diff
@@ -113,6 +113,17 @@
         </Member>
       </EnumType>
 
+      <Term Name="Timezone" Type="Edm.String" AppliesTo="Property Parameter">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or parameter shall be interpreted in the given time zone" />
+        <Annotation Term="Core.LongDescription">
+          <String>The value list for time zones is not prescribed by this vocabulary.
```
HeikoTheissen, line 120:
```suggestion
```

vocabularies/Common.xml
HeikoTheissen, line 121:
```suggestion
          <String>If an [IANA](https://www.iana.org/time-zones) time zone is specified, browsers can perform the necessary conversion.
```

HeikoTheissen (commented):
OK if I delete just the "not prescribed" sentence?

vocabularies/Common.xml
HeikoTheissen: > And it's not the "browser" that performs the conversion - it has to be client-side code (that might run in the browser).

The necessary Javascript code is one line, but the real work is done by the browser (or perhaps the operating system).

uhlmannm: It seems that CET is a deprecated IANA time zone. In the process of consuming this annotation we will need to find out whether all ABAP time zones are actually deprecated IANA time zones or just some.

BerSie (approved):
OK. With that we can at least display the timezone. And in principle we have the possibility to convert it.

uhlmannm (approved):

PR #70, opened 2020-11-20
Remove QuickInfo from VL and VLM

BerSie:
As Common.QuickInfo can be used on every artifact, it can be used for annotations, too. So there is no need to have a dedicated property. With this change this experimental property will be removed.


ralfhandl (approved):
Please review

BerSie (commented):
Thanks for adding that

PR #71, opened 2020-12-01
Update Common.xml

HeikoTheissen:


vocabularies/Common.xml
```diff
@@ -1063,6 +1063,9 @@ only `TargetProperties` and `TargetEntities` are relevant. They are addressed vi
         <Property Name="SourceEntities" Type="Collection(Edm.NavigationPropertyPath)" Nullable="false">
           <Annotation Term="Core.Description" String="Changes to one or more of these entities may affect the targets. An empty path means the annotation target." />
         </Property>
+        <Property Name="SourceAnnotations" Type="Collection(Edm.AnnotationPath)" Nullable="false">
+          <Annotation Term="Core.Description" String="Each `Edm.AnnotationPath` must be valid for [`UI.ReferenceFacet/Target`](UI.md#ReferenceFacet). Changes to the values of one or more of the referenced structural properties may affect the targets." />
```
ralfhandl, line 1067: Can you please explain in the LongDescription of the complex type how these will be used? After all annotations don't change in $metadata.

ralfhandl (changes requested):

vocabularies/Common.xml
HeikoTheissen: No quite sure whether my change addresses what you meant. Please check again.

uhlmannm:
Hi @BerSie ,

I am not deep enough into the annotation topic to provide a meaningful review. Could you please have a look?

Many thanks
Mathias.

vocabularies/Common.xml
```diff
@@ -1063,6 +1063,12 @@ only `TargetProperties` and `TargetEntities` are relevant. They are addressed vi
         <Property Name="SourceEntities" Type="Collection(Edm.NavigationPropertyPath)" Nullable="false">
           <Annotation Term="Core.Description" String="Changes to one or more of these entities may affect the targets. An empty path means the annotation target." />
         </Property>
+        <Property Name="SourceFacets" Type="Collection(Edm.AnnotationPath)" Nullable="false">
+          <Annotation Term="Core.Description">
+            <String>A collection of `Edm.AnnotationPath`s that are also valid for [`UI.ReferenceFacet/Target`](UI.md#ReferenceFacet), i.e., that reference a UI facet.
```
ralfhandl, line 1068:
```suggestion
            <String>A collection of annotation paths that are also valid for [`UI.ReferenceFacet/Target`](UI.md#ReferenceFacet), i.e., that reference a UI facet.
```

ralfhandl (approved):

ralfhandl (approved):

BerSie:
Hi @HeikoTheissen , it would be really helpful to get an example of using this new property. What is the purpose and what is a use case you can do with that property that is not covered with the other one. And do you have a stakeholder that is requiring (I feel not sure if the coupling of fields in a facet are a good "source" for a side effect) - and if so, I would remove the reference (re-direction) to a facet and use the annotation path directly. What do you think, @HeikoTheissen , @ralfhandl ?

ralfhandl:
An example would certainly help!

My understanding was that the annotation paths have to lead to annotations with a term that is tagged with `UI.ThingPerspective`.

Maybe don't talk about reference facets in the description, and maybe rename to "SourcePerspectives" instead of "SourceFacets".

ralfhandl (dismissed):
See last comment

BerSie (approved):
Looks now good to me

PR #73, opened 2020-12-09
Adding Common.ExternalID

BerSie:
New terms Common.ExternalID and Common.ExternalIDs


RaizoC:
What is the reason for introducing two different annotations? Would one be sufficient to cover both use cases?
Who would evaluate this information and how?

ralfhandl (commented):
Same question as @RaizoC: seeing them side-by-side the second one doesn't look that different, and covers also the first one as a special case.

Why not only add the second array form?

bernhardthimmel:
@RaizoC , @ralfhandl 
The reasoning in the meeting was rather based on the client effort. It is not clear how many objects will really need a list of codes and there are many details that need to be clarified like how to concatenate these codes and how to exactly handle the code/name combinations then. The one-component list does not have these issues.
So do we want to have a list parameter that almost always only has one entry and for which the behavior does not work yet when multiple entries are added. By splitting this, the different enhancements can be split more easily.
@RaizoC : I'll also forward you some background slides.

RaizoC:
Ok, understood, but then this highly depends on how this finally will be used. We have a clear UX guideline how such combined fields should be displayed, so the question is whether this is done already on server side (by always providing one field) or on client side. As it is a UX guideline (that also might change over time) probably the better approach is to do this on client side.
In either case I would vote for one single annotation as I doubt it will be easier or more difficult for the UI in the given proposal to evaluate both or to evaluate one by checking if it only consists of one property.

BerSie:
> Ok, understood, but then this highly depends on how this finally will be used. We have a clear UX guideline how such combined fields should be displayed, so the question is whether this is done already on server side (by always providing one field) or on client side. As it is a UX guideline (that also might change over time) probably the better approach is to do this on client side.
> In either case I would vote for one single annotation as I doubt it will be easier or more difficult for the UI in the given proposal to evaluate both or to evaluate one by checking if it only consists of one property.

The decision how these fields visualized together is and will stay on the client side. Having an external id gives applications that have a very technical key (guid) and besides a (maybe translatable) description a user-readable/recognizable key the possibility to define these relationships and FE is getting the chance to render this readable key e.g. together with the description as defined in the UX guideline (unfortunately I was not able to find this definition anywhere ;-), but it's TextFirst currently.
One discussion point was if we need the collection. This was brought up that there might be use cases with combined keys where this is needed. @bernhardthimmel had afaik only use cases for the single id scenarios. The plan was that (based on stakeholder requirements) the first term will be implemented and if there is a need for the collection-based term, it has to be implemented in addition. But the latter one is from a definition and implementation perspective definitely not the easiest one and as we haven't a stakeholder for it yet, we wanted to separate it.

ralfhandl:
> the latter one is from a definition and implementation perspective definitely not the easiest one and as we haven't a stakeholder for it yet, we wanted to separate it.

In this case I would not define the second annotation now. If you want to capture the idea, make it an XML comment.

bernhardthimmel:
> > the latter one is from a definition and implementation perspective definitely not the easiest one and as we haven't a stakeholder for it yet, we wanted to separate it.
> 
> In this case I would not define the second annotation now. If you want to capture the idea, make it an XML comment.

That is ok with me. @BerSie : Would you remove the Common.ExternalIDs? We still have the internal discussion about it, so do not really need to keep it in the comments.

ralfhandl (dismissed):

bernhardthimmel (approved):

bernhardthimmel (approved):

PR #77, opened 2020-12-16
UI.*Hidden applies to EntityType

HeikoTheissen:


ralfhandl (approved):

drktfl:
Hi,

what is the rationale behind this change? And what are the semantics now? The description is unclear as it only mentions EntitySets, which are the natural target for UI.CreateHidden, in my opinion. This change definitely increases complexity for consumers.

Thanks,
Dirk

HeikoTheissen:
UI.*Hidden annotations are often dynamic expressions. However, these are not possible in the case of implicit entity sets that are defined via a containment navigation property, see https://issues.oasis-open.org/browse/ODATA-1416. In such cases, the UI.*Hidden annotation can only be placed at the entity _type_.

drktfl:
Does it really help to just extend the possible targets if the underlying problem is something else?

This would in the long run mean that _every_ annotation that can target an EntitySet can implicitly also target an EntityType. Or where would this stop?

HeikoTheissen:
That is the issue still being discussed in https://issues.oasis-open.org/browse/ODATA-1416.

PR #79, opened 2021-01-04
Disable intent-based navigation

BerSie:
Give application developers the chance to define the enablement of the DataFieldForIntentBasedNavigation based e.g. on the user context, selected context and existing parameters.

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1446,7 +1446,7 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
         <Property Name="Action" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Name of the Action on the Semantic Object. If not specified, let user choose which of the available actions to trigger." />
         </Property>
-        <Property Name="NavigationAvailable" Type="Edm.Boolean" Nullable="true" DefaultValue="true">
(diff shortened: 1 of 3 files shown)


RaizoC (dismissed):

vocabularies/UI.xml
```diff
@@ -1446,6 +1446,10 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
         <Property Name="Action" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Name of the Action on the Semantic Object. If not specified, let user choose which of the available actions to trigger." />
         </Property>
+        <Property Name="NavigationAvailable" Type="Edm.Boolean" Nullable="true" DefaultValue="true">
```
ralfhandl, line 1449:
```suggestion
        <Property Name="NavigationAvailable" Type="Edm.Boolean" Nullable="false" DefaultValue="true">
```

I assume we do not want `null` here

ralfhandl (changes requested):

vocabularies/UI.xml
BerSie: Hmm, yes you are right. I thought at first it doesn't matter, but I agree now ...

ralfhandl (approved):

PR #82, opened 2021-01-19
Common.skip annotation for sap-skipfilter query option

HeikoTheissen:


vocabularies/Common.md
```diff
@@ -29,86 +29,87 @@ Term|Type|Description
 [SecondaryKey](./Common.xml#L206:~:text=<Term%20Name="-,SecondaryKey,-") *(Deprecated)*|\[PropertyPath\]|<a name="SecondaryKey"></a>Use term `AlternateKeys` from the OASIS Core vocabulary instead
 [MinOccurs](./Common.xml#L218:~:text=<Term%20Name="-,MinOccurs,-")|Int64|<a name="MinOccurs"></a>The annotated set or collection contains at least this number of items
 [MaxOccurs](./Common.xml#L222:~:text=<Term%20Name="-,MaxOccurs,-")|Int64|<a name="MaxOccurs"></a>The annotated set or collection contains at most this number of items
-[AssociationEntity](./Common.xml#L226:~:text=<Term%20Name="-,AssociationEntity,-")|\[NavigationPropertyPath\]|<a name="AssociationEntity"></a>Entity representing an n:m association with attributes
```
ThomasChadzelek, line 32: This Common.md file is derived from one of the others? Which one is the original? (Because this diff is too large, I would like to just review your manual changes)

vocabularies/Common.xml
```diff
@@ -223,6 +223,14 @@
         <Annotation Term="Core.Description" String="The annotated set or collection contains at most this number of items" />
       </Term>
 
+      <Term Name="skip" Type="Edm.Int64" AppliesTo="Collection">
+        <Annotation Term="Core.IsInstanceAnnotation"/>
+        <Annotation Term="Core.Description" String="The server assumed this value for $skip when it determined the annotated collection"/>
```
ThomasChadzelek, line 228: Sounds like the annotation would even be expected if $skip was explicitly given? Or does "assume" intend to rule this out?

P.S.: "The query option sap-skiplocation=<entity URI>[,<n>] instructs the server to return an entity collection with the number of entries specified by $top, with a $skip value chosen by the server such that the entity whose URI is given in sap-skiplocation is visible, preferably with <n> entities before it." --> where will this be documented? I like it!

ThomasChadzelek (dismissed):

vocabularies/Common.md
HeikoTheissen: Common.xml is the original.

vocabularies/Common.xml
HeikoTheissen: Yes. If the request contains a `$skip`, the server need not assume a value.

uhlmannm (approved):

PR #83, opened 2021-01-25
New term UI.IsCopyAction

ralfhandl:


BerSie (dismissed):

RaizoC:
Hmm, now reading it the copy-semantics seems very special and is taken out of the create semantics. I do not understand how this different and more special semantics impacts anything on client side different than the pure "create"-semantics (which I think is very valid). If there really is a use case to handle that one differently (does it need to be unique?) further constraints need to be added to the "copy" semantics, i.e. that it is a bound "create"-action that has the same (bound) entity as result entity.

ralfhandl:
@RaizoC currently we only have a use case for "copy". Should I remove "create" for now, or comment it out?

BerSie:
> Hmm, now reading it the copy-semantics seems very special and is taken out of the create semantics. I do not understand how this different and more special semantics impacts anything on client side different than the pure "create"-semantics (which I think is very valid). If there really is a use case to handle that one differently (does it need to be unique?) further constraints need to be added to the "copy" semantics, i.e. that it is a bound "create"-action that has the same (bound) entity as result entity.

I think the UI can check if an entry was selected before activating the copy function (it knows that an instance has to be selected). For create (simple create without reference) you would not expect that.

RaizoC:
My feedback rather goes into the direction that this is too specific as the client might only be interested in the "create"-semantics. The UI selection depends on whether the action is instance-bound or not, so for this no semantics is needed. What else would the client do with this information?

RaizoC:
I still wonder if these semantics are disjoint, i.e. from the description every "copy" action would also be tagged as "create" action. Is this the intention or does "copy" implies the "create" fact?

ralfhandl:
We can remove "create" for now because https://jira.tools.sap/browse/FIORITECHP1-20377 only requires "copy"

ArunBharathiKumarasamy (dismissed):

RaizoC (approved):

BerSie:
Why did we put it on the DataFieldForAction and not on the action itself (it should be bound to the action and not its usage/reference)? Sometimes these pull requests are running too long :-) (at least for my old brain).

ralfhandl:
> Why did we put it on the DataFieldForAction and not on the action itself (it should be bound to the action and not its usage/reference)?

Good point!

RaizoC (dismissed):

RaizoC (approved):
FE already implemented the same.

vocabularies/UI.xml
```diff
@@ -1412,6 +1412,18 @@ The trend is
         </Annotation>
       </Term>
 
+      <Term Name="IsCopyAction" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Record">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.RequiresType" String="UI.DataFieldForAction" />
+        <Annotation Term="Core.Description" String="The annotated `DataFieldForAction` record within a `UI.LineItem` annotation references an action that deep-copies an instance of the annotated entity type" />
```
HeikoTheissen, line 1418:
```suggestion
        <Annotation Term="Core.Description" String="The annotated [`DataFieldForAction`](#DataFieldForAction) record within a [`UI.LineItem`](#LineItem) annotation references an action that deep-copies an instance of the annotated entity type" />
```

vocabularies/UI.xml
```diff
@@ -1412,6 +1412,18 @@ The trend is
...
+        <Annotation Term="Core.LongDescription">
+          <String>The referenced action MUST be bound to the annotated entity type and creates a new instance of the same entity type as a deep copy of the bound instance.
```
HeikoTheissen, line 1420:
```suggestion
          <String>The referenced action MUST be bound to the annotated entity type and create a new instance of the same entity type as a deep copy of the bound instance.
```

ralfhandl:
> FE already implemented the same.

I've kept it "experimental", please notify me once this has also been implemented in RAP and proven to work end-to-end.

Thanks in advance!

HeikoTheissen (changes requested):
Add links

PR #84, opened 2021-02-02
New term UI.DisplayTimeZone

ralfhandl:


RaizoC (dismissed):

HeikoTheissen:
Please align this with #68.

vocabularies/UI.xml
```diff
@@ -1291,7 +1291,10 @@ The trend is
         <Annotation Term="Validation.Pattern" String="^[+-]?[0-9][0-9]:[0-5][0-9]$" />
         <Annotation Term="Core.Description" String="Display the annotated timestamp field in this time zone" />
         <Annotation Term="Core.LongDescription">
-          <String>The display time zone is represented as an offset ±hh:mm and can be a path expression</String>
+          <String>The display time zone is represented as an offset ±hh:mm and can be a path expression.
+If a path expression is used to reference a field containing the numeric offset,
+this offset field can in turn be annotated with `Common.Text` pointing to a third field
+containing a textual description of the display time zone.</String>
```
HeikoTheissen, line 1297: Why do you need both a textual description and the timezone offset `+/-hh:mm`? The timezone offset can be transported with the `Edm.DateTimeOffset` property itself, as proposed in #68.

HeikoTheissen (changes requested):

vocabularies/UI.xml
ralfhandl: @RaizoC what do you think about placing the offset in the timestamp value?

@HeikoTheissen since there is no definitive code list for symbolic time zones a display timezone such as `CST` cannot be used for calculations in the client. The timestamp would have to be displayed "as is", trusting that the numeric offset in the DateTimeOffset value matches the symbolic timezone in the DisplayTimeZone annotation.

RaizoC: @ralfhandl I don't think we should put some semantics into the offset of the Edm.DateTimeOffset content as this might be misleading regarding its interpretation (and would also be incompatible to introduce such a semantics afterwards). Up to now the value is just a representation of one and the same timestamp without any semantics defined. I like the approach with a dedicated annotated property as it also allows to use it accordingly for filtering, sorting, etc. Further if we would have the need to add further semantics / reasonable display time zones to the timestamp we could easily do that (e.g. timestamp in sending and receiving location or timestamp in start and destination airport.

HeikoTheissen: > @HeikoTheissen since there is no definitive code list for symbolic time zones a display timezone such as `CST` cannot be used for calculations in the client.

I doubt that any smartphone would function if there wasn't a standardized time zone database. See, for example, https://github.com/eggert/tz/blob/master/europe#L1490 for the definition of Europe/Berlin. The same file also defines (on line #L743) CET "for backward compatibility".

And https://github.com/eggert/tz/blob/master/northamerica#L204 defines CST6CDT, but CST does not appear as a time zone ID.

ralfhandl: @HeikoTheissen Maybe this will work: https://www.iana.org/time-zones

Please tell me what you find out.

HeikoTheissen: The eggert/tz repository is mentioned on https://data.iana.org/time-zones/tz-link.html.

vocabularies/UI.xml
```diff
@@ -1287,14 +1287,13 @@ The trend is
 
       <Term Name="DisplayTimeZone" Type="Edm.String" AppliesTo="Property">
         <Annotation Term="Common.Experimental" />
-        <Annotation Term="Core.RequiresType" String="Edm.DateTimeOffset" />
```
HeikoTheissen: You removed the `RequiresType` in order to be able to annotate separate date and time properties.
* How do you declare a (date, time) pair?
* Annotating only a date with an offset makes no sense
* Annotating only a time makes sense (and can lead to something like "10pm on the previous day").

HeikoTheissen (changes requested):

vocabularies/UI.xml
```diff
@@ -1285,13 +1285,17 @@ The trend is
...
-        <String>Contains information for time- or date-time-fields in which time zone the time value should be displayed.</String>
+      <Term Name="DisplayTimeZone" Type="Edm.String" AppliesTo="Property">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Validation.Pattern" String="^[+-]?[0-9][0-9]:[0-5][0-9]$" />
+        <Annotation Term="Core.Description" String="Display the annotated timestamp, time, or date field in this time zone" />
+        <Annotation Term="Core.LongDescription">
+          <String>The display time zone is represented as an offset ±hh:mm and can be a path expression.
+If a path expression is used to reference a field containing the numeric offset,
+this offset field can in turn be annotated with `Common.Text` pointing to a third field
+containing the symbolic name or textual description of the display time zone.</String>
```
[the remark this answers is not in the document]
ralfhandl: Yes, having an offset field with variable values and a constant Common.Text value seems a bit weird.

vocabularies/UI.xml
```diff
@@ -1285,13 +1285,17 @@ The trend is
         </Member>
       </EnumType>
 
-      <!-- Under discussion
-        <Term Name="DisplayTimeZone" Type="?Edm.Int32?" AppliesTo="Property">
-        <Annotation Term="Core.Description">
-        <String>Contains information for time- or date-time-fields in which time zone the time value should be displayed.</String>
+      <Term Name="DisplayTimeZone" Type="Edm.String" AppliesTo="Property">
```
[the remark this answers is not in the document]
ralfhandl: Good idea

vocabularies/UI.xml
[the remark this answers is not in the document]
HeikoTheissen: > Yes, having an offset field with variable values and a constant Common.Text value seems a bit weird.

But at client-side function (`Apply`) may be possible.

vocabularies/UI.xml
```diff
@@ -1285,7 +1285,7 @@ The trend is
         </Member>
       </EnumType>
 
-      <Term Name="DisplayTimeZone" Type="Edm.String" AppliesTo="Property">
+      <Term Name="DisplayTimezoneOffset" Type="Edm.String" AppliesTo="Property">
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Validation.Pattern" String="^[+-]?[0-9][0-9]:[0-5][0-9]$" />
         <Annotation Term="Core.Description" String="Display the annotated timestamp field in this time zone" />
```
HeikoTheissen, line 1291:
```suggestion
        <Annotation Term="Core.Description" String="Display the annotated timestamp field in a time zone with this offset" />
```

HeikoTheissen (changes requested):

vocabularies/UI.xml
ralfhandl: Went back to timestamp only

vocabularies/UI.xml
```diff
@@ -1285,13 +1285,17 @@ The trend is
         </Member>
       </EnumType>
 
-      <!-- Under discussion
-        <Term Name="DisplayTimeZone" Type="?Edm.Int32?" AppliesTo="Property">
-        <Annotation Term="Core.Description">
-        <String>Contains information for time- or date-time-fields in which time zone the time value should be displayed.</String>
+      <Term Name="DisplayTimezoneOffset" Type="Edm.String" AppliesTo="Property">
+        <Annotation Term="Common.Experimental" />
```
HeikoTheissen, line 1289:
```suggestion
        <Annotation Term="Common.Experimental" />
        <Annotation Term="Core.RequiresType" String="Edm.DateTimeOffset" />
```
```

**Review by HeikoTheissen — CHANGES_REQUESTED** [review]

File: vocabularies/UI.xml [hunk 7]
```diff
@@ -1285,13 +1285,17 @@ The trend is
         </Member>
       </EnumType>
 
-      <!-- Under discussion
-        <Term Name="DisplayTimeZone" Type="?Edm.Int32?" AppliesTo="Property">
-        <Annotation Term="Core.Description">
-        <String>Contains information for time- or date-time-fields in which time zone the time value should be displayed.</String>
+      <Term Name="DisplayTimezoneOffset" Type="Edm.String" AppliesTo="Property">
```
HeikoTheissen, line 1288:
```suggestion
      <Term Name="DisplayTimezoneOffset" Type="Edm.String" Nullable="false" AppliesTo="Property">

```
Or is this really nullable?

HeikoTheissen (changes requested):

vocabularies/UI.xml
ralfhandl: What if the timestamp field is null? Do we still require a non-null offset?

HeikoTheissen (changes requested):
Bring back the RequiresType (see my previous suggestion).

HeikoTheissen (commented):
Are you sure you want to include `lib/transform.js` and `package-lock.json` in this pull request?

ralfhandl:
If you approve #85, I'll update this branch and they will vanish 😄

HeikoTheissen (approved):

vocabularies/UI.xml
```diff
@@ -1285,13 +1285,18 @@ The trend is
...
```
HeikoTheissen, line 1288:
```suggestion
      <Term Name="DisplayTimezoneOffset" Type="Edm.String" AppliesTo="Property Parameter">

```

HeikoTheissen (changes requested):
Sorry, forgot that: should also apply to parameters, not only properties. But then the Path can only be to another parameter, which does not help us.

HeikoTheissen (changes requested):
Browsers can convert UTC into IANA timezones using {{Intl.DateTimeFormat}}. Therefore a {{DisplayTimezoneOffset}} may not be needed at all.

ralfhandl:
Closed in favor of #68

Issue #91, opened 2021-02-15
Terms in Analytics vocabulary should apply to entity set instead of type


HeikoTheissen:
https://github.com/SAP/odata-vocabularies/blob/cad0f8411978c4576da810d69362b0104bb702dd/vocabularies/Analytics.xml#L79

This term should depend on the container, not on the schema. Set `AppliesTo="EntitySet Collection"` and tag with `Core.AppliesViaContainer`.

Similarly, the use of `Analytics.Dimension`, `Analytics.Measure` and `Analytics.AccumulativeMeasure` should be discouraged in favor of properties in a new term `Analytics.ApplySupported` (to be designed).

gregorwolf:
Hello @HeikoTheissen and @ralfhandl ,

today I've learned that when I add the annotation:

```
@Aggregation.ApplySupported.Transformations      : true
```

to my sample project [sap-support-message-reporting](https://github.com/gregorwolf/sap-support-message-reporting/commit/1a9048606dc0bae210112a463b2ca0819843ccf8) it does some magic and makes the Fiori Application Generator work with this local CAP Project:

![Screenshot 2022-06-10 at 14 34 11](https://user-images.githubusercontent.com/2639535/173067891-e2414e79-abd5-41ac-9e30-3d086edb5639.png)

This issue is beside https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Aggregation.V1.md#applysupportedtype-applysupportedbase the only place where something can be found about the ApplySupported annotation. I hope this comment might help others. I know it might not be directly related to the issue.

HeikoTheissen:
@gregorwolf, the `ApplySupported` annotation is documented [here](https://github.com/oasis-tcs/odata-vocabularies/blob/aggregation-cs03-release-candidate/vocabularies/Org.OData.Aggregation.V1.md#ApplySupported).

I close this issue now, because we decided to keep the `Analytics` terms targeting the entity type. See the closed pull request linked to this issue.

ralfhandl:
Hi @gregorwolf,

`@Aggregation.ApplySupported` is mentioned in the UI5 documentation:
* [Configuring Charts](https://sapui5.hana.ondemand.com/sdk#/topic/653ed0f4f0d743dbb33ace4f68886c4e)
[...]

PR #93, opened 2021-02-22
Analytics terms shall not apply to schema

HeikoTheissen:
- [x] Merge oasis-tcs/odata-vocabularies#101

#91


HeikoTheissen:
Must not be merged before oasis-tcs/odata-vocabularies#101

GeraldKrause (dismissed):
Looks good to me.

ralfhandl (dismissed):

vocabularies/Analytics.xml
```diff
@@ -76,7 +100,8 @@
         </Annotation>
       </Term>
 
-      <Term Name="AggregatedProperties" Type="Collection(Analytics.AggregatedPropertyType)" Nullable="false" AppliesTo="EntityType">
+      <Term Name="AggregatedProperties" Type="Collection(Analytics.AggregatedPropertyType)" Nullable="false" AppliesTo="EntitySet Collection">
```
GeraldKrause, line 103: The alias names introduced with an AggregatedProperties annotation are subsequently used in other annotations. One major use case is to refer to them in UI.CHart annotations. Since Chart annotations are applied to entity types, term AggregatedProperties must also be applicable to entity types. The alternative would be to apply UI annotations also on entity types, which we do not want. Therefore, I see need to keep the AppliesTo attribute of this term as it is today.

GeraldKrause (dismissed):
see my comment in the code.

vocabularies/Analytics.xml
HeikoTheissen: > Since Chart annotations are applied to entity types, term AggregatedProperties must also be applicable to entity types.

I don't understand how the one follows from the other. A `PropertyPath` in a `UI.Chart` annotation is evaluated at runtime, and if the entity instance contains a (dynamic) property with the given name, it succeeds. The property need _not_ be contained a priori in the entity type that carries the `UI.Chart` annotation (it never does, it's a dynamic property anyway).

Therefore, if I want to allow a certain dynamic property in a certain entity _set_, I annotate that entity set with `AggregatedProperties/*/Name` = name of the dynamic property. Then this dynamic property can be created on instances of that entity set and the `UI.Chart` annotation can evaluate it. There is no need to allow the dynamic entity on all instances of the entity _type_.

Makes sense?

ralfhandl: Property paths are model paths and traverse the statically known model described by CSDL. You cannot use dynamic properties in model paths.

HeikoTheissen: I thought the purpose of `Analytics.AggregatedProperties` was defining dynamic properties that can then be used elsewhere in a `PropertyPath` (e.g., in `UI.Chart/Measures`).

ralfhandl: I think that is not possible. We can use dynamic properties in instance paths in `Path` expressions, but not in model paths.

vocabularies/Analytics.xml
```diff
@@ -76,7 +100,8 @@
...
+        <Annotation Term="Core.AppliesViaContainer"/>
         <Annotation Term="Core.Description" String="Dynamic properties for aggregate expressions with specified aggregation method defined on the annotated entity type." />
         <Annotation Term="Core.LongDescription">
```
HeikoTheissen, line 106: This long description is meant to make it possible.

vocabularies/Analytics.xml
HeikoTheissen: See the long description (and my comment) below.

HeikoTheissen: The long description wants to make it possible:

> Other annotations may refer in property paths to dynamic properties declared in any AgrgegatedProperties annotation of the given entity type to leverage the results of the aggregate expression determined in the context of an entity collection of the annotated type.

GeraldKrause: I reply to both comments above:
1. re: Heiko: I see this as a conceptual gap / asymmetry in definition
  A Chart annotation is applied to a type and thus available for all sets of that type. 
  The AggregatedProperties term is an instruction how to create a dynamic property with an aggregated value that is 
  then visualized in the chart. If this instruction is not available for some sets of a type with a chart annotation, then the   
  chart definition has a gap that cannot be filled.
2. re: Ralf: I see this as a contradiction to the results we have gained in previous discussions
  See [this mail conversation](https://sap-my.sharepoint.com/:u:/p/gerald_krause/EYgfvhtEO9BKu2kBtD2qi_EBBKwoyV4aUeaBAMfwQXEvVA?e=vlut3k). If we now say that a property path must not point to a dynamic property, then the whole concept of the AggregateProperties term becomes worthless and we have to think about a new one to support the requirement to describe a chart, a data point, and other UI elements with aggregated values. This would for example also mean that we couldnt reference a custom aggregate in a chart annotation, if this custom aggregate does not have a structural property with same name.

As a final thought, looking at the purpose of this PR to move capability terms to instances (sets), I come to the conclusion that we confused AggregatedProperty here, because it is not a capability.

HeikoTheissen: > We can use dynamic properties in instance paths in `Path` expressions, but not in model paths.

Then `UI.ChartDefinitionType/Measures` and `Dimensions` should have been `Collection(Edm.PrimitiveType)` so that a `Path` can be applied to them.

GeraldKrause: > Then `UI.ChartDefinitionType/Measures` and `Dimensions` should have been `Collection(Edm.PrimitiveType)` so that a `Path` can be applied to them.

I don't think that this would fit, because Edm.Path holds an instance path and _resolves to the value identified by the path_. This is not what we need for UI.Chart and the other relevant UI terms, which make a reference to the (dynamic) property, not to its value. Then it would need to rather become Edm.String IMO.

HeikoTheissen: Then I suggest #100 instead.

GeraldKrause: I'll wait for Friday's meeting to let you present this PR, but from what I can see it does not yet cover all cases. Dynamic properties with aggregated values my be further referenced from UI.PresentationVariant/Total, .../RequestAtLeast, .../SortOrder/Property, UI.SelectionVariant/SelectOptions/PropertyName

HeikoTheissen: I think all these can also be expressed using `$Path` elements:

`UI.PresentationVariant/Total` - deprecate in favor of `UI.PresentationVariant/TotalValue` of type `Edm.PrimitiveType`

`UI.PresentationVariant/RequestAtLeast` - deprecate in favor of `UI.PresentationVariant/RequestAtLeastValue` of type `Edm.PrimitiveType`

`UI.PresentationVariant/SortOrder/Property` - deprecate of favor of `UI.PresentationVariant/SortOrder/Key` of type `Edm.PrimitiveType`

`UI.PresentationVariant/SelectOptions/Property` - deprecate of favor of `UI.PresentationVariant/SelectOptions/Value` of type `Edm.PrimitiveType` or, more radically, in favor of `UI.SelectionVariant/Filter` of type `Edm.Boolean` that takes a dynamic expression like
```json
"@UI.SelectionVariant": {"Filter": {"$Eq": [
  {"$Path": "CompanyCode"},
  "0001"
]}}
```

GeraldKrause: I cannot see how Edm.Path could be eomployed for our purpose. The CSDL spec clearly states that it resolves to the addressed value and not to the element. For all usages I mentioned above we need elements.

CDSL spec: "An instance path is used within a Value Path expression to traverse a graph of type instances and resolves to the value identified by the path."

HeikoTheissen: Edm.Path allows annotations to be used in chart definitions:
https://github.com/SAP/odata-vocabularies/blob/10e562348941e8cbdb3b074a2f3c0d9a84ff880c/examples/UI.Chart-sample.xml#L37-L40
Such a chart definition remains valid even if the name of the currency property changes (and we had such changes in OData services generated by AE).

HeikoTheissen: Will be dealt with in separate PR.

ralfhandl (approved):

ralfhandl (approved):

HeikoTheissen:
Closed in favor of #104.

PR #98, opened 2021-03-01
Normalize line-endings for XML files

ralfhandl:


.gitattributes
```diff
@@ -1,3 +1,4 @@
 * text=auto
+*.xml text
```
HeikoTheissen, line 2: Help me understand: How will the system behaviour change because of this PR?

ralfhandl: XML files will be checked out with CRLF on Windows machines.

Currently we have a mix of CRLF and LF, for example Common.xml is with LF in some branches/PRs which makes updating from master a bit painful

HeikoTheissen: But that means if the same file is committed first from a Windows machine and later from a Linux machine, then the second commit affects the whole file (because of the different line endings)?

ralfhandl: No. In the repo all text files are stored with LF only. 

Git checks out text files on Windows machines with CRLF, and normalizes them to LF on checkin.

HeikoTheissen (approved):

PR #100, opened 2021-03-03
deprecate UI.Chart/Dimensions and Measures

HeikoTheissen:
This is meant to resolve the problem with `AggregatedProperties` in #93.


erlethor:
I guess that means that the client libraries (e.g. UI5) need to support both, the old and new way, as long as old backends are used by the customer. Just mentioning it. I am all for improving things 👍🏻 in the vocs

BerSie:
As @erlethor already mentioned this will cause additional effort for all clients and will lead to much more complex coding as all variants has to be supported. So it would be really good to know what are the benefits for these changes - can you pls. elaborate on that?

HeikoTheissen:
Closed in favor of #104.

PR #104, opened 2021-03-10
AggregatedProperty of type Edm.AnnotationPath

HeikoTheissen:
Closes a remaining issue from SAP/odata-vocabularies#93.

### DIFF
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -1447,9 +1447,16 @@
         },
         "SortOrderType": {
             "$Kind": "ComplexType",
+            "@Core.Description": "Either `Property` or `DynamicProperty` must be present",
             "Property": {
                 "$Type": "Edm.PropertyPath",
+                "$Nullable": true,
                 "@Core.Description": "Sort property"
+            },
+            "DynamicProperty": {
+                "$Type": "Edm.AnnotationPath",
+                "$Nullable": true,
+                "@Core.Description": "Dynamic property defined by an annotation and used as sort property"
             },
             "Descending": {
                 "$Type": "Edm.Boolean",
--- a/vocabularies/Common.xml
+++ b/vocabularies/Common.xml
@@ -1229,8 +1229,12 @@ Upon modification field control logic is invoked so that meta-information like h
         </Annotation>
       </Term>
       <ComplexType Name="SortOrderType">
-        <Property Name="Property" Type="Edm.PropertyPath" Nullable="false">
+        <Annotation Term="Core.Description" String="Either `Property` or `DynamicProperty` must be present" />
+        <Property Name="Property" Type="Edm.PropertyPath">
           <Annotation Term="Core.Description" String="Sort property" />
+        </Property>
+        <Property Name="DynamicProperty" Type="Edm.AnnotationPath">
+          <Annotation Term="Core.Description" String="Dynamic property defined by an annotation and used as sort property" />
         </Property>
         <Property Name="Descending" Type="Edm.Boolean" Nullable="true">
           <Annotation Term="Core.Description" String="Sort direction, ascending if not specified otherwise" />
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -688,6 +688,9 @@ The trend is
         <Property Name="Measures" Type="Collection(Edm.PropertyPath)" Nullable="false">
           <Annotation Term="Core.Description" String="Measures of the chart, e.g. size and color in a bubble chart" />
         </Property>
+        <Property Name="DynamicMeasures" Type="Collection(Edm.AnnotationPath)" Nullable="false">
+          <Annotation Term="Core.Description" String="Dynamic properties introduced by annotations and used as measures of the chart" />
+        </Property>
         <Property Name="MeasureAttributes" Type="Collection(UI.ChartMeasureAttributeType)" Nullable="false">
           <Annotation Term="Core.Description">
             <String>Describes Attributes for Measures. All Measures used in this collection must also be part of the Measures Property.</String>
--- a/vocabularies/Common.md
+++ b/vocabularies/Common.md
@@ -96,14 +96,14 @@ Term|Type|Description
 [FilterDefaultValueHigh](Common.xml#L1194) *([Experimental](Common.md#Experimental))*|PrimitiveType?|<a name="FilterDefaultValueHigh"></a>A default upper limit for the property to be used in 'less than or equal' filter expressions.
 [DerivedFilterDefaultValue](Common.xml#L1199) *([Experimental](Common.md#Experimental))*|String?|<a name="DerivedFilterDefaultValue"></a>Function import to derive a default value for the property from a given context in order to use it in filter expressions.<p>Function import has two parameters of complex types:<br/> - `parameters`, a structure resembling the entity type the parameter entity set related to the entity set of the annotated property<br/> - `properties`, a structure resembling the type of the entity set of the annotated property<br/> The return type must be of the same type as the annotated property.<br/> Arguments passed to the function import are used as context for deriving the default value. The function import returns this default value, or null in case such a value could not be determined.</p>
 [SortOrder](Common.xml#L1223)|\[[SortOrderType](#SortOrderType)\]|<a name="SortOrder"></a>List of sort criteria<p>The items of the annotated entity set or the items of the collection of the annotated entity type are sorted by the first entry of the SortOrder collection. Items with same value for this first sort criteria are sorted by the second entry of the SortOrder collection, and so on.</p>
-[RecursiveHierarchy](Common.xml#L1255)|[RecursiveHierarchyType?](#RecursiveHierarchyType)|<a name="RecursiveHierarchy"></a>Defines a recursive hierarchy.
-[CreatedAt](Common.xml#L1287)|DateTimeOffset?|<a name="CreatedAt"></a>Creation timestamp
-[CreatedBy](Common.xml#L1291)|[UserID?](#UserID)|<a name="CreatedBy"></a>First editor
-[ChangedAt](Common.xml#L1295)|DateTimeOffset?|<a name="ChangedAt"></a>Last modification timestamp
-[ChangedBy](Common.xml#L1299)|[UserID?](#UserID)|<a name="ChangedBy"></a>Last editor
-[OriginalProtocolVersion](Common.xml#L1311)|String?|<a name="OriginalProtocolVersion"></a>Original protocol version of a converted (V4) CSDL document, allowed values `2.0` and `3.0`
-[ApplyMultiUnitBehaviorForSortingAndFiltering](Common.xml#L1316) *([Experimental](Common.md#Experimental))*|[Tag](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#Tag)|<a name="ApplyMultiUnitBehaviorForSortingAndFiltering"></a>Sorting and filtering of amounts in multiple currencies needs special consideration<p>TODO: add link to UX documentation on https://experience.sap.com/fiori-design/</p>
-[mediaUploadLink](Common.xml#L1322) *([Experimental](Common.md#Experimental))*|URL|<a name="mediaUploadLink"></a>URL for uploading new media content to a Document Management Service<p>In contrast to the `@odata.mediaEditLink` this URL allows to upload new media content without directly changing a stream property. The upload request typically uses HTTP POST with `Content-Type: multipart/form-data` following RFC 7578. The upload request must contain one multipart representing the content of the file. The `name` parameter in the `Content-Disposition` header (as described in RFC 7578) is irrelevant, but the `filename` parameter is expected. If the request succeeds the response will contain a JSON body of `Content-Type: application/json` with a JSON property `readLink`. The newly uploaded media resource can be linked to the stream property by changing the `@odata.mediaReadLink` to the value of this `readLink` in a subsequent PATCH request to the OData entity.</p>
+[RecursiveHierarchy](Common.xml#L1259)|[RecursiveHierarchyType?](#RecursiveHierarchyType)|<a name="RecursiveHierarchy"></a>Defines a recursive hierarchy.
+[CreatedAt](Common.xml#L1291)|DateTimeOffset?|<a name="CreatedAt"></a>Creation timestamp
+[CreatedBy](Common.xml#L1295)|[UserID?](#UserID)|<a name="CreatedBy"></a>First editor
+[ChangedAt](Common.xml#L1299)|DateTimeOffset?|<a name="ChangedAt"></a>Last modification timestamp
+[ChangedBy](Common.xml#L1303)|[UserID?](#UserID)|<a name="ChangedBy"></a>Last editor
+[OriginalProtocolVersion](Common.xml#L1315)|String?|<a name="OriginalProtocolVersion"></a>Original protocol version of a converted (V4) CSDL document, allowed values `2.0` and `3.0`
+[ApplyMultiUnitBehaviorForSortingAndFiltering](Common.xml#L1320) *([Experimental](Common.md#Experimental))*|[Tag](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#Tag)|<a name="ApplyMultiUnitBehaviorForSortingAndFiltering"></a>Sorting and filtering of amounts in multiple currencies needs special consideration<p>TODO: add link to UX documentation on https://experience.sap.com/fiori-design/</p>
+[mediaUploadLink](Common.xml#L1326) *([Experimental](Common.md#Experimental))*|URL|<a name="mediaUploadLink"></a>URL for uploading new media content to a Document Management Service<p>In contrast to the `@odata.mediaEditLink` this URL allows to upload new media content without directly changing a stream property. The upload request typically uses HTTP POST with `Content-Type: multipart/form-data` following RFC 7578. The upload request must contain one multipart representing the content of the file. The `name` parameter in the `Content-Disposition` header (as described in RFC 7578) is irrelevant, but the `filename` parameter is expected. If the request succeeds the response will contain a JSON body of `Content-Type: application/json` with a JSON property `readLink`. The newly uploaded media resource can be linked to the stream property by changing the `@odata.mediaReadLink` to the value of this `readLink` in a subsequent PATCH request to the OData entity.</p>
 
 ## <a name="TextFormatType"></a>[TextFormatType](Common.xml#L107)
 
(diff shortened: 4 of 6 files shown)


HeikoTheissen:
Other annotations to be covered:
UI.PresentationVariant/Total, .../RequestAtLeast, UI.SelectionVariant/SelectOptions/PropertyName

vocabularies/UI.xml
```diff
@@ -688,6 +694,15 @@ The trend is
         <Property Name="Measures" Type="Collection(Edm.PropertyPath)" Nullable="false">
           <Annotation Term="Core.Description" String="Measures of the chart, e.g. size and color in a bubble chart" />
         </Property>
+        <Property Name="DynamicMeasures" Type="Collection(Edm.AnnotationPath)" Nullable="false">
```
HeikoTheissen, line 697: I assume we do not want `DynamicDimensions`.

ralfhandl: Compute the year from a date, then group by year. 

Would that be a dynamic dimension?

HeikoTheissen: > Would that be a dynamic dimension?

Yes, but have we got a term that allows us to express such a thing? Also, grouping by such computed dimensions is discouraged for performance reasons.

examples/DynamicProperties-sample.xml
```diff
@@ -0,0 +1,93 @@
...
+              </Collection>
+            </PropertyValue>
+            <PropertyValue Property="Measures">
+              <Collection>
+                <PropertyPath>NumberOfItems</PropertyPath>
+              </Collection>
+            </PropertyValue>
+            <PropertyValue Property="DynamicMeasures">
+              <Collection>
+                <AnnotationPath>@Analytics.AggregatedProperties#sum/0</AnnotationPath>
```
HeikoTheissen, line 56: The [OData spec](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_AnnotationPath) says:
>A non-null path MUST resolve to an annotation.

Does that mean that it cannot resolve to a property or a collection item _within_ an annotation?

ralfhandl: Correct, to reference a property within an annotation use `Edm.PropertyPath` or `Edm.NavigationPropertyPath`

HeikoTheissen: And to cover everything, use an `Edm.ModelElementPath`. Can we even relax the existing type in `UI.Chart/Measures` from `Edm.PropertyPath` to `Edm.ModelElementPath` without breaking compatibility?

ralfhandl: > use an `Edm.ModelElementPath`

This would require a new vocabulary (version) using OData-CSDL-Version 4.01, which we so far avoided.

HeikoTheissen: Good to know this boundary condition.

vocabularies/Analytics.xml
```diff
@@ -39,14 +39,38 @@
...
 
       <Term Name="Dimension" Type="Core.Tag" BaseTerm="Aggregation.Groupable" Nullable="false" DefaultValue="true" AppliesTo="Property">
+        <Annotation Term="Core.Revisions">
+          <Collection>
+            <Record>
+              <PropertyValue Property="Kind" EnumMember="Core.RevisionKind/Deprecated"/>
+              <PropertyValue Property="Description" String="Deprecated in favor of [`AnalyticalContext/Dimension`](#AnalyticalContext)"/>
+            </Record>
+          </Collection>
+        </Annotation>
```
HeikoTheissen, line 49:
```suggestion
      <Term Name="Dimension" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property">
        <Annotation Term="Core.AppliesViaContainer"/>
```
It would be simpler to keep the property-targeted term but add the "applies via container" tag (and remove the deprecated base term?).

vocabularies/Analytics.xml
```diff
@@ -39,14 +39,38 @@
...
 
       <Term Name="Measure" Type="Core.Tag" BaseTerm="Aggregation.Aggregatable" Nullable="false" DefaultValue="true" AppliesTo="Property">
+        <Annotation Term="Core.Revisions">
+          <Collection>
+            <Record>
+              <PropertyValue Property="Kind" EnumMember="Core.RevisionKind/Deprecated"/>
+              <PropertyValue Property="Description" String="Deprecated in favor of [`AnalyticalContext/Measure`](#AnalyticalContext)"/>
+            </Record>
+          </Collection>
+        </Annotation>
```
HeikoTheissen, line 61:
```suggestion
      <Term Name="Measure" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property">
        <Annotation Term="Core.AppliesViaContainer"/>
```

vocabularies/Analytics.xml
```diff
@@ -39,14 +39,38 @@
...
 
       <Term Name="AccumulativeMeasure" BaseTerm="Analytics.Measure" Type="Core.Tag" Nullable="false" DefaultValue="true">
+        <Annotation Term="Core.Revisions">
+          <Collection>
+            <Record>
+              <PropertyValue Property="Kind" EnumMember="Core.RevisionKind/Deprecated"/>
+              <PropertyValue Property="Description" String="Deprecated in favor of [`AnalyticalContext/AccumulativeMeasure`](#AnalyticalContext)"/>
+            </Record>
+          </Collection>
+        </Annotation>
```
HeikoTheissen, line 73:
```suggestion
      <Term Name="AccumulativeMeasure" Type="Core.Tag" Nullable="false" DefaultValue="true">
        <Annotation Term="Core.AppliesViaContainer"/>
```

vocabularies/Analytics.xml
```diff
@@ -104,7 +141,29 @@
...
+          <Annotation Term="Core.Description" String="The property holds the key of a dimension" />
+        </Property>
+        <Property Name="Measure" Type="Core.Tag" Nullable="false" DefaultValue="true">
+          <Annotation Term="Core.Description" String="The property holds the numeric value of a measure" />
+        </Property>
+        <Property Name="AccumulativeMeasure" Type="Core.Tag" Nullable="false" DefaultValue="true">
+          <Annotation Term="Core.Description" String="The measure has non-negative and additive values; it can be used in whole-part charts, e.g. the Donut" />
+        </Property>
+      </ComplexType>
+
```
HeikoTheissen, line 162:
```suggestion
```

HeikoTheissen (commented):
Should we keep the property-targeted `Dimension` etc.?

vocabularies/Analytics.xml
HeikoTheissen: Current proposal is cleaner, because the deprecated and the future terms are separated.

vocabularies/Analytics.json
```diff
@@ -113,14 +131,29 @@
...
             "@Core.LongDescription": "\n            Other annotations may refer in property paths to dynamic properties declared in any AgrgegatedProperties annotation of the \n\t\t\tgiven entity type to leverage the results of the aggregate expression determined in the context of an entity collection of \n\t\t\tthe annotated type.\n          "
         },
+        "AggregatedProperty": {
+            "$Kind": "Term",
+            "$Type": "Analytics.AggregatedPropertyType",
+            "$AppliesTo": [
+                "EntityType"
+            ],
+            "@Core.Description": "Dynamic property for aggregate expression with specified aggregation method defined on the annotated entity type.",
+            "@Core.LongDescription": "A property path to an `AggregatedProperty` annotation can be used to reference the dynamic property holding the aggregated value."
```
ralfhandl, line 150: How should this work? The term declares a new dynamic property with 

~~~js
@Analytics.AggregatedProperty: {
  Name: "foo"
}
~~~

So in JSON data there will be a dynamic property `foo:42` that is not in the model and cannot be referenced via a PropertyPath.

vocabularies/Analytics.json
```diff
@@ -133,8 +166,42 @@
...
+            },
+            "AccumulativeMeasure": {
+                "$Type": "Core.Tag",
+                "$DefaultValue": true,
+                "@Core.Description": "The measure has non-negative and additive values; it can be used in whole-part charts, e.g. the Donut"
+            }
+        },
         "$Annotations": {
             "Aggregation.CustomAggregate": {
+                "@Core.LongDescription#Analytics": "A property path to an `Aggregation.CustomAggregate` annotation can be used to reference the dynamic property holding the value of the custom aggregate.",
```
ralfhandl, line 204: What's this?

Trying to inject "custom" documentation into a foreign vocabulary term?

examples/DynamicProperties-sample.xml
```diff
@@ -0,0 +1,85 @@
...
+            <PropertyValue Property="ChartType" EnumMember="UI.ChartType/Bubble"/>
+            <PropertyValue Property="Dimensions">
+              <Collection>
+                <PropertyPath>Country</PropertyPath>
+              </Collection>
+            </PropertyValue>
+            <PropertyValue Property="Measures">
+              <Collection>
+                <PropertyPath>NumberOfItems</PropertyPath>
+                <PropertyPath>@Analytics.AggregatedProperty#sum</PropertyPath>
```
ralfhandl, line 48: This property path references the complex "property" `@Analytics.AggregatedProperty#sum` of type `Analytics.AggregatedPropertyType` whose value is the constant object
~~~js
 { Name: "Total", AggregationMethod: "sum", AggregatableProperty: "Amount" }
~~~

I would not know how to use this complex type as a measure, given that its "value" is constant on all instances of the "Sales" entity type.

There seem to be some hidden assumptions here on how to extend client interpretation logic for "measures".

Please make these explicit.

ralfhandl (changes requested):

vocabularies/Analytics.json
HeikoTheissen: A `UI.Chart/Measures` will _not_ reference the dynamic property `foo`, but the annotation path:
```json
"@UI.Chart": {"Measures": ["@Analytics.AggregatedProperty"]}
```

vocabularies/Analytics.json
HeikoTheissen: Yes. Is this not allowed?

examples/DynamicProperties-sample.xml
HeikoTheissen: https://github.com/SAP/odata-vocabularies/blob/2b4eece969576df6ace263fcae47add2f857885a/vocabularies/Analytics.xml#L123-L125
Not explicit enough?

vocabularies/Analytics.json
ralfhandl: Then we have to document in `UI.Chart/Measures` that the property path can either point to 
1. a declared property of primitive type, in which case the run-time values of that property are the measure values, or to
2.  a structured "property" of type `Analytics.AggregatedPropertyType`, in which case the design-time value of that structured "property" points to a dynamic property whose run-time values are the measure values.

To my knowledge clients currently only expect #1, so adding #2 would be a breaking change.

ralfhandl (changes requested):
We cannot define here that some terms in other vocabularies need to be interpreted differently. This has to be explicitly defined in the other vocabularies, and in a compatible way for already existing terms.

vocabularies/Analytics.json
HeikoTheissen: That would mean changing the documentation in many many places. I can do that, but I would prefer explaining it at the target annotation. See https://github.com/SAP/odata-vocabularies/blob/2b4eece969576df6ace263fcae47add2f857885a/vocabularies/Analytics.xml#L123-L125

examples/DynamicProperties-sample.xml
ralfhandl: Then let's add changes to those annotations that now refer here also in this pull request and check whether the combined documentation is easy to understand.

vocabularies/Analytics.json
ralfhandl: It has no effect: do you plan to extend the markdown generator to also consult other vocabularies for "remote-injected" descriptions?

If not, better add this to some place in this vocabulary that already shows up in the markdown.

ralfhandl (changes requested):
See comment

vocabularies/Common.json
```diff
@@ -1461,7 +1461,8 @@
             "$Kind": "ComplexType",
             "Property": {
                 "$Type": "Edm.PropertyPath",
-                "@Core.Description": "Sort property"
+                "@Core.Description": "Sort property",
+                "@Core.LongDescription": "May also resolve to an annotation with the term [`Aggregation.CustomAggregate`](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.html)\n            or [`Analytics.AggregatedProperty`](Analytics.md#AggregatedPropertyType). In this case the sort property is the dynamic property\n            contributed by the aggregation that is defined by the annotation."
```
ralfhandl, line 1465: Sorry, this is a breaking change, you cannot do this.

Currently you can use these property paths to construct an `$orderby` query option, e.g. `$orderby=Size`.

This won't work with an annotation: `$orderby=@Aggregation.CustomAggregate#sum` is not what you want to achieve.

ralfhandl (changes requested):

vocabularies/Common.xml
```diff
@@ -1344,6 +1363,11 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         </Annotation>
       </Term>
 
+      <Term Name="StructuralPropertyPath" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Term Property">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="A term or term property with this tag whose type is (a collection of) `Edm.PropertyPath` MUST resolve to a structural property of the entity type that is targeted by the term" />
```
HeikoTheissen, line 1368:
```suggestion
      <Term Name="PrimitivePropertyPath" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Term Property">
        <Annotation Term="Common.Experimental" />
        <Annotation Term="Core.Description" String="A term or term property with this tag whose type is (a collection of) `Edm.PropertyPath` MUST resolve to a primitive property of the entity type that is targeted by the term" />
```
@BerSie, would you prefer this?

BerSie: Definitely - thanks! But as an additional remark - the property doesn't have to be part of the targeted entity itself, it can also be within an associated entity (using navigation properties), so maybe I would omit this last part (as you always use only path expressions that are "in scope".

examples/DynamicProperties-sample.xml
```diff
@@ -0,0 +1,89 @@
...
+            </PropertyValue>
+            <PropertyValue Property="Measures">
+              <Collection>
+                <PropertyPath>NumberOfItems</PropertyPath>
+              </Collection>
+            </PropertyValue>
+            <PropertyValue Property="DynamicMeasures">
+              <Collection>
+                <AnnotationPath>@Analytics.AggregatedProperty#sum</AnnotationPath>
+                <AnnotationPath>/self.Container/me/SalesOrders@Aggregation.CustomAggregate#WeightedAverage</AnnotationPath>
```
ralfhandl, line 53: This absolute path is nested within an annotation on the entity type "Sales", pointing to something that is only valid for one special use of this entity type.

How would this work if there are other uses of entity type "Sales", e.g. a top-level entity set for "Sales" next to the "me" singleton?

HeikoTheissen: Good observation.

The absolute path refers to an annotation, which defines a custom aggregate on "a special usage of the `Sales` entity type". For other uses of the same entity type, the custom aggregate is not defined/not allowed.

On the other hand, the `UI.Chart` annotation, of which this absolute path is a property, is defined for _all_ usages of the `Sales` entity type. I conclude that, for usages other than `me/SalesOrders`, the chart has effectively only _one_ dynamic measure: the aggregated property defined by the other annotation path.

HeikoTheissen: Suggested wording:
> If the annotation referenced by an annotation path does not apply to the same collection of entities            as the `UI.Chart` annotation, the annotation path MUST be silently ignored.

ralfhandl: Oh no, please not yet another one of those "meta" annotations.

The suggested wording is misleading because UI.Chart only applies to types, never directly to collections.

HeikoTheissen: The definition of `UI.Chart` says "Visualization of multiple data points", this is a collection. What about that?
> If the annotation referenced by an annotation path does not apply to the same collection of entities as the one being visualized according to the `UI.Chart` annotation, the annotation path MUST be silently ignored.

vocabularies/Common.json
```diff
@@ -1542,6 +1561,17 @@
...
+        "PrimitivePropertyPath": {
+            "$Kind": "Term",
+            "$Type": "Core.Tag",
+            "$DefaultValue": true,
+            "$AppliesTo": [
+                "Term",
+                "Property"
+            ],
+            "@Common.Experimental": true,
+            "@Core.Description": "A term or term property with this tag whose type is (a collection of) `Edm.PropertyPath` MUST resolve to a primitive structural property of an entity type"
```
ralfhandl, line 1574: Oh no, not another meta annotation!

ralfhandl (changes requested):

vocabularies/Common.json
HeikoTheissen: @BerSie, you asked for this restriction to "primitive property paths", would you agree if we specify it only in the descriptions rather than through a new "meta-annotation"?

BerSie: Yes, that's for me sufficient.

ralfhandl: How many other use cases do we have for this new meta-annotation, i.e. on which other terms/properties of type `Edm.PropertyPath` would we need to add this?

Would this only affect SAP vocabularies, or do we see the need for this also in OASIS vocabularies?

vocabularies/UI.xml
```diff
@@ -1071,8 +1115,22 @@ The trend is
           </Annotation>
         </Property>
         <Property Name="RequestAtLeast" Type="Collection(Edm.PropertyPath)" Nullable="false">
+          <Annotation Term="Common.PrimitivePropertyPath"/>
```
ralfhandl, line 1118: Why is this restricted to primitive properties?

ralfhandl (changes requested):

vocabularies/Common.json
ralfhandl: Ok, in that case I think it is helpful to add this meta-annotation.

vocabularies/UI.xml
HeikoTheissen: @BerSie's wish

vocabularies/Common.xml
```diff
@@ -1345,6 +1369,11 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         </Annotation>
       </Term>
 
+      <Term Name="PrimitivePropertyPath" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Term Property">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="A term or term property with this tag whose type is (a collection of) `Edm.PropertyPath` MUST resolve to a primitive structural property of an entity type" />
```
ralfhandl, line 1374:
```suggestion
        <Annotation Term="Core.Description" String="A term or term property with this tag whose type is (a collection of) `Edm.PropertyPath` MUST resolve to a primitive structural property" />
```

ralfhandl (changes requested):

vocabularies/UI.xml
BerSie: I can't remember on having commented to the RequestAtLeast property annotation ...

ralfhandl, line 1118:
```suggestion
```

ralfhandl (changes requested):

vocabularies/UI.xml
ralfhandl: The "request at least" properties go into $select, which already excludes `@Analytics.AggregatableProperty#sum`. 

$select also allows complex properties, and it is not clear from the description of "request at least" that this should not be allowed.

ralfhandl (changes requested):

vocabularies/UI.xml
HeikoTheissen: Sorry, my fault.

ralfhandl (approved):

vocabularies/UI.xml
```diff
@@ -1073,6 +1117,19 @@ The trend is
         <Property Name="RequestAtLeast" Type="Collection(Edm.PropertyPath)" Nullable="false">
           <Annotation Term="Core.Description" String="Properties that should always be included in the result of the queried collection" />
         </Property>
+        <Property Name="RequestAtLeastDynamic" Type="Collection(Edm.AnnotationPath)" Nullable="false">
+          <Annotation Term="Core.Description" String="Dynamic properties introduced by annotations that should always be included in the result of the queried collection" />
```
ralfhandl, line 1121: How would that work? Could you give a request example?

vocabularies/UI.xml
```diff
@@ -1142,8 +1199,31 @@ The trend is
...
+            <Collection>
+              <Record>
+                <PropertyValue Property="Kind" EnumMember="Core.RevisionKind/Modified"/>
+                <PropertyValue Property="Description" String="Now nullable if `DynamicPropertyName` is present"/>
+              </Record>
+            </Collection>
+          </Annotation>
+        </Property>
+        <Property Name="DynamicPropertyName" Type="Edm.AnnotationPath">
+          <Annotation Term="Core.Description" String="Dynamic property introduced by annotations for which value ranges are specified" />
```
ralfhandl, line 1216: How would filtering work on dynamic properties? As far as I know our services don't (fully) support dynamic properties.

ralfhandl (changes requested):

vocabularies/UI.xml
HeikoTheissen: If the dynamic property comes from an aggregation, it shall be requested through an `$apply=aggregate(...)` clause.

vocabularies/UI.xml
HeikoTheissen: `$apply=groupby(...,aggregate(... as dynprop))/filter(dynprop lt 1000)`

ralfhandl: An example would be helpful here: which combination of annotations will trigger the client to create this non-trivial request?

vocabularies/UI.xml
ralfhandl: Where do we describe the client logic necessary to construct such a request?

vocabularies/UI.xml
HeikoTheissen: extended example accordingly

vocabularies/UI.xml
HeikoTheissen: To be discussed in Albatross first.

uhlmannm: Albatross: 
- At the moment no use case for `RequestAtLeastDynamic` is known. Hence skipping it for the moment.
- The explanation for `RequestAtLeast` that it translates into `$select` (no `$apply`) or `groupby` (in `$apply`) is missing.

vescalier:
UI vocabulary changes look good to me.

ralfhandl (approved):

PR #106, opened 2021-03-17
Graph.traceId: new instance annotation for error messages

ralfhandl:
- add instance annotations used in SAP Graph error responses
- remove obsolete term `links`


vocabularies/Graph.xml
```diff
@@ -33,39 +36,27 @@
...
+      </Term>
+      <ComplexType Name="DetailsType">
         <Annotation Term="Common.Experimental" />
-        <Property Name="source" Type="Edm.PropertyPath">
-          <Annotation Term="Core.Description" String="Path to primitive property in the annotated entity type" />
+        <Annotation Term="Core.Description" String="Graph-specific details for error responses" />
+        <Property Name="url" Type="Edm.String">
+          <Annotation Term="Common.Experimental" />
+          <Annotation Term="Core.Description" String="URL sent to the business system tenant" />
         </Property>
```
HeikoTheissen, line 56:
```suggestion
        <Property Name="url" Type="Edm.String">
          <Annotation Term="Common.Experimental" />
          <Annotation Term="Core.IsURL" />
          <Annotation Term="Core.Description" String="URL sent to the business system tenant" />
        </Property>
```

vocabularies/Graph.xml
```diff
@@ -33,39 +36,27 @@
...
-        </Property>
-        <Property Name="filter" Type="Collection(Graph.binding)" Nullable="false">
-          <Annotation Term="Core.Description" String="Map of single-valued property paths in the target entity to property paths in the annotated entity type" />
-          <Annotation Term="Core.LongDescription" String="Used to construct a `$filter` expression and triggers the `Read` (list) operation on the target entity set" />
-        </Property>
-      </ComplexType>
-
-      <ComplexType Name="binding">
+        <Annotation Term="Common.IsInstanceAnnotation" />
+        <Annotation Term="Core.Description" String="Graph-specific details for error responses" />
```
HeikoTheissen, line 48: If it is only for error responses, shouldn't the term be called `ErrorDetails`?

HeikoTheissen (changes requested):

vocabularies/Graph.xml
ralfhandl: Given that it is directly within the `error` object of an OData error response, that would look redundant. Now it is:
~~~jsonc
"error": {
  "@Graph.Details": {
     // ...
  }
}
~~~

HeikoTheissen (approved):

PR #109, opened 2021-04-15
fix: accidental nullability of terms

ralfhandl:


vocabularies/Common.xml
```diff
@@ -47,48 +48,48 @@
...
         <Annotation Term="Core.Description" String="1 for first version of a service, incremented when schema changes incompatibly and service is published with a different URI" />
       </Term>
-      <Term Name="ServiceSchemaVersion" Type="Edm.Int32" AppliesTo="Schema">
+      <Term Name="ServiceSchemaVersion" Type="Edm.Int32" Nullable="false" AppliesTo="Schema">
         <Annotation Term="Core.Description" String="0 for first schema version within a service version, incremented when schema changes compatibly" />
       </Term>
 
       <!-- Section: General Semantics -->
-      <Term Name="Label" Type="Edm.String">
+      <Term Name="Label" Type="Edm.String" Nullable="false">
```
HeikoTheissen, line 59: Terms containing a human-readable text should remain nullable, because they might be annotated using an `Edm.Path`, which may point to `null`.

vocabularies/Common.xml
```diff
@@ -113,7 +114,7 @@
         </Member>
       </EnumType>
 
-      <Term Name="Timezone" Type="Edm.String" AppliesTo="Property Parameter">
+      <Term Name="Timezone" Type="Edm.String" Nullable="false" AppliesTo="Property Parameter">
```
HeikoTheissen, line 117:
```suggestion
      <Term Name="Timezone" Type="Edm.String" AppliesTo="Property Parameter">
```
Shall remain nullable, since it may be dynamic via `Edm.Path`.

vocabularies/Common.xml
```diff
@@ -188,12 +189,12 @@
         <Annotation Term="Core.Description" String="Annotated property or parameter is a unit of measure" />
       </Term>
 
-      <Term Name="UnitSpecificScale" Type="Edm.PrimitiveType" AppliesTo="Property">
+      <Term Name="UnitSpecificScale" Type="Edm.PrimitiveType" Nullable="false" AppliesTo="Property">
```
HeikoTheissen, line 192:
```suggestion
      <Term Name="UnitSpecificScale" Type="Edm.PrimitiveType" AppliesTo="Property">
```
Shall remain nullable, since it may be dynamic via `Edm.Path`. Same for the next three.

HeikoTheissen (changes requested):
I didn't look through all, but in many cases, a term should be nullable, because it may be computed via an `Edm.Path`, and that may point to `null`. Only if a null value makes absolutely no sense would `Nullable="false"` be justified. But in such cases, what happens if the annotation is absent? Mustn't there be a default anyway?

vocabularies/Common.xml
ralfhandl: Then we should describe the desired behavior if the time zone is null: assume UTC?

vocabularies/Common.xml
ralfhandl: Then we should describe how `null` is to be treated: as zero?

I'd rather require that the path points to a non-nullable property.

vocabularies/Common.xml
ralfhandl: @BerSie How would the UI deal with a label that is null?

vocabularies/Common.xml
HeikoTheissen: We have that already, see the LongDescription:
> In the absence of this annotation, points in time are typically interpreted in the current user's or default time zone.

Can be rephrased:
> If this annotation is absent or null, points in time ...

HeikoTheissen (changes requested):

vocabularies/Common.xml
HeikoTheissen: If you think that null makes no sense, then yes.

vocabularies/Common.xml
BerSie: With an empty label (I would expect and hope for it)

vocabularies/Common.xml
ralfhandl: Yes, makes no sense.

vocabularies/Common.xml
ralfhandl: @BerSie so we can set it explicitly to nullable=true?

HeikoTheissen: But `Common.Text` is typically a property, this must definitively remain nullable.

vocabularies/Common.xml
ralfhandl: Adapted text as proposed, set explicitly to nullable

vocabularies/Common.xml
BerSie: I would like to test it before saying anything wrong ... :-)

ralfhandl: > But `Common.Text` is typically a property, this must definitively remain nullable.

CHAR/STRING properties in ABAP are never nullable, so in that prominent use case Common.Text never is null.

vocabularies/Common.xml
```diff
@@ -113,18 +114,18 @@
         </Member>
       </EnumType>
 
-      <Term Name="Timezone" Type="Edm.String" AppliesTo="Property Parameter">
+      <Term Name="Timezone" Type="Edm.String" Nullable="true" AppliesTo="Property Parameter">
```
HeikoTheissen, line 117: Why do you explicitly set the default value?

vocabularies/Common.xml
```diff
@@ -417,17 +418,17 @@ Example for dynamic use: in a travel expense report the property `DestinationCou
           <Annotation Term="Core.Description" String="..." />
         </Property>
       </ComplexType>
-      <Term Name="Timestamp" Type="Edm.DateTimeOffset">
+      <Term Name="Timestamp" Type="Edm.DateTimeOffset" Nullable="false">
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="..." />
         <Annotation Term="Common.IsInstanceAnnotation" />
       </Term>
```
HeikoTheissen, line 425: Aren't instance annotations nullable by nature, because you can never assume the server generates them?

vocabularies/Common.xml
```diff
@@ -1294,19 +1295,19 @@ Upon modification field control logic is invoked so that meta-information like h
 
       <!-- Metadata annotations that can also appear as instance annotations -->
 
-      <Term Name="CreatedAt" Type="Edm.DateTimeOffset" AppliesTo="EntityType">
+      <Term Name="CreatedAt" Type="Edm.DateTimeOffset" Nullable="false" AppliesTo="EntityType">
```
HeikoTheissen, line 1298:
```suggestion
      <Term Name="CreatedAt" Type="Edm.DateTimeOffset" AppliesTo="EntityType">
```
In the stateful SD OData service, before an instance is saved, it has CreatedAt = null.

vocabularies/Common.xml
ralfhandl: To express that we mean it to be nullable and haven't just forgotten to say it, as in the other cases.

Also allows to `grep` for potential errors without false positives.

vocabularies/Common.xml
ralfhandl: Absence of an instance annotation means it is `undefined`, which is different from `null`.

vocabularies/Common.xml
ralfhandl: Ok, explicitly nullable now

vocabularies/Common.xml
```diff
@@ -657,7 +658,7 @@ The minimum structure is
         </Annotation>
       </TypeDefinition>
 
-      <Term Name="ValueListRelevantQualifiers" Type="Collection(Common.SimpleIdentifier)">
+      <Term Name="ValueListRelevantQualifiers" Type="Collection(Common.SimpleIdentifier)" Nullable="false">
```
HeikoTheissen, line 661: Don't I want to use null to address the unqualified value list annotation?

ralfhandl: We do not specify this yet: could be `null`, could be the empty string. Given that some backends have problems representing `null`, the empty string might be a better choice.

@BerSie: what is Fiori expecting?

BerSie: This knows @uhlmannm better - it's evaluated in the OData model. I pesonally would expect the empty string.

vocabularies/Common.xml
BerSie: Unfortunately it's not handled like I expected (don't ask me what the result is ... it left me speechless). I personally would stay for Common.Label with Nullable="false". For Common.Text I'm not sure if we can/should prevent the referenced property to be null (for non-ABAP platforms).

ralfhandl: Done.

vocabularies/Common.xml
ralfhandl: @ThomasChadzelek: can you help with this?

vocabularies/Common.xml
```diff
@@ -47,48 +48,48 @@
...
 
-      <Term Name="TextFor" Type="Edm.PropertyPath" AppliesTo="Property">
+      <Term Name="TextFor" Type="Edm.PropertyPath" Nullable="false" AppliesTo="Property">
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.RequiresType" String="Edm.String" />
         <Annotation Term="Core.Description" String="The annotated property contains a descriptive text for values of the referenced property." />
       </Term>
 
-      <Term Name="ExternalID" Type="Edm.String" AppliesTo="Property">
+      <Term Name="ExternalID" Type="Edm.String" Nullable="false" AppliesTo="Property">
```
HeikoTheissen, line 85:
```suggestion
      <Term Name="ExternalID" Type="Edm.String" Nullable="true" AppliesTo="Property">
```
This should behave like `Text`, i.e., be nullable.

vocabularies/Common.xml
```diff
@@ -47,48 +48,48 @@
...
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="A human readable identifier for values of the annotated property. Value MUST be a dynamic expression when used as metadata annotation." />
         <Annotation Term="Core.LongDescription" String="If the annotated property is (part of) a foreign key of a resource, the external id is a human readable (part of an) identifier of this resource. There is a one-to-one relationship between each possible value of the annotated property and the corresponding external id." />
       </Term>
 
       <!--
-      <Term Name="ExternalIDs" Type="Collection(Edm.String)" AppliesTo="Property">
+      <Term Name="ExternalIDs" Type="Collection(Edm.String)" Nullable="false" AppliesTo="Property">
```
HeikoTheissen, line 92:
```suggestion
      <Term Name="ExternalIDs" Type="Collection(Edm.String)" Nullable="true" AppliesTo="Property">
```
This should behave like `Text`, i.e., be nullable.

vocabularies/Common.xml
```diff
@@ -100,7 +101,7 @@
         <Annotation Term="Core.Description" String="An identifier to distinguish multiple texts in different languages for the same entity" />
       </Term>
 
-      <Term Name="TextFormat" Type="Common.TextFormatType" AppliesTo="Property Parameter ReturnType">
+      <Term Name="TextFormat" Type="Common.TextFormatType" Nullable="false" AppliesTo="Property Parameter ReturnType">
```
HeikoTheissen, line 104:
```suggestion
      <Term Name="TextFormat" Type="Common.TextFormatType" Nullable="true" AppliesTo="Property Parameter ReturnType">
```
This should behave like `Text`, i.e., be nullable.

HeikoTheissen (changes requested):

vocabularies/Common.xml
ralfhandl: Nullable IDs are a bit strange: what does it mean if two different instances both have `null` as their external ID?

vocabularies/Common.xml
ralfhandl: TextFormat is an enumeration with two values: `plain` or `html`. What should `null` mean?

vocabularies/Common.xml
HeikoTheissen: Not every instance must have an external ID, just like not every SAP unit of measurement has a UNCEFACT representation.

BerSie: An external ID is something that is for the end user an id of an object instance - this should be something, that the user can read and that's unique. Everything else seems to be not an ID to me.

vocabularies/Common.xml
HeikoTheissen: Then the description is wrong:
> Human-readable text that may contain formatting information.

Should be:
> Format in which human-readable text may contain formatting information.

vocabularies/Common.xml
ralfhandl: Good point: what would be visualized if the external ID is `null`, and how would that be visually distinguished from a non-null external ID?

ralfhandl: @bernhardthimmel: you "invented" the external ID: can/should it be nullable?

vocabularies/Common.xml
ralfhandl: Changed description

vocabularies/Common.xml
uhlmannm: I checked with Uwe. The value null would not make the default qualifier relevant. Nullable=false seems Ok then.

HeikoTheissen (approved):

vocabularies/Common.xml
bernhardthimmel: I know it is already resolved but just for the record, yes I also think it should be nullable. E.g. there might also be issues with permissions so that you only get a technical id but that you are not able to see the text or the external id of the related object. Some colleagues in project management mentioned such a use case.
As for visualization, good question.

BerSie (approved):
OK

PR #112, opened 2021-04-22
CSRFTokenRequired

HeikoTheissen:
example will follow

### DIFF
--- a/vocabularies/UI.json
+++ b/vocabularies/UI.json
@@ -9,6 +9,14 @@
                 }
             ]
         },
+        "https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Capabilities.V1.json": {
+            "$Include": [
+                {
+                    "$Namespace": "Org.OData.Capabilities.V1",
+                    "$Alias": "Capabilities"
+                }
+            ]
+        },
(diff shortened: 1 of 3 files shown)


vocabularies/UI.json
```diff
@@ -1928,6 +1936,40 @@
...
+            "$Collection": true,
+            "$Type": "UI.CSRFTokenRequiredType",
+            "$AppliesTo": [
+                "EntityContainer",
+                "Collection",
+                "Property"
+            ],
+            "@Core.AppliesViaContainer": true,
+            "@Core.Description": "Requests for this container, collection or property must contain a token proving that the user willingly made the request",
+            "@Core.LongDescription": "In the absence of this annotation, the token is not required for GET requests, but is required for all others."
```
ralfhandl, line 1951:
```suggestion
            "@Core.LongDescription": "In the absence of this annotation, the token is not required for GET requests, but is required for all modifying requests."
```

ralfhandl (commented):
See above

vocabularies/UI.json
HeikoTheissen: At the moment, it is required even for non-modifying POST $batch requests.

ralfhandl: "all others" is wrong, you don't need it for HEAD and OPTIONS requests

PR #115, opened 2021-06-08
Add DistinctValuesSupported on value lists

BerSie:
Adding a property DistinctValuesSupported to the Terms ValueList and ValueListMapping to indicate that the value list supports a distinct aggregation on the property in focus (in ValueListParameterInOut). This avoids having duplicate entries in the result table of the value list.


vocabularies/Common.xml
```diff
@@ -622,6 +622,9 @@ The minimum structure is
         <Property Name="CollectionRoot" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Service root of the value list collection; not specified means local to the document containing the annotation" />
         </Property>
+        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true" DefaultValue="true">
```
HeikoTheissen, line 625:
```suggestion
        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true"> 
```
A `DefaultValue` for a term property makes no sense, because it is considered only in POST und PUT payloads: [https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_DefaultValue](OData-CSDL, sec. 7.2.7)

vocabularies/Common.xml
```diff
@@ -695,6 +698,9 @@ The minimum structure is
         <Property Name="CollectionPath" Type="Edm.String" Nullable="false">
           <Annotation Term="Core.Description" String="Resource path of an OData collection with possible values, relative to the document containing the value list mapping" />
         </Property>
+        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true" DefaultValue="true">
```
HeikoTheissen, line 701:
```suggestion
        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true"> 
```

HeikoTheissen (changes requested):

vocabularies/Common.xml
ralfhandl: It makes sense, the cited specification passage says
> A primitive or enumeration property MAY define a default value that is used if the property is not explicitly represented in an annotation

However: the default value should be `false` because we cannot assume all existing value helps to suddenly support distinct values.

vocabularies/Common.xml
ralfhandl, line 701:
```suggestion
        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true" DefaultValue="false"> 
```

ralfhandl, line 701:
```suggestion
        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true" DefaultValue="false"> 
```

ralfhandl (changes requested):

vocabularies/Common.xml
ralfhandl: Keep default value and use `false`.

HeikoTheissen: If we use `DefaultValue="false"`, we don't need nullable.

HeikoTheissen: If we use `DefaultValue="false"`, we don't need nullable.

vocabularies/Common.xml
HeikoTheissen: Should have read more closely...

ralfhandl, line 625:
```suggestion
        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true" DefaultValue="false"> 
```

ralfhandl (changes requested):

vocabularies/Common.xml
```diff
@@ -622,6 +622,9 @@ The minimum structure is
         <Property Name="CollectionRoot" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Service root of the value list collection; not specified means local to the document containing the annotation" />
         </Property>
+        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true" DefaultValue="false">
```
HeikoTheissen, line 625:
```suggestion
        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" DefaultValue="false"> 
```

vocabularies/Common.xml
```diff
@@ -695,6 +698,9 @@ The minimum structure is
         <Property Name="CollectionPath" Type="Edm.String" Nullable="false">
           <Annotation Term="Core.Description" String="Resource path of an OData collection with possible values, relative to the document containing the value list mapping" />
         </Property>
+        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" Nullable="true" DefaultValue="false">
```
HeikoTheissen, line 701:
```suggestion
        <Property Name="DistinctValuesSupported" Type="Edm.Boolean" DefaultValue="false"> 
```

HeikoTheissen (changes requested):

vocabularies/Common.xml
ralfhandl: Nullable defaults to true in CSDL XML, so we have to explicitly specify it.

And we want to always explicitly specify it for properties in vocabularies so that in a few months we still know what we intended.

vocabularies/Common.xml
BerSie: Why should Nullable be false here?

ralfhandl: We have a default value of `false`, so we can omit the property in an annotation if distinct values are not supported.

I don't know what `DistinctValuesSupported:null` would mean - how would the UI react?

BerSie: Same behavior as in the "false" case. Should this be the reason to forbid the null value?

ralfhandl: Yes, we don't have tri-state logic of "yes", "no", "don't know".

BerSie: "don't know" would be "no", but I see that the Nullable false is more explicit.

HeikoTheissen (approved):

ralfhandl (approved):

PR #116, opened 2021-06-09
Suppress check of scale for measured quantities

ralfhandl:


vocabularies/CodeList.json
```diff
@@ -91,6 +91,15 @@
...
             "@Core.LongDescription": "The Configuration Deprecation Code indicates whether a code list value is valid (deprecation code is empty/space), deprecated (deprecation code `W`), or revoked (deprecation code `E`). "
+        },
+        "DoNotCheckScale": {
+            "$Kind": "Term",
+            "$Type": "Edm.Boolean",
+            "$AppliesTo": [
+                "Property"
+            ],
+            "@Core.Description": "Do not check the allowed number of fractional digits",
+            "@Core.LongDescription": "The annotated property contains a measured quantity, and the property value may have more fractional digits than defined for its unit of measure."
```
uhlmannm, line 102: ... quantity, and the user may enter more fractional digits ...

This would switch off the validation of user input with respect to decimals.

vocabularies/CodeList.xml
```diff
@@ -70,6 +70,16 @@
         </Annotation>
       </Term>
 
+      <Term Name="DoNotCheckScale" Type="Edm.Boolean" Nullable="false" AppliesTo="Property">
```
HeikoTheissen, line 73:
```suggestion
      <Term Name="DoNotCheckScale" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property">
```
Why is this not a tag?

vocabularies/CodeList.json
HeikoTheissen: So is this a UI annotation? It seems to tell the UI "do not complain about too many decimal places, just drop them silently".

If yes, this belongs in the UI vocabulary.

ralfhandl: Moved to UI

vocabularies/CodeList.xml
ralfhandl: What value does that indirection add?

HeikoTheissen: Don't know why Core.Tag was introduced but it is used in similar cases. If it does not add value, why was it introduced?

HeikoTheissen (approved):

uhlmannm (approved):

RaizoC (approved):

PR #117, opened 2021-06-18
Add Common.SideEffectsType/Discretionary

RaizoC:


vocabularies/Common.json
```diff
@@ -1356,6 +1356,13 @@
                 "@Common.Experimental": true,
                 "@Core.Description": "Indicates whether the side-effect has already happened",
                 "@Core.LongDescription": "The value of this property typically is a Path expression pointing to a boolean property. It can be used by clients to defer expensive refresh calls until they are actually needed and instead just request the referenced indicator property. Servers can choose to return indicator properties even if not explicitly requested."
+            },
+            "Discretional": {
```
ralfhandl, line 1360: Maybe better use [discretionary](https://www.dict.cc/?s=discretionary)?

Unfortunately the JSON is auto-generated from the XML, so you have to change the XML file.

ralfhandl (changes requested):

RaizoC:
changed from verb to noun Discretionary and updated description

ralfhandl (dismissed):

vocabularies/Common.xml
```diff
@@ -1118,6 +1118,11 @@ only `TargetProperties` and `TargetEntities` are relevant. They are addressed vi
           <Annotation Term="Core.Description" String="Indicates whether the side-effect has already happened" />
           <Annotation Term="Core.LongDescription" String="The value of this property typically is a Path expression pointing to a boolean property. It can be used by clients to defer expensive refresh calls until they are actually needed and instead just request the referenced indicator property. Servers can choose to return indicator properties even if not explicitly requested." />
         </Property>
+        <Property Name="Discretionary" Type="Edm.Boolean" Nullable="true">
+          <Annotation Term="Common.Experimental" />
+          <Annotation Term="Core.Description" String="Indicates whether the client can decide if a side-effect should be triggered or not" />
```
HeikoTheissen, line 1123:
```suggestion
          <Annotation Term="Core.Description" String="If `true` then the client can decide if a side-effect should be triggered or not" />
```
Clarify what happens for `null` value.

HeikoTheissen (changes requested):

vocabularies/Common.xml
ralfhandl: Or make the property `Nullable="false" DefaultValue="false"`.

RaizoC: I took over the modelling from the TriggeredIndicator which somehow is similar and optional and usually "false". So I wonder if we also should change this one. Together with a default the annotation is still optional to be added and thus compatible, right?

ralfhandl: Correct. Better also change the `TriggeredIndicator`.

HeikoTheissen: `Nullable="false"` might fail if the value is computed via a `Path`.

RaizoC: Ok, but for the new term I do not expect a path expression, but rather a static information, i.e. true / false.

HeikoTheissen: But you say otherwise:
https://github.com/SAP/odata-vocabularies/blob/5c73c570bb7283847a02009fa965c7e7181b6de0/vocabularies/Common.xml#L1124

RaizoC: Yes, copy and paste is not good, just changed it.

HeikoTheissen (approved):

ralfhandl (approved):

vocabularies/Common.xml
```diff
@@ -1118,6 +1118,11 @@ only `TargetProperties` and `TargetEntities` are relevant. They are addressed vi
           <Annotation Term="Core.Description" String="Indicates whether the side-effect has already happened" />
           <Annotation Term="Core.LongDescription" String="The value of this property typically is a Path expression pointing to a boolean property. It can be used by clients to defer expensive refresh calls until they are actually needed and instead just request the referenced indicator property. Servers can choose to return indicator properties even if not explicitly requested." />
         </Property>
+        <Property Name="Discretionary" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
```
BerSie, line 1121: why nullable: false? There already annotations out there and it seems only applicable for TriggerActions.

ralfhandl: Default value is false, which then applies to all existing annotations.

HeikoTheissen: You can still omit `Discretionary`, in this case the default value `false` is implied. I find that more elegant than nullable, because that would imply the default value `null`.

BerSie: OK - accepted :-)

BerSie (approved):

PR #118, opened 2021-06-21
Run refresh workflow also for pull requests

ralfhandl:
The workflow wasn't run for https://github.com/SAP/odata-vocabularies/pull/117

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-  pull_request:
+  pull_request_target:
 
 jobs:
   build:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,6 +1,10 @@
 name: Node CI
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:


HeikoTheissen (commented):
I don't understand why this would help. What does "on: pull_request:" mean?

ralfhandl:
This means that the workflow is triggered every time someone opens or changes a pull request, also from a forked repository, as in #117.

I guess this runs into authorization issues, need to read up more on this topic. Or hope for you picking it up 😄

HeikoTheissen:
I believe this has got nothing to do with the nodejs.yml file. The owner of a forked repository must explicitly activate actions:

> Because this repository contained workflow files when it was forked, we have disabled them from running on this fork. Make sure you understand the configured workflows and their expected usage before enabling Actions on this repository.

HeikoTheissen:
Please check first whether the actions run if @RaizoC enabled them in the forked repository.

.github/workflows/nodejs.yml
```diff
@@ -1,6 +1,10 @@
 name: Node CI
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request_target:
```
HeikoTheissen, line 7: This would then be different from what oasis-tcs/odata-vocabularies uses.

ralfhandl:
Yes, and without this change we are also different.

I added `pull_request_target` in the odata-openapi repository to get checks running on pull requests from forks.

HeikoTheissen:
I forked this repository, enabled actions and committed a change and the action was executed (https://github.com/HeikoTheissen/odata-vocabularies/actions/runs/957020264). So there is no need to change the nodejs.yml, unless you want to have the same settings as in oasis-tcs/odata-vocabularies.

PR #128, opened 2021-08-11
Composition navigation properties and composition child entity sets

ralfhandl:
Discuss:
- [x] is the annotation `CompositionChild` redundant as it can be determined from the navigation property bindings (association sets in OData V2) and annotation `Composition` on the navigation property?
- [x] is the partner navigation property required? No.

### DIFF
--- a/examples/Common.Composition-sample.xml
+++ b/examples/Common.Composition-sample.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+<edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.0">
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/Common.xml">
+    <edmx:Include Namespace="com.sap.vocabularies.Common.v1" Alias="Common" />
+  </edmx:Reference>
+
+  <edmx:DataServices>
+    <Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="example.Common.Composition" Alias="this">
+
+      <EntityContainer Name="container">
+        <EntitySet Name="SalesOrder" EntityType="this.SalesOrder_Type">
+          <NavigationPropertyBinding Path="_Item" Target="SalesOrderItem" />
+        </EntitySet>
+        <EntitySet Name="SalesOrderItem" EntityType="this.SalesOrderItem_Type" />
+      </EntityContainer>
+
+      <EntityType Name="SalesOrder_Type">
+        <Key>
+          <PropertyRef Name="SalesOrder" />
+        </Key>
+        <Property Name="SalesOrder" Type="Edm.String" Nullable="false" MaxLength="10" />
+        <!-- ... and a lot more -->
+        <NavigationProperty Name="_Item" Type="Collection(this.SalesOrderItem_Type)" Partner="_SalesOrder" />
+      </EntityType>
+
+      <EntityType Name="SalesOrderItem_Type">
+        <Key>
+          <PropertyRef Name="SalesOrder" />
+          <PropertyRef Name="SalesOrderItem" />
+        </Key>
+        <Property Name="SalesOrder" Type="Edm.String" Nullable="false" MaxLength="10" />
+        <Property Name="SalesOrderItem" Type="Edm.String" Nullable="false" MaxLength="6" />
+        <!-- ... and a lot more -->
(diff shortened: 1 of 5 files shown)


RaizoC (approved):

RaizoC (approved):

RaizoC (approved):

vocabularies/Common.xml
```diff
@@ -569,6 +569,22 @@ The minimum structure is
...
+        </Annotation>
+      </Term>
+
+      <Term Name="CompositionChild" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="EntitySet">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The entities in this set are composition children and have an existential dependency on their composition parent" />
+        <Annotation Term="Core.LongDescription">
+          <String>The navigation property representing logical composition from the "parent" entity type to the "child" entity type is annotated with [Common.Composition](#Composition)</String>
+        </Annotation>
+      </Term>
```
HeikoTheissen, line 586:
```suggestion
        <Annotation Term="Core.LongDescription">
          <String>The entities related via this navigation property have an existential dependency on their composition parent. The entity set of the composition parent MUST contain a NavigationPropertyBinding for this navigation property.</String>
        </Annotation>
      </Term>

```

HeikoTheissen (changes requested):

vocabularies/Common.xml
```diff
@@ -569,6 +569,22 @@ The minimum structure is
         </Property>
       </ComplexType>
 
+      <Term Name="Composition" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="NavigationProperty">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The annotated non-containment navigation property represents a logical composition" />
```
HeikoTheissen, line 574:
```suggestion
        <Annotation Term="Core.Description" String="The annotated navigation property represents a logical composition, even though it is non-containment" />
```

HeikoTheissen (changes requested):

HeikoTheissen (dismissed):

vocabularies/Common.xml
```diff
@@ -569,6 +569,15 @@ The minimum structure is
         </Property>
       </ComplexType>
 
+      <Term Name="Composition" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="NavigationProperty">
```
HeikoTheissen, line 572:
```suggestion
      <Term Name="Composition" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="NavigationProperty">
        <Annotation Term="Core.AppliesViaContainer"/>
```
Do we want to add this?

ralfhandl: No, I wanted to directly annotate the navigation properties, see the example.

examples/Common.Composition-sample.xml
```diff
@@ -0,0 +1,45 @@
...
+        </Key>
+        <Property Name="SalesOrder" Type="Edm.String" Nullable="false" MaxLength="10" />
+        <Property Name="SalesOrderItem" Type="Edm.String" Nullable="false" MaxLength="6" />
+        <!-- ... and a lot more -->
+        <NavigationProperty Name="_SalesOrder" Type="this.SalesOrder_Type" Nullable="false" Partner="_Item">
+          <ReferentialConstraint Property="SalesOrder" ReferencedProperty="SalesOrder" />
+        </NavigationProperty>
+      </EntityType>
+
+      <Annotations Target="this.SalesOrder/_Item">
```
HeikoTheissen, line 39: Invalid target, what do you mean?
* The entity type `this.SalesOrder_Type/_Item`
* Or the entity set `this.container.SalesOrder/_Item`

HeikoTheissen (changes requested):

examples/Common.Composition-sample.xml
ralfhandl: Type, fixed

HeikoTheissen (approved):

HeikoTheissen:
Do we require that a `Composition` navprop has a partner navprop? Must a partner navprop also be accompanied by a navprop binding?

ralfhandl:
_Ideally_ it has a partner, and ideally the partner has referential constraints to identify the parts of the key that are already determined by the composition parent.

I'll have to write a small PoC to figure out whether we can work without this additional information.

PR #134, opened 2021-10-05
Session: add AdditionalNewActions

ralfhandl:


ralfhandl:
> As the complete term is still experimental

No, the Session vocabulary and the StickySessionSupported term were both "released" on Oct 2, 2019, so we better stick with the same pattern as for "Draft"

RaizoC (approved):

BerSie (approved):

PR #139, opened 2021-10-21
Common.DraftRoot: ShareAction

ralfhandl:
- [ ] Can draft documents only be shared as a whole, or is it possible to share only a sub-tree of a draft document?


martin4all (approved):
Ok

martin4all (dismissed):
Ok

RaizoC (dismissed):

martin4all (approved):
Ok

BerSie (commented):
What is the logic of the Users list? If I send the ShareAction request twice with different users - does the second list win, or is it cumulative? And related to this question: will there be an unshare or a share with an empty list?

ralfhandl:
> What is the logic of the Users list? If I send the ShareAction request twice with different users - does the second list win, or is it cumulative? And related to this question: will there be an unshare or a share with an empty list?

@martin4all @RaizoC can you please answer this question?

RaizoC:
> > What is the logic of the Users list? If I send the ShareAction request twice with different users - does the second list win, or is it cumulative? And related to this question: will there be an unshare or a share with an empty list?
> 
> @martin4all @RaizoC can you please answer this question?

I think it was defined that the list is always complete, i.e not cumulative. Thus, calling the share action without any users will logically un-share the draft, i.e. it is not shared with anybody. Calling the action within one request multiple times lets the last one win, while within one change set it is not deterministic which one wins. The behavior is thus similar if you send a PATCH request for the same instance on the same property where only one wins.

RaizoC (approved):

BerSie (approved):

PR #148, opened 2021-11-02
Boilerplate for sap.github.io

HeikoTheissen:


ralfhandl:
Interesting idea. I had more thought of using a [Github pages theme](https://docs.github.com/en/pages/getting-started-with-github-pages/adding-a-theme-to-your-github-pages-site-with-the-theme-chooser) and thus have a consistent footer for all pages, including the README and all other manually written pages.

HeikoTheissen:
A theme would be better, of course. Are you able to define one for the GitHub Pages for this vocabulary? I am not authorized.

PR #150, opened 2021-11-08
Masking: remove MaskedAlways, add RevealOnDemand

ralfhandl:


uhlmannm (approved):

vocabularies/Common.json
```diff
@@ -303,16 +303,15 @@
...
-            "$Type": "Core.Tag",
-            "$DefaultValue": true,
+            "$Type": "Edm.Boolean",
             "$AppliesTo": [
                 "Property"
             ],
             "@Common.Experimental": true,
-            "@Core.Description": "Property contains sensitive data that is by default not transferred",
-            "@Core.LongDescription": "If the annotation evaluates to true, the unmasked property value is never transferred in responses.\n\nIf the annotation evaluates to false, the unmasked property value can be requested with the custom query option `masked-values=false`."
+            "@Common.IsInstanceAnnotation": true,
```
uhlmannm, line 313: Wenn ich das Beispiel im Wiki anschaue, scheint mir RevealOnDemand eine normale Annotation zu sein, die auf einen Pfad in den Daten zeigt. Als Instanzannotation würde ich sie doch anders verwenden?

ralfhandl: Typical use in OData V4 would be as an instance annotation, with no indication in $metadata where it may appear, because the configuration of the masking feature is rather dynamic.

The use as metadata annotation is only necessary in OData V2 services, or with frameworks that do not support instance annotations yet and have to use declared properties as a workaround.

uhlmannm: You are right. The example is for V2.
What about "with frameworks that do not support instance annotations yet"? So this means that the UI will always need to check for the metadata annotation? Would it then not make sense to just have that and let it point either to a path or have a boolean value?

PR #151, opened 2021-12-02
Add Common.WebSocketBaseURL

ralfhandl:


RaizoC (approved):

RaizoC (dismissed):

RaizoC (approved):

BerSie:
Wow that was fast ... but I would like to discuss this new term, as I have several questions:
1. Is this really something that should be annotated in a backend or would it be something for an application manifest to define datasource connections?
2. Is the URL an absolute one (hopefully not) or a relative one? In the latter case - what about the scheme (ws/wss) and the port?

RaizoC:
1. This is a special endpoint that is advertised together with the OData service, so also supports certain features offered by the service (like draft collaboration or similar). The subprotocol is also defined as "V10.PCP.SAP.COM". Port imho should be the same as the used http port and wss to be used (in case of https). Obviously there might be other specific endpoints independent of that, that then may be listed in the manifest.
2. The url definitely for this special case is a relative url on the same host as the providing OData service.

BerSie:
> 1. Port imho should be the same as the used http port and wss to be used (in case of https)

Hmm, I checked it and indeed this was changed apparently - in 7.40 you had to maintain a websocket port. Nowadays it's reusing the http port. Good to know ...thanks :-) But I'm not sure if we can assume that is always the case. So in principle it's possible that http and websocket use the same port, but that's not a law. What should a client do if the server doesn't work this way?

ralfhandl:
Is that a problem? The `WebSocketBaseURL` needs to be an absolute URL as it starts differently, so it will contain the scheme (`ws:` or `wss:`) as well as host and port.

BerSie:
> The `WebSocketBaseURL` needs to be an absolute URL



> 2\. The url definitely for this special case is a relative url

OK - can you both clarify that :-)

ralfhandl:
While we are at it: should we make this a structured term?

The current use case (different scheme, same host and port, different path as the OData service) would be
~~~json5
@Common.WebSocket: {
  scheme: "wss",
  path: "/sap/bc/apc/sap/odata",
  subprotocol: "V10.PCP.SAP.COM"
}
~~~

Other use cases can also specify `host`, `port`, or other properties that we might need in the future?

PR #155, opened 2021-12-06
Common.Locale annotation

HeikoTheissen:


HeikoTheissen:
Shouldn't every term that applies to a `Parameter` also apply to `ReturnType`?

ralfhandl:
> Shouldn't every term that applies to a `Parameter` also apply to `ReturnType`?

The term [Core.OptionalParameter](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#OptionalParameter) probably should not apply to `ReturnType` 😄

vocabularies/Common.xml
```diff
@@ -124,6 +124,16 @@
         <Annotation Term="Core.RequiresType" String="Edm.DateTimeOffset" />
       </Term>
 
+      <Term Name="Locale" Type="Edm.String" Nullable="true" AppliesTo="Property Parameter ReturnType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The value of the annotated property or parameter shall be interpreted in the given locale" />
```
ralfhandl, line 129:
```suggestion
        <Annotation Term="Core.Description" String="The value of the annotated property, parameter, or return type must be interpreted in the given locale" />
```

vocabularies/Common.xml
```diff
@@ -124,6 +124,16 @@
...
+        <Annotation Term="Core.LongDescription">
+          <String>Locales shall be specified through a [Unicode Locale Identifier](https://www.unicode.org/reports/tr35/tr35.html#Identifiers).
```
ralfhandl, line 131:
```suggestion
          <String>Locales must be specified through a [Unicode Locale Identifier](https://www.unicode.org/reports/tr35/tr35.html#Identifiers).
```

HeikoTheissen:
> The term [Core.OptionalParameter](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#OptionalParameter) probably should not apply to `ReturnType` 😄

I mean all the `IsSomething` tags, for example.

vocabularies/Common.xml
```diff
@@ -124,6 +124,16 @@
...
+          If this annotation is absent or null, the current user's or default locale is used.
+          An example for the applicability of this annotation are calendar weeks, whose meaning is locale-dependent.</String>
```
ralfhandl, line 133: Don't you want to repair the regex for "calendar week" to also allow week 54?

ralfhandl (changes requested):

ralfhandl:
> I mean all the `IsSomething` tags, for example.

Good point, let's check those and extend their applicability where necessary.

vocabularies/Common.xml
HeikoTheissen: I cannot judge the precise difference between "shall" and "must", but I feel uncomfortable saying "You must ..." in connection with an annotation, because clients are always allowed to not understand a given annotation.

vocabularies/Common.xml
HeikoTheissen: Yes, but for the fiscal week I would not want to specify any pattern, because I don't know how high that can go.

vocabularies/Common.xml
ralfhandl: "shall" is an outdated synonym for "must". Use "shall" if you want to sound old-fashioned, or like a biblical prophet 😄

vocabularies/Common.xml
ralfhandl: Then let's remove the pattern altogether if we can't find out what the globally correct pattern is.

vocabularies/Common.xml
```diff
@@ -114,16 +114,26 @@
         </Member>
       </EnumType>
 
-      <Term Name="Timezone" Type="Edm.String" Nullable="true" AppliesTo="Property Parameter">
+      <Term Name="Timezone" Type="Edm.String" Nullable="true" AppliesTo="Property Parameter ReturnType">
         <Annotation Term="Common.Experimental" />
-        <Annotation Term="Core.Description" String="The point in time represented by the annotated property or parameter shall be interpreted in the given time zone" />
+        <Annotation Term="Core.Description" String="The point in time represented by the annotated property, parameter or return type shall be interpreted in the given time zone" />
```
ralfhandl, line 119:
```suggestion
        <Annotation Term="Core.Description" String="The point in time represented by the annotated property, parameter or return type must be interpreted in the given time zone" />
```
See https://datatracker.ietf.org/doc/html/rfc2119#section-1 and https://www.plainlanguage.gov/guidelines/conversational/shall-and-must/:
- "must" and "shall" mean the same
- "must" is better

vocabularies/Common.xml
```diff
@@ -114,16 +114,26 @@
...
         <Annotation Term="Core.LongDescription">
           <String>Time zones shall be specified according to the [IANA](https://www.iana.org/time-zones) standard.
```
ralfhandl, line 121:
```suggestion
          <String>Time zones must be specified according to the [IANA](https://www.iana.org/time-zones) standard.
```

vocabularies/Common.xml
```diff
@@ -114,16 +114,26 @@
...
             If this annotation is absent or null, points in time are typically interpreted in the current user's or default time zone.</String>
         </Annotation>
         <Annotation Term="Core.RequiresType" String="Edm.DateTimeOffset" />
       </Term>
 
+      <Term Name="Locale" Type="Edm.String" Nullable="true" AppliesTo="Property Parameter ReturnType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The value of the annotated property, parameter or return type shall be interpreted in the given locale" />
```
ralfhandl, line 129:
```suggestion
        <Annotation Term="Core.Description" String="The value of the annotated property, parameter or return type must be interpreted in the given locale" />
```

ralfhandl (changes requested):

vocabularies/Common.xml
HeikoTheissen: Remove it also for calendar weeks, or only for fiscal weeks?

ralfhandl: For calendar weeks we would have to extend it to cover week 54 that can occur in the US, correct?

ralfhandl (approved):

ralfhandl:
The "fiscal" terms were introduced by @GeraldKrause, he should also review this PR.

GeraldKrause (changes requested):
Why remove the regexp's from the fiscal terms? They are also given for calendar terms. In the end, they formally define the range of valid values and can be applied by an implementation for validation.

HeikoTheissen:
> Why remove the regexp's from the fiscal terms? They are also given for calendar terms. In the end, they formally define the range of valid values and can be applied by an implementation for validation.

Are you sure that their value range is identical to that of a calendar week? (After learning that there is a calendar week 54, I am not so sure about my knowledge in this area any more.)

ralfhandl (approved):

GeraldKrause:
> > Why remove the regexp's from the fiscal terms? They are also given for calendar terms. In the end, they formally define the range of valid values and can be applied by an implementation for validation.
> 
> Are you sure that their value range is identical to that of a calendar week? (After learning that there is a calendar week 54, I am not so sure about my knowledge in this area any more.)

You are right. I checked my mail conversation with Thomas Schneider, subject matter expert in S/4 and found a mail from 2019 where the CDS ABAP Core Annotation spec for fiscalweek and fiscalyearweek also mention the WW pattern and the states for the latter: "The last two digits represent the fiscal week in the year." So, the previous description for this term was more restrictive than needed.

GeraldKrause (dismissed):

PR #157, opened 2021-12-15
SemanticObject applies to NavigationProperties

BerSie:
Common.Semantic annotation could also be applied to NavigationProperties (as the foreign keys might not be provided, e.g. by using cds.odata.refs flavor)

### DIFF
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -319,7 +319,8 @@
             "$AppliesTo": [
                 "EntitySet",
                 "EntityType",
-                "Property"
+                "Property",
+                "NavigationProperty"
             ],
             "@Core.Description": "Name of the Semantic Object represented as this entity type or identified by this property"
         },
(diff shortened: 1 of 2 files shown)


RaizoC (approved):

HeikoTheissen:
Can you give an example annotation and how it would affect the UI?

ralfhandl (approved):

BerSie:
> Can you give an example annotation and how it would affect the UI?

```
<Annotations Target="SAP__self.RootType/Id">
  <Annotation Term="SAP__common.SemanticObject" String="MyRootEntity"/>
</Annotations>
<Annotations Target="SAP__self.RootType/_Node">
  <Annotation Term="SAP__common.SemanticObject" String="MyNodeEntity"/>
</Annotations>
```
With that we are able to provide a navigation (and showing a QuickView) without having foreign keys defined in the RootEntity.

HeikoTheissen (approved):

PR #164, opened 2022-02-04
Describe annotations, delete URIType, add FontName, Border, and Margin

ralfhandl:
Reverts SAP/odata-vocabularies#163

### DIFF
--- a/vocabularies/PDF.json
+++ b/vocabularies/PDF.json
@@ -1,0 +1,118 @@
+{
+    "$Version": "4.0",
+    "$Reference": {
+        "https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.json": {
+            "$Include": [
+                {
+                    "$Namespace": "Org.OData.Core.V1",
+                    "$Alias": "Core"
+                }
+            ]
+        },
+        "https://sap.github.io/odata-vocabularies/vocabularies/Common.json": {
+            "$Include": [
+                {
+                    "$Namespace": "com.sap.vocabularies.Common.v1",
+                    "$Alias": "Common"
+                }
+            ]
+        }
+    },
+    "com.sap.vocabularies.PDF.v1": {
+        "$Alias": "PDF",
+        "@Common.Experimental": true,
+        "@Core.Description": "Terms for PDF response format",
+        "@Core.LongDescription": "The PDF vocabulary provides information about the PDF format of a response",
+        "@Core.Description#Published": "2021-10-27 © Copyright 2021 SAP SE. All rights reserved",
+        "@Core.Links": [
+            {
+                "rel": "alternate",
+                "href": "https://sap.github.io/odata-vocabularies/vocabularies/PDF.xml"
+            },
+            {
+                "rel": "latest-version",
+                "href": "https://sap.github.io/odata-vocabularies/vocabularies/PDF.json"
+            },
+            {
+                "rel": "describedby",
+                "href": "https://github.com/sap/odata-vocabularies/blob/main/vocabularies/PDF.md"
+            }
+        ],
+        "Features": {
+            "$Kind": "Term",
+            "$Type": "PDF.FeaturesType",
+            "$AppliesTo": [
+                "EntityContainer"
+            ],
+            "@Common.Experimental": true,
+            "@Core.Description": "Features for the PDF"
+        },
+        "FeaturesType": {
+            "$Kind": "ComplexType",
+            "@Common.Experimental": true,
+            "DocumentDescriptionReference": {
+                "@Core.IsURL": true,
+                "@Core.Description": "Reference of the Service for the DocumentDescription"
+            },
+            "DocumentDescriptionCollection": {
+                "@Core.Description": "Name of entity set containing the DocumentDescription"
+            },
+            "ArchiveFormat": {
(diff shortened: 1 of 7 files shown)


vocabularies/PDF.json
```diff
@@ -0,0 +1,118 @@
...
+            "Margin": {
+                "$Type": "Edm.Int16",
+                "$Nullable": true,
+                "@Core.Description": "Margin size"
+            },
+            "Border": {
+                "$Type": "Edm.Int16",
+                "$Nullable": true,
+                "@Core.Description": "Border size of the table"
+            },
```
HerzogIgzorn, line 96: The properties _FontName, FontSize, Margin, Border_ seem to be default values in contrast to other properties like _Signature_ or _CoverPage_, which define whether a specific feature is available or not. Shouldn't this be reflected in the name of the property?

vocabularies/PDF.md
```diff
@@ -0,0 +1,31 @@
...
+[Signature](./PDF.xml#L56:~:text=<ComplexType%20Name="-,FeaturesType,-")|Boolean?|Signing the document
+[CoverPage](./PDF.xml#L59:~:text=<ComplexType%20Name="-,FeaturesType,-")|Boolean?|Cover Page
+[FontName](./PDF.xml#L62:~:text=<ComplexType%20Name="-,FeaturesType,-")|String?|Font name
+[FontSize](./PDF.xml#L65:~:text=<ComplexType%20Name="-,FeaturesType,-")|Int16?|Font size
+[Margin](./PDF.xml#L68:~:text=<ComplexType%20Name="-,FeaturesType,-")|Int16?|Margin size
+[Border](./PDF.xml#L71:~:text=<ComplexType%20Name="-,FeaturesType,-")|Int16?|Border size of the table
+[FitToPage](./PDF.xml#L74:~:text=<ComplexType%20Name="-,FeaturesType,-")|Boolean?|Fit to page<br>If this option is selected, the font size is automatically selected in such a way that all columns of a table fit on one page. Other layout options like margin, border and composite cell spacing are adapted accordingly, with respect to the chose scaling factor.
+[ResultSizeDefault](./PDF.xml#L80:~:text=<ComplexType%20Name="-,FeaturesType,-")|Int32?|Default result size<br>Default result size for PDF documents. Used if a the $top has not been provided.
```
HerzogIgzorn, line 30: Int16 should be good enough for this property.

vocabularies/PDF.json
```diff
@@ -0,0 +1,118 @@
...
+            "FitToPage": {
+                "$Type": "Edm.Boolean",
+                "$Nullable": true,
+                "$DefaultValue": false,
+                "@Core.Description": "Fit to page",
+                "@Core.LongDescription": "If this option is selected, the font size is automatically selected in such a way that all columns of a table fit on one page. Other layout options like margin, border and composite cell spacing are adapted accordingly, with respect to the chose scaling factor."
+            },
+            "ResultSizeDefault": {
+                "$Type": "Edm.Int32",
```
HerzogIgzorn, line 105: Int16 should be good enough for this property

vocabularies/PDF.json
```diff
@@ -0,0 +1,118 @@
...
+            "$Kind": "ComplexType",
+            "@Common.Experimental": true,
+            "DocumentDescriptionReference": {
+                "@Core.IsURL": true,
+                "@Core.Description": "Reference of the Service for the DocumentDescription"
+            },
+            "DocumentDescriptionCollection": {
+                "@Core.Description": "Name of entity set containing the DocumentDescription"
+            },
+            "ArchiveFormat": {
```
HerzogIgzorn, line 60: Does this property has any effect on how we process the PDF export? Is this necessary to identify whether we can set the doEnableAccessibility property on the DocumentDescription?

simlin (commented):
Looks fine for me

uhlmannm (approved):

vocabularies/PDF.json
bernhardgr42: Our API has the type Int32. So this value can be higher

bernhardgr42:
@HerzogIgzorn : please read my comments.

vocabularies/PDF.json
HerzogIgzorn: Ok, if there is a technical reason for Int32, it's fine.

ralfhandl:
@HerzogIgzorn are you satisfied with the resolution of your remarks, and do you approve this PR now?

Thanks in advance!

vocabularies/PDF.json
HerzogIgzorn: Although you might not be able to ask the PDF Export Service about the actual defaults, it makes more sense that the GW provides defaults for those properties. In this case you should also name them like you did with _ResultSizeDefault_. Otherwise we would have to hardcode some defaults and then this properties are obsolete.

vocabularies/PDF.json
HerzogIgzorn: A quick call with Bernhard revealed that there is currently no possibility to configure the archive format on the DocumentDescription. The ArchiveFormat property on the PDF.Features makes only sense if we can utilize this in any way. Please inform the V2 GW colleagues in case you change anything on the DocumentDescription.

ralfhandl:
Closed in favor of #170, please continue the review there.

PR #167, opened 2022-02-16
Changing type of Margin, Border, FitToPage, FontSize, & FontName

bernhardgr42:
Margin, Border, FitToPage, FontSize, & FontName are new features of the PDF_API. The annotation shows you if this option is available from the server.
=> Changing the types to boolean.

We have no default FontName, FonstSize and so on. Only the pdf API knows this. If the pdf API add this default knowledge, it could be (for an older FES), that this information is not available. Then we would have to set NULL for these default FontName, FontSize, etc. setting. Therefore, I would not add this here.


ralfhandl:
Seems to be a duplicate of #170

PR #175, opened 2022-04-06
Multilineparameter

BerSie:
Adding _Parameter_ to the AppliesTo of UI.MultiLineText allows the usage of this annotation on operation parameters (e.g. a note on creating a vacation request :-)


vocabularies/UI.xml
```diff
@@ -1349,7 +1349,7 @@ The trend is
         </Annotation>
       </Term>
 
-      <Term Name="MultiLineText" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property PropertyValue">
+      <Term Name="MultiLineText" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property PropertyValue Parameter">
```
HeikoTheissen, line 1352:
```suggestion
      <Term Name="MultiLineText" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property PropertyValue String Parameter">
```
`AppliesTo="Parameter"` is OK, but if it shall apply to a `PropertyValue` (that is, in a complex-typed annotation), then we should also allow `AppliesTo="String"` (for primitive annotations).

vocabularies/UI.xml
```diff
@@ -1349,7 +1349,7 @@ The trend is
...
         <Annotation Term="Core.Description" String="Properties annotated with this annotation should be rendered as multi-line text (e.g. text area)" />
```
ralfhandl, line 1353:
```suggestion
        <Annotation Term="Core.Description" String="Properties and parameters annotated with this annotation should be rendered as multi-line text (e.g. text area)" />
```

HeikoTheissen (changes requested):

ralfhandl (commented):
I would adjust the description to match the new applicability

vocabularies/UI.xml
ralfhandl: I don't find `String` in the list of [allowed values for `AppliesTo`](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Applicability)

ralfhandl: If you want to annotate a _term_ of type "String" with `MultiLineText`, then we can add `Term` to the `AppliesTo` list.

HeikoTheissen: > I don't find `String` in the list of [allowed values for `AppliesTo`](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Applicability)
> If you want to annotate a _term_ of type "String" with `MultiLineText`, then we can add `Term` to the `AppliesTo` list.

But I (as vocabulary author) don't want to annotate the term, the service developer wants to annotate their _usage_ of the term. Means `AppliesTo="Property Parameter PropertyValue Annotation"`?

ralfhandl: Yes, if you only want to annotate the usage of a term, then `Annotation` is the correct value.

Do we have use cases for annotating annotations with this term, or can we wait until such use cases pop up?

HeikoTheissen:
```
<Annotation Term="Core.Description" String="A very long description">
  <Annotation Term="UI.MultiLineText"/>
</Annotation>
```
seems very natural to me.

ralfhandl (approved):

PR #176, opened 2022-04-11
Generalize UI.DataField to collections of primitive values in addition to primitive values

ralfhandl:
- Allow a collection of primitive values in addition to a single primitive value for `Value` property of `DataField`
- Restrict the `DataFieldWithX` sub-types of `DataField` to a single primitive value (same as now)

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1569,6 +1569,7 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
       <ComplexType Name="DataFieldWithAction" BaseType="UI.DataField">
         <Annotation Term="Core.Description" String="A piece of data that allows triggering an OData action" />
         <Annotation Term="Core.LongDescription" String="The action is tied to a data value which should be rendered as a hyperlink. This is in contrast to [DataFieldForAction](#DataFieldForAction)) which is not tied to a specific data value." />
+        <Property Name="Value" Type="Edm.PrimitiveType" Nullable="false" />
         <Property Name="Action" Type="Common.QualifiedName" Nullable="false">
           <Annotation Term="Core.Description" String="Qualified name of an Action, Function, ActionImport or FunctionImport in scope" />
         </Property>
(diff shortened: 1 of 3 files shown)


vocabularies/UI.xml
```diff
@@ -1554,9 +1554,15 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
...
-        <Property Name="Value" Type="Edm.PrimitiveType" Nullable="false">
+        <Property Name="Value" Type="Edm.Untyped" Nullable="false">
           <Annotation Term="Core.Description" String="The data field's value" />
           <Annotation Term="Core.IsLanguageDependent" />
+          <Annotation Term="Validation.DerivedTypeConstraint">
+            <Collection>
+              <String>Edm.PrimitiveType</String>
+              <String>Collection(Edm.PrimitiveType)</String>
+            </Collection>
+          </Annotation>
```
HeikoTheissen, line 1565: Is `Collection(Edm.PrimitiveType)` a qualified type name?
https://github.com/oasis-tcs/odata-vocabularies/blob/32ec59be5f831018d5fc40ac46288b017a889faf/vocabularies/Org.OData.Core.V1.xml#L452-L454

Or should the type definition say "... or a collection thereof"?

vocabularies/UI.xml
```diff
@@ -1554,15 +1554,22 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
...
           <Annotation Term="Core.IsLanguageDependent" />
         </Property>
       </ComplexType>
 
       <ComplexType Name="DataFieldWithAction" BaseType="UI.DataField">
         <Annotation Term="Core.Description" String="A piece of data that allows triggering an OData action" />
         <Annotation Term="Core.LongDescription" String="The action is tied to a data value which should be rendered as a hyperlink. This is in contrast to [DataFieldForAction](#DataFieldForAction)) which is not tied to a specific data value." />
+        <Property Name="Value" Type="Edm.PrimitiveType" Nullable="false" />
```
HeikoTheissen, line 1572: The specification says that the derived type "inherits" the properties. Does that imply the possiblity that their type is restricted?

ralfhandl: Correct, see [CSDL, 7 Structural Property](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_StructuralProperty)

> If a structural property with the same name is defined in any of this type’s base types, then the property’s type MUST be a type derived from the type specified for the property of the base type and constrains this property to be of the specified subtype for instances of this structured type.

`Edm.PrimitiveType` is a subtype of `Edm.Untyped`, so this is allowed.

BerSie (approved):
I like this approach to enhance the UI.Datafield with multi value functionality in a compatible way

HeikoTheissen (approved):

ThomasChadzelek (approved):

uhlmannm (approved):

PR #178, opened 2022-04-14
UI.DataFieldMulti for collections of primitive values

ralfhandl:
This is an alternative to #176


HeikoTheissen (changes requested):
I prefer changing the type of `UI.DataField/Value` to `Edm.Untyped`.

PR #180, opened 2022-04-25
Dynamic Properties example

HeikoTheissen:
> As a general note, the example lacks an Aggregation.ApplySupported annotation, right?

The example is not meant to illustrate that.

> Either remove NumberOfItems from the Chart annotation or adjust the example and define it as CustomAggregate or as another AggregatedProperty.

Removed it.

> Finally, changing #sum to #totalAmount and #max to #maxAmount would improve readability of the example

Done.

### DIFF
--- a/examples/DynamicProperties-sample.json
+++ b/examples/DynamicProperties-sample.json
@@ -83,7 +83,7 @@
                             "Property": "Country"
                         },
                         {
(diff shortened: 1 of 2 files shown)


GeraldKrause (approved):
LGTM

PR #182, opened 2022-05-16
Effect of $search on RecursiveHierarchy

HeikoTheissen:
This PR provides SAP-specific information about recursive hierarchies in additional entity type properties. This only works under the assumption that the recursive hierarchy is defined on the requested resource itself.

For the more general case, where the hierarchy-carrying entity type can differ from the requested entity type, instance annotations must be used instead of properties, see #184.


ralfhandl (approved):

GeraldKrause (approved):

vocabularies/Hierarchy.md
```diff
@@ -13,10 +13,18 @@ Term|Type|Description
...
-[SiblingRankProperty](./Hierarchy.xml#L59:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the sibling rank of a node<br>The sibling rank of a node is the index of the node in the sequence of all nodes with the same parent created by preorder traversal of the hierarchy structure after evaluating the $filter expression in the request excluding any conditions on key properties. The first sibling is at position 0.
-[PreorderRankProperty](./Hierarchy.xml#L63:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the preorder rank of a node<br>The preorder rank of a node expresses its position in the sequence of nodes created from preorder traversal of the hierarchy structure after evaluating the $filter expression in the request excluding any conditions on key properties. The first node in preorder traversal has rank 0.
+[ExternalKeyProperty](./Hierarchy.xml#L55:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath|Property holding the external key value for a node
```
ThomasChadzelek, line 24: Not related to this change, actually. This is the single mandatory property here and I wonder what it is really needed for? What is its role and where can I see some example or so?

vocabularies/Hierarchy.xml
```diff
@@ -45,24 +45,36 @@
...
+          the values of these properties are derived from the output set of the last hierarchy-related transformation
+          (`ancestors`, `descendants`, `traverse`) during the evaluation of an `$apply` query option,
+          they are not affected by subsequent transformations or query options (like `$filter` or `$top`).</String>
+        </Annotation>
         <Property Name="ExternalKeyProperty" Type="Edm.PropertyPath" Nullable="false">
           <Annotation Term="Core.Description" String="Property holding the external key value for a node" />
         </Property>
         <Property Name="DescendantCountProperty" Type="Edm.PropertyPath" Nullable="true">
-          <Annotation Term="Core.Description" String="Property holding the number of descendants of a node" />
-          <Annotation Term="Core.LongDescription" String="The descendant count of a node is the number of its descendants in the hierarchy structure of the result considering only those nodes matching any specified $filter and $search. A property holding descendant counts has an integer data type." />
```
ThomasChadzelek, line 53: Core.LongDescription lost here?

vocabularies/Hierarchy.md
```diff
@@ -13,10 +13,18 @@ Term|Type|Description
...
+[DescendantCountProperty](./Hierarchy.xml#L58:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property holding the number of descendants of a node
+[DrillStateProperty](./Hierarchy.xml#L61:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the drill state of a node<br>The drill state is indicated by one of the following string values: `collapsed`, `expanded`, `leaf`. For an expanded node, its children are included in the result collection. For a collapsed node, the children are included in the entity set, but they are not part of the result collection. Retrieving them requires a relaxed filter expression or a separate request filtering on the parent node ID with the ID of the collapsed node. A leaf does not have any child in the entity set.
```
ThomasChadzelek, line 26: I wonder a bit about this enum-like string. Looks strange to me. Why send such long strings on the wire? Human readability in a response targeted at a machine?

"Retrieving them requires a relaxed filter expression or a separate request filtering on the parent node ID with the ID of the collapsed node." Is very technical. Not sure if the advice is really spot on. Maybe less is more here?

ThomasChadzelek: "result collection" (here) vs. "output set" (line 16ff.)?

"entity set" vs. "input collection" or so - I guess this wording is also not aligned with $apply

vocabularies/Hierarchy.md
```diff
@@ -13,10 +13,18 @@ Term|Type|Description
...
+[SiblingRankProperty](./Hierarchy.xml#L65:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property holding the sibling rank of a node<br>The sibling rank of a node is the index of the node in the sequence of all nodes with the same parent created by preorder traversal. The first sibling has rank 0.
```
ThomasChadzelek, line 27: Does "preorder" matter here?

vocabularies/Hierarchy.md
```diff
@@ -13,10 +13,18 @@ Term|Type|Description
...
+[PreorderRankProperty](./Hierarchy.xml#L69:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property holding the preorder rank of a node<br>The preorder rank of a node expresses its position in the sequence of nodes created from preorder traversal. The first node in preorder traversal has rank 0.
+[IsLeafProperty](./Hierarchy.xml#L73:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Boolean property indicating whether a node is a leaf
+[LevelProperty](./Hierarchy.xml#L76:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property holding a node's level, which increases when descending the hierarchy
```
ThomasChadzelek, line 30: "descending the hierarchy" is kind of an ad-hoc term. Maybe "a child's level is strictly larger than the level of its parent"? And maybe even "it is possible that root nodes are not on level 1 and direct children may be more than one level below their parents" where the 2nd part is a bit redundant, but emphasizes what might be astonishing (copied from https://help.sap.com/doc/e4fc5d26a38c4dc28802a3b87cd5eb10/2.0.05/en-US/SAP_HANA_Hierarchy_Developer_Guide_en.pdf).

vocabularies/Hierarchy.xml
```diff
@@ -45,24 +45,36 @@
...
+          <Annotation Term="Core.Description" String="Integer property holding the number of descendants of a node" />
```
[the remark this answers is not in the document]
HeikoTheissen: There is no annotation `Validation.AllowedTypes` that would make this machine-readable.

HeikoTheissen:
> Sorry, but what is the enum "expanded, collapsed, leaf" good for if I cannot distinguish collapsed nodes from leaves? Does that really lead to a good UX?

If the client requests descendants up to level 3, the server may not want to check whether the nodes on level 3 are actually leaves, because that's too expensive. So it would return level 3 nodes as `collapsed`, and higher level ($<3$) nodes as `expanded` or `leaf`. Only when the user clicks on a level 3 node do they learn whether it is a leaf.

I consider that a meaningful use case and do not want to exclude it in this term definition.

ThomasChadzelek:
OK, only time will tell. Or Markus Königstein, paying attention to "V2 compatibility" - do you have an eye on that?

HeikoTheissen:
Of course our servers can promise never to return `collapsed` for leaves, if they feel they can ensure that.

ThomasChadzelek:
How does this relate to IsLeafProperty? If I $select that, I would get the exact information about the requested resource, independent of the output set? That would be nice, the client can then determine what it needs and how much it is willing to pay.

HeikoTheissen:
> How does this relate to IsLeafProperty? If I $select that, I would get the exact information about the requested resource, independent of the output set? That would be nice, the client can then determine what it needs and how much it is willing to pay.

Yes, and the server can determine which of these properties to offer.

ThomasChadzelek (dismissed):

ralfhandl (dismissed):

vocabularies/Hierarchy.md
```diff
@@ -13,10 +13,20 @@ Term|Type|Description
 ## <a name="RecursiveHierarchyType"></a>[RecursiveHierarchyType](./Hierarchy.xml#L46:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-") *([Experimental](Common.md#Experimental))*
 
 
+The nullable property paths in this complex type refer to properties of an entity type that contain
+          information about a node in a recursive hierarchy. When a resource of the annotated entity type is requested
+          and hierarchy-related transformations (`ancestors`, `descendants`, `traverse`) applied using the hierarchy qualifier of the annotation,
```
GeraldKrause, line 18:
```suggestion
          and hierarchy-related transformations (`ancestors`, `descendants`, `traverse`) applied to it using the resource as first parameter and the hierarchy qualifier of the annotation as second parameter,
```

I read your description as if you apply the simplification/restriction we discussed in our call: The first parameter of ancestors/descendants is identical the input set. If that's what you intended I would state it clearly as suggested

ThomasChadzelek (approved):
Who am I to judge?

ralfhandl (approved):

HeikoTheissen:
> Einspruch, Euer Ehren! Das war bisher für mich die zentrale technische Property, die ich zum Blättern über die "_limited_ hierarchy" brauche! Welcher Ersatz wäre denn dafür vorgesehen?

[`ExpandedDescendantCountProperty`](https://github.com/SAP/odata-vocabularies/pull/182/commits/9e8246ea4e89ad1cb9e6ab2ea5d4272a36a00f72)

vocabularies/Hierarchy.xml
```diff
@@ -45,27 +45,133 @@
...
+the following collections are distinguished:
+- sub-hierarchy: the output set of a `descendants` transformation filtering a single node value:
+  US sales organizations (first two rows of request)
+- [matching](#MatchCount) nodes: US sales organizations with "New York" in their name
+  (output set of `filter` transformation in third row of request)
+- unlimited hierarchy: the output set of the last `ancestors`, `descendants` or `traverse` transformation
+  disregarding numeric fifth parameters: (first three rows of request)
+  `descendants(..., filter(ID eq 'US'), keep start)/ancestors(..., filter(contains(Name, 'New York')))`
+- limited hierarchy: the output set of the last `ancestors`, `descendants`, `traverse` or
+  [`Hierarchy.TopLevels`](#TopLevels) transformation: (first four rows of request)</String>
```
HeikoTheissen, line 72: These definitions, on which the properties of `RecursiveHierarchy` are based, are somewhat fluffy and reflect the thinking of OData V2 hierarchies, where the following kinds of `$filter` were supported:
- filter on node ID: determines a sub-hierarchy
- filter on parent ID: determines children of a node
- filter on `hierarchy_level`: limits the hierarchy
- filter on other property: determines matching nodes.

These are also the kinds of filter that the underlying database supports. OData V4 goes conceptually beyond that, but then the properties of `RecursiveHierarchy` no longer work.

Is that restriction OK?

vocabularies/Hierarchy.md
```diff
@@ -9,14 +9,66 @@ Terms for Hierarchies
...
-[PreorderRankProperty](./Hierarchy.xml#L63:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the preorder rank of a node<br>The preorder rank of a node expresses its position in the sequence of nodes created from preorder traversal of the hierarchy structure after evaluating the $filter expression in the request excluding any conditions on key properties. The first node in preorder traversal has rank 0.
+[ExternalKeyProperty](./Hierarchy.xml#L72:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath|Property holding the external key value for a node
+[NodeTypeProperty](./Hierarchy.xml#L75:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the type of a node<br>In a recursive hierarchy with mixed types, nodes of different type can have different properties and be represented by entities of different subtypes of a commmon entity type. The common entity type is the target of the `RecursiveHierarchy` annotation.
```
ThomasChadzelek, line 66: There is no restriction w.r.t. a type here (e.g. Edm.String), right? Would that make sense?

vocabularies/Hierarchy.md
```diff
@@ -9,14 +9,66 @@ Terms for Hierarchies
...
-[SiblingRankProperty](./Hierarchy.xml#L59:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the sibling rank of a node<br>The sibling rank of a node is the index of the node in the sequence of all nodes with the same parent created by preorder traversal of the hierarchy structure after evaluating the $filter expression in the request excluding any conditions on key properties. The first sibling is at position 0.
-[PreorderRankProperty](./Hierarchy.xml#L63:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the preorder rank of a node<br>The preorder rank of a node expresses its position in the sequence of nodes created from preorder traversal of the hierarchy structure after evaluating the $filter expression in the request excluding any conditions on key properties. The first node in preorder traversal has rank 0.
+[ExternalKeyProperty](./Hierarchy.xml#L72:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath|Property holding the external key value for a node
```
ThomasChadzelek, line 65: There is no restriction w.r.t. a type here (e.g. Edm.String), right? Would that make sense?

vocabularies/Hierarchy.md
```diff
@@ -9,14 +9,66 @@ Terms for Hierarchies
...
+[ChildCountProperty](./Hierarchy.xml#L83:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property of type `Edm.Int64` holding the number of children a node has in the unlimited hierarchy
+[DescendantCountProperty](./Hierarchy.xml#L86:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property of type `Edm.Int64` holding the number of descendants a node has in the unlimited hierarchy
+[LimitedDescendantCountProperty](./Hierarchy.xml#L89:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property of type `Edm.Int64` holding the number of descendants a node has in the limited hierarchy
+[DrillStateProperty](./Hierarchy.xml#L92:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|String property holding the drill state of a node<br>Possible drill states are: <br>- `expanded` if a node has children in the limited hierarchy <br>- `collapsed` if a node has children in the unlimited hierarchy but not in the limited hierarchy <br>- `leaf` if a node has no children in the unlimited hierarchy
```
ThomasChadzelek, line 70: Maybe we could be less strict here and only request strings that start with "C", "E", or "L" (disregarding case)? Or even use a number, e.g.
- 0 for a leaf
- +1 for a collapsed node ("+" looks like the icon where you could expand)
- -1 for an expanded node (the "opposite" of collapsed)
But maybe that is a bit far fetched?

vocabularies/Hierarchy.md
```diff
@@ -9,14 +9,66 @@ Terms for Hierarchies
...
+[DistanceFromRootProperty](./Hierarchy.xml#L101:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property of type `Edm.Int64` holding the number of ancestors a node has in the (sub-)hierarchy
+[SiblingRankProperty](./Hierarchy.xml#L104:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Integer property of type `Edm.Int64` holding the sibling rank of a node<br>The sibling rank of a node is the index of the node in the sequence of all nodes in the unlimited hierarchy with the same parent. The first sibling has rank 0. Updating this property means changing the relative sibling order.
```
ThomasChadzelek, line 72: Should we add a hint that updating does not make sense when an orderby is present which (of course) influences sibling order? That is, only w/o orderby you get the "native" order which you may be able to PATCH.

By the way: if "unlimited hierarchy" is relevant (as opposed to "(sub-)hierarchy"), then how would that be identified in a PATCH request which targets a single entity directly, w/o $apply?

vocabularies/Hierarchy.md
HeikoTheissen: To be discussed in V4 tree meeting.

vocabularies/Hierarchy.md
ThomasChadzelek: For the node type, I would expect the qualified name of the entity type. Or is that up to the service to even use some number or enum?

For the external key value, I guess "external" means that you can display it, and then a string makes most sense to me. Or could that be even a timestamp and then you need to format it incl. all dicussions about timezone?

HeikoTheissen: The external key may have a complex type made up from the node type and a node-specific id.

vocabularies/Hierarchy.xml
```diff
@@ -45,27 +45,136 @@
...
 
+      <Term Name="MatchCount" Type="Edm.Int64" AppliesTo="Collection">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Common.IsInstanceAnnotation" />
+        <Annotation Term="Core.Description" String="Instance annotation on the result of an `$apply` query option containing the number of matching nodes after hierarchical transformations" />
+        <Annotation Term="Core.LongDescription">
+          <String>The service designates a subset of the `$apply` result as "matching nodes" derived from the start nodes
+          of `ancestors` and `descendants` as determined by the transformations in their fourth parameters.
+          This subset is typically the output set of a `filter` or `search` transformation that evaluates node
+          attributes other than the node value.&lt;br&gt;This instance annotation is available if [`RecursiveHierarchy/MatchedProperty`](#RecursiveHierarchyType)
```
HeikoTheissen, line 138:
```suggestion
          This subset is the output set of the `filter` or `search` transformation in the fourth parameter of the last `ancestors` transformation.
          &lt;br&gt;This instance annotation is available if [`RecursiveHierarchy/MatchedProperty`](#RecursiveHierarchyType)
```
Would this be equivalent for our purposes? It is certainly easier to implement.

ThomasChadzelek: Looks good to me. I agree that "ancestors" is the relevant transformation for this kind of search/filter: you need to locate matches and the complete the tree "upwards". A "descendants" transformation has a different semantics, it restricts to a sub-tree, but that is not what a "match" needs to indicate. You will notice that effect by looking at the resulting root node(s).

ThomasChadzelek (approved):

HeikoTheissen:
We plan to merge this PR at 2023-02-24T15:00:00+01:00 (after the V4 tree meeting). Until then
- we need at least one approving review by someone with write access to this repository
- please raise any objections you might have with the current proposal especially of the `Hierarchy.RecursiveHierarchy` annotation type.

ThomasChadzelek (dismissed):

vocabularies/Hierarchy.xml
```diff
@@ -45,27 +45,188 @@
...
+            <String>If a `NodeTypeProperty` exists, the external key is unique only in combination with it.
+            Or the external key can coincide with the [`Aggregation.RecursiveHierarchy/NodeProperty`](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.html#RecursiveHierarchyType).</String>
+          </Annotation>
+        </Property>
+        <Property Name="NodeTypeProperty" Type="Edm.PropertyPath" Nullable="true">
+          <Annotation Term="Core.Description" String="String property holding the type of a node" />
+          <Annotation Term="Core.LongDescription">
+            <String>In a recursive hierarchy with mixed types, nodes can
+            &lt;br&gt;- have a type-specific (navigation) property whose name is the node type or
+            &lt;br&gt;- be represented by entities of different subtypes of a commmon entity type that is
```
ralfhandl, line 94:
```suggestion
            &lt;br&gt;- be represented by entities of different subtypes of a common entity type that is
```

ralfhandl (changes requested):
It's a bit surprising that all properties of `RecursiveHierarchyType` are nullable.

HeikoTheissen:
> It's a bit surprising that all properties of `RecursiveHierarchyType` are nullable.

None of them is so important that it must always be present.

ralfhandl:
> None of them is so important that it must always be present.

Is there a use case where none of the properties is present and the annotation only serves as a marker?

ralfhandl (dismissed):

ThomasChadzelek:
> > Is there a use case where none of the properties is present and the annotation only serves as a marker?
> 

From the perspective of UI5's v4.ODataModel, DrillStateProperty is an absolute must. If you don't know whether a node is a leaf or not, you are lost.

LimitedDescendantCountProperty and DistanceFromRootProperty are mandatory as soon as you want to start with n > 1 initially expanded levels, which is the reason we invented TopLevels and the main benefit of V4 over V2, right?

The rest is optional ("graceful degradation")

HeikoTheissen:
> From the perspective of UI5's v4.ODataModel, DrillStateProperty is an absolute must. If you don't know whether a node is a leaf or not, you are lost.
> LimitedDescendantCountProperty and DistanceFromRootProperty are mandatory as soon as you want to start with n > 1 initially expanded levels, which is the reason we invented TopLevels and the main benefit of V4 over V2, right?

But this annotation is meant for all future UI frameworks at SAP, which may not have the same needs as UI5.

ralfhandl:
> But this annotation is meant for all future UI frameworks at SAP, which may not have the same needs as UI5.

How about starting with non-nullable DrillStateProperty and relax this when the first UI framework arrives that doesn't require it?

ralfhandl (approved):

PR #183, opened 2022-05-17
Allow Common.FieldControl on parameters

ralfhandl:
Use case is to mark action parameters as "mandatory".


uhlmannm (approved):

uhlmannm:
> If this applies to parameters, why not also to return types?
The intention was to indicate that action parameters are mandatory. What would FieldControl mean for the return type?

HeikoTheissen (approved):

RaizoC (approved):

ThomasChadzelek (approved):

BerSie (approved):

PR #184, opened 2022-05-24
RecursiveHierarchy as instance annotation

HeikoTheissen:


vocabularies/Hierarchy.md
```diff
@@ -8,15 +8,19 @@ Terms for Hierarchies
...
-[SiblingRankProperty](./Hierarchy.xml#L59:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the sibling rank of a node<br>The sibling rank of a node is the index of the node in the sequence of all nodes with the same parent created by preorder traversal of the hierarchy structure after evaluating the $filter expression in the request excluding any conditions on key properties. The first sibling is at position 0.
-[PreorderRankProperty](./Hierarchy.xml#L63:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|PropertyPath?|Property holding the preorder rank of a node<br>The preorder rank of a node expresses its position in the sequence of nodes created from preorder traversal of the hierarchy structure after evaluating the $filter expression in the request excluding any conditions on key properties. The first node in preorder traversal has rank 0.
+[ExternalKey](./Hierarchy.xml#L52:~:text=<ComplexType%20Name="-,RecursiveHierarchyType,-")|String|The external key value for a node
```
ThomasChadzelek, line 18: Isn't that s.th. which is already part of the ordinary data? We just need to know its name? This is not a computed or dynamic value at all.

HeikoTheissen: Same for `NodeType`?

ThomasChadzelek: Not sure. That might be computed, but I am only guessing. I could imagine cases where you implement logic saying "node type is X if we have property Y available, else node type is Z" w/o having node type as a declared structural property which is persisted in data base. But I'm not a backend guy ;-)

PR #186, opened 2022-06-03
Counter annotation for hierarchical transformations

HeikoTheissen:


ThomasChadzelek (approved):

vocabularies/Hierarchy.json
```diff
@@ -86,6 +86,17 @@
...
+        "StartNodeCounts": {
+            "$Kind": "Term",
+            "$Collection": true,
+            "$Type": "Edm.Int64",
+            "$AppliesTo": [
+                "Collection"
+            ],
+            "@Common.IsInstanceAnnotation": true,
+            "@Core.Description": "Instance annotation on the result of an `$apply` query option containing the number of start nodes for each `ancestors` or `descendants` transformation",
+            "@Core.LongDescription": "The annotation value is an integer collection whose n-th entry is the number of start nodes determined\n          for the n-th `ancestors` or `descendants` transformation in the order in which they occur in the `$apply` query option.\n          Start nodes are determined by the fourth parameter of an `ancestors` or `descendants` transformation."
```
uhlmannm, line 99: Typically the response would contain a collection with the records resulting from $apply plus the other query options. Where would this instance annotation be added? With every element of this collection?

ThomasChadzelek: This should work like "@odata.count", which was actually my source of inspiration. When you use $count=true as a system query option, the response object ("message body" as the JSON format spec calls it) carries this instance annotation next to the "value" property which holds the array of data. See also [Thoughts On Counts](https://jira.tools.sap/browse/CPOUI5ODATAV4-1591)

HeikoTheissen: See also [OData-JSON](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_InstanceAnnotations), example 53.

uhlmannm (approved):

uhlmannm:
Would it make sense to also review this from RAP side?

HeikoTheissen:
Merged into #182.

PR #187, opened 2022-06-09
DataField.Value does not allow mixed-type collections

HeikoTheissen:


vocabularies/UI.json
```diff
@@ -1828,7 +1828,19 @@
                 "@Core.Description": "The data field's value",
                 "@Validation.DerivedTypeConstraint": [
                     "Edm.PrimitiveType",
-                    "Collection(Edm.PrimitiveType)"
+                    "Collection(Edm.Binary)",
+                    "Collection(Edm.Boolean)",
+                    "Collection(Edm.Date)",
+                    "Collection(Edm.DateTimeOffset)",
+                    "Collection(Edm.Decimal)",
```
ralfhandl, line 1835:
```suggestion
                    "Collection(Edm.Decimal)",
                    "Collection(Edm.Double)",
```

If we allow `Edm.Single`, we should also allow `Edm.Double`, given that two of our backends support `Edm.Double` and do not support `Edm.Single`.

ralfhandl (changes requested):

ralfhandl (approved):

BerSie (approved):
OK - these types are a good starting point.

ralfhandl (approved):

uhlmannm (approved):

ThomasChadzelek (approved):

PR #188, opened 2022-07-08
Make term nullability explicit

ralfhandl:


BerSie (dismissed):
From my point of view OK - I mainly checked Common, Communication, PersonalData and UI. But what's the reason for doing it?

ralfhandl:
@qmacro stumbled across `HeaderInfo` being nullable, which doesn't make much sense to me, and is confusing for readers of the vocabulary.

vocabularies/UI.xml
```diff
@@ -55,7 +55,7 @@ a collection of business object instances, e.g. as a list or table.</String>
 
       <!-- Semantic Views / Perspectives -->
 
-      <Term Name="HeaderInfo" Type="UI.HeaderInfoType" AppliesTo="EntityType">
+      <Term Name="HeaderInfo" Type="UI.HeaderInfoType" Nullable="false" AppliesTo="EntityType">
```
HeikoTheissen, line 58: If I want an Employee's header info to be the same as their Manager's header info, I could write:
```jsonc
"Employee": {
  "$Kind": "EntityType",
  "$BaseType": "self.Manager",
  "Manager": {
    "$Kind": "NavigationProperty",
    "$Type": "self.Manager"
  },
  "@UI.HeaderInfo": {"$Path": "Manager/@UI.HeaderInfo"}, ...
}
```
But if the Employee has no Manager, the `$Path` would evaluate to a forbidden null value.

Shouldn't annotations that might be used with dynamic values be nullable?

ralfhandl: Correct, but does that make sense for HeaderInfo? How would an Object Page without HeaderInfo look like?

ralfhandl: Terms that legally can have a null value, for example Common.FilterDefaultValue, are now explicitly marked as Nullable.

Same for terms that are explicitly defined to be used with path expressions.

HeikoTheissen: > Correct, but does that make sense for HeaderInfo? How would an Object Page with a null HeaderInfo look like?

The client could use a default, for example list all fields alphabetically.

ralfhandl: @BerSie what will the client "Fiori Elements" do if it encounters `@UI.HeaderInfo: null`?

qmacro (commented):
From a consumer / reader POV this change makes a lot of sense to me, thanks. I'd wondered about the `?` suffix on types for the terms in our vocabs and remained vaguely confused for a good while.

qmacro:
Background on where my questions about the `?` suffix came from - writing a small guide to understanding annotations, specifically this bit https://github.com/SAP-samples/odata-basics-handsonsapdev/blob/annotations/bookshop/README.md#the-ui-annotation-vocabulary

qmacro (dismissed):
LGTM, thanks!

vocabularies/UI.xml
BerSie: There might be places where this is handled gracefully, but I wouldn't bet on that. Already today the vocabulary states that the HeaderInfo is mandatory for the "main entity types". And it's not only properties , where can take it e.g. alphabetically from a list, it's also the object type name. From a consumption (and UX) perspective I regards the header information as mandatory.

GeraldKrause (commented):
Checked Analytics vocabulary; looks good to me

ralfhandl:
Merged `main` branch because `Common.DefaultValuesFunction` is no longer experimental, see #189

qmacro (approved):

HeikoTheissen (approved):

PR #192, opened 2022-07-20
Common.DraftXxx: sync with RAP implementation

ralfhandl:
1. DiscardAction is no longer experimental
2.	AdditionalNewActions  is no longer  experimental
3.	EditAction has a new optional parameter „PreserveChanges“ (Edm.Boolean)
  -  The action can provide an optional parameter “PreserveChanges” of type “Edm.Boolean” that may be used to ensure outdated drafts of another user are not overwritten (see below).
4.	NewAction and AdditionalNewActions have an optional parameter “ResultIsActiveEntity” (Edm.Boolean) 
  - The action can provide an optional parameter "ResultIsActiveEntity" of type Edm.Boolean to indicate whether via the action a draft or an active instance should be created. Default is “false”, i.e. the creation of a draft instance.
5.	DraftNode: PreparationAction is deprecated
  - [x] @RaizoC wonders if the complete annotation “DraftNode” is deprecated as it is always empty (in the meantime). That a node is activated via and belongs to the draft root is indicated via a separate annotation `DraftActivationVia`. @MarcelWaechter: can you confirm the annotation “DraftNode” is NOT used any more in FE V4 (and V2)?
6.	ShareAction has a new parameter “ShareAll” (Edm.Boolean)


RaizoC (approved):

vocabularies/Common.xml
```diff
@@ -1069,37 +1069,55 @@ The minimum structure is
...
         <Property Name="DiscardAction" Type="Common.QualifiedName" Nullable="true">
-          <Annotation Term="Common.Experimental" />
           <Annotation Term="Core.Description" String="Action that discards a draft document" />
         </Property>
         <Property Name="EditAction" Type="Common.QualifiedName" Nullable="true">
           <Annotation Term="Core.Description" String="Action that creates an edit draft" />
+          <Annotation Term="Core.LongDescription">
+            <String>The action is bound to the active document root node and has the following signature:
+
+            - `PreserveChanges`: optional, of type `Boolean`, `true` means that an outdated draft of another user is not overwritten</String>
```
HeikoTheissen, line 1086:
```suggestion
- `PreserveChanges`: optional, of type `Boolean`, `true` means that an outdated draft of another user is not overwritten</String>
```
You want a bulleted list in Markdown, right?

vocabularies/Common.xml
```diff
@@ -1069,37 +1069,55 @@ The minimum structure is
...
         <Property Name="NewAction" Type="Common.QualifiedName" Nullable="true">
           <Annotation Term="Core.Description" String="Action that creates a new draft" />
-          <Annotation Term="Core.LongDescription" String="New drafts may also be created by POSTing an empty entity without any properties to the entity set." />
+          <Annotation Term="Core.LongDescription" String="" />
+          <Annotation Term="Core.LongDescription">
+            <String>The action is bound to the draft document root entity set and has the following signature:
+
+            - `ResultIsActiveEntity` of type `Boolean` with default `false`, `true` means that an active entity is created
+
+            New drafts may also be created by POSTing an entity with property IsActiveEntity=false (default) to the entity set.</String>
```
HeikoTheissen, line 1097:
```suggestion
- `ResultIsActiveEntity` of type `Boolean` with default `false`, `true` means that an active entity is created

New drafts may also be created by POSTing an entity with property IsActiveEntity=false (default) to the entity set.</String>
```

vocabularies/Common.xml
```diff
@@ -1069,37 +1069,55 @@ The minimum structure is
...
         </Property>
         <Property Name="AdditionalNewActions" Type="Collection(Common.QualifiedName)" Nullable="false">
-          <Annotation Term="Common.Experimental" />
-          <Annotation Term="Core.Description" String="Additional actions that create a new draft" />
-          <Annotation Term="Core.LongDescription" String="Additional actions beside the default POST or standard `NewAction` that create a new draft." />
+          <Annotation Term="Core.Description" String="Additional actions beside the default POST or standard `NewAction`that create a new draft" />
+          <Annotation Term="Core.LongDescription">
+            <String>The actions are bound to the draft document root entity set and have the following signature:
+
+            - `ResultIsActiveEntity` of type `Boolean` with default `false`, `true` means that an active entity is created</String>
```
HeikoTheissen, line 1105:
```suggestion
- `ResultIsActiveEntity` of type `Boolean` with default `false`, `true` means that an active entity is created</String>
```

vocabularies/Common.xml
```diff
@@ -1069,37 +1069,55 @@ The minimum structure is
...
+          </Annotation>
         </Property>
         <Property Name="ShareAction" Type="Common.QualifiedName" Nullable="true">
-          <Annotation Term="Core.Description" String="Action that shares a draft document with other users" />
+          <Annotation Term="Core.Description" String="Action that shares a draft document with other users and restricts access to the listed users in their specified roles" />
           <Annotation Term="Core.LongDescription">
             <String>The action is bound to the draft document root node and has the following signature:
 
             - `Users`: collection of structure with properties
```
HeikoTheissen, line 1114:
```suggestion
- `Users`: collection of structure with properties
```

vocabularies/Common.xml
```diff
@@ -1069,37 +1069,55 @@ The minimum structure is
...
 
-            &#xA0;&#xA0;- `UserID` of type `String` and
+            &#xA0;&#xA0;&#x25E6; `UserID` of type `String`
```
HeikoTheissen, line 1115:
```suggestion
  - `UserID` of type `String`
```

vocabularies/Common.xml
```diff
@@ -1069,37 +1069,55 @@ The minimum structure is
...
 
-            &#xA0;&#xA0;- `UserAccessRole` of type `String` with possible values `O` (owner, can perform all draft actions), and `E` (editor, can change the draft)
+            &#xA0;&#xA0;&#x25E6; `UserAccessRole` of type `String` with possible values `O` (owner, can perform all draft actions), and `E` (editor, can change the draft)
```
HeikoTheissen, line 1117:
```suggestion
  - `UserAccessRole` of type `String` with possible values `O` (owner, can perform all draft actions), and `E` (editor, can change the draft)
```

vocabularies/Common.xml
```diff
@@ -1069,37 +1069,55 @@ The minimum structure is
...
 
-            It restricts access to the listed users in their specified roles.</String>
+            - `ShareAll` of type `Boolean` with default value `false`, `true` means sharing with all users that are authorized to create drafts;
+            the `Users` parameter is optional in this case, provide it to actively notify the listed users</String>
```
HeikoTheissen, line 1120:
```suggestion
- `ShareAll` of type `Boolean` with default value `false`, `true` means sharing with all users that are authorized to create drafts; the `Users` parameter is optional in this case, provide it to actively notify the listed users</String>
```

HeikoTheissen (changes requested):
Purely typographical review

RaizoC:
If we do not deprecate the complete "DraftNode" (which is only used as tag and thus redundant to "DraftActivationVia", we should at least remove or deprecate the ValidationFunction inside that was never used.

ralfhandl:
> we should at least remove or deprecate the ValidationFunction

That is already deprecated:
https://github.com/SAP/odata-vocabularies/blob/c941c4ef39997efcdcd92f93e41e61e26a203f80/vocabularies/Common.xml#L1136-L1146

ralfhandl:
> @RaizoC Should we deprecate the whole `DraftNode` now?

Deprecated the term

RaizoC:
> > @RaizoC Should we deprecate the whole `DraftNode` now?
> 
> Deprecated the term

I would deprecate it as it only adds redundancy, but we @MarcelWaechter should confirm.

HeikoTheissen (approved):

HeikoTheissen (approved):

PR #193, opened 2022-07-28
Allow annotating UI.DataPointType records with UI.Hidden

ralfhandl:


RaizoC (dismissed):

BerSie (commented):
I have nothing against making this possible usage explicit, but in principal it's already possible to use the UI.Hidden annotation on DataPointType records or KPIType records (UI.Hidden applies to Record amongst others).
And in addition: (at least) DataPoints are usually referenced either in a UI.DataField or in a UI.ReferenceFacet. And here it's already possible to use the UI.Hidden annotation. If it's already correctly interpreted by the consuming client framework is something completely different.

ralfhandl:
@BerSie and @RaizoC please re-approve so we can merge

BerSie (commented):
Maybe my comment was a bit too defensive :-) - it would be good to know why we now emphasize this usage. In principle it was already "technically" possible (as UI.Hidden applies to Record) and the usage of UI.Datapoint in DataFields or Facets can have UI.Hidden annotations as well. And there are a lot of Records (that are used in DataFields or Facets) that currently doesn't have this applicable term. What's the difference?

ralfhandl:
`UI.Hidden` applies to `Record`, which is sufficient for clients, even if maybe a bit over-generic.

PR #197, opened 2022-09-12
Fix: reference Aggregation and Analytics vocabularies in UI vocabulary

ralfhandl:
Validation.AllowedTerms references terms from these vocabularies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,9 +17,9 @@
       }
     },
     "node_modules/@eslint/eslintrc": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/@eslint/eslintrc/-/eslintrc-1.3.1.tgz",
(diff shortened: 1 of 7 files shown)


UweSchwerk (changes requested):
What about Common - it had a similar issue ?

HeikoTheissen:
> What about Common - it had a similar issue ? (Common.SortOrderType/DynamicProperty)

The Common vocabulary is already referenced in the UI vocabulary.

HeikoTheissen (approved):

ralfhandl:
And the UI vocabulary is referenced from Common.

BerSie:
Hmm, I would expect that the Analytics vocabulary should be referenced explicitly if a term if this vocabulary is used. So, I think, Uwe is right.

ralfhandl:
> `Common.SortOrderType/DynamicProperty` is annotated with `Validation.AllowedTerms` containing `Analytics.AggregatedProperty` - even though `Analytics` is not defined

Thanks, I missed that one, reference added.

UweSchwerk (approved):

HeikoTheissen (approved):

HeikoTheissen (approved):

PR #198, opened 2022-09-26
UI.IsCopyAction: improved descriptions

ralfhandl:


vocabularies/UI.xml
```diff
@@ -1415,9 +1415,9 @@ The trend is
       <Term Name="IsCopyAction" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Record">
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.RequiresType" String="UI.DataFieldForAction" />
-        <Annotation Term="Core.Description" String="The annotated `DataFieldForAction` record within a `UI.LineItem` annotation references an action that deep-copies an instance of the annotated entity type" />
+        <Annotation Term="Core.Description" String="The annotated [`DataFieldForAction`](#DataFieldForAction) record within a [`UI.LineItem`](#LineItem) annotation references an action that deep-copies an instance of the annotated entity type" />
```
BerSie, line 1418: Do we need the restriction to UI.LineItem?

ralfhandl: No, this was just the discussed use case.

What other places might this be useful, and how would Fiori Elements deal with the annotation there?

BerSie: I think we could have it in the UI.Identification (it would occur in the Object Page header). It would make sense to have the copy function in the OP as well, correct? Not sure if that would be already supported.

HeikoTheissen (approved):

BerSie (approved):

PR #199, opened 2022-09-26
UI.LeadingEntitySet

ralfhandl:
Mark an entity set as the preferred entry point for UIs

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1769,6 +1769,7 @@ This switches off the validation of user input with respect to decimals.</String
       </Term>
 
       <Term Name="LeadingEntitySet" Type="Edm.String" Nullable="false" AppliesTo="EntityContainer">
+        <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="The referenced entity set is the preferred starting point for UIs using this service" />
       </Term>
 
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1768,6 +1768,10 @@ This switches off the validation of user input with respect to decimals.</String
         </Annotation>
       </Term>
 
+      <Term Name="LeadingEntitySet" Type="Edm.String" Nullable="false" AppliesTo="EntityContainer">
+        <Annotation Term="Core.Description" String="The referenced entity set is the preferred starting point for UIs using this service" />
+      </Term>
+
     </Schema>
   </edmx:DataServices>
 </edmx:Edmx>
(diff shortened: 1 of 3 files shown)


vocabularies/UI.xml
```diff
@@ -1768,6 +1768,10 @@ This switches off the validation of user input with respect to decimals.</String
         </Annotation>
       </Term>
 
+      <Term Name="LeadingEntitySet" Type="Edm.String" Nullable="false" AppliesTo="EntityContainer">
+        <Annotation Term="Core.Description" String="The referenced entity set is the preferred starting point for UIs using this service" />
```
HeikoTheissen, line 1772: For me, the "starting point for a UI" is not an entity set, but an annotation on an entity set/type, for example
```
/ns.Schema/LeadingEntityType@UI.LineItem#professional
```
This instructs the UI not just which entity set to read, but also how to display it (in a table, not an object page) and in which table configuration (professional).

The path to the "starting annotation" (example given above) should be specified in the app descriptor, not in the service metadata. Because if a UI loads more than one service (each with its own LeadingEntitySet), it still needs a rule where to start.

HeikoTheissen (changes requested):

vocabularies/UI.xml
ralfhandl: Creating a BAS project for a Fiori Elements app will eventually ask you which entity set to start from:
image

The selected entity set ends up in the app descriptor:
~~~json
    "pages": {
      "ListReport|Zrha_Airline": {
        "entitySet": "Zrha_Airline",
~~~

Annotations on this entity set and/or its entity type kick in later, the starting point is an entity set.

The `UI.LeadingEntitySet` could be used as a suggestion in this project creation wizard, currently the lexicographically first entity set seems to get suggested.

BerSie (approved):

vocabularies/UI.xml
HeikoTheissen: > The UI.LeadingEntitySet could be used as a suggestion in this project creation wizard, currently the lexicographically first entity set seems to get suggested.

You mean the lexicographically first entity is the first in the dropdown box, but the user can choose any other, right?

ralfhandl: Yes, and it would be nicer if the default were what the service designer intended as the leading entity set.

HeikoTheissen (approved):

RaizoC (approved):

BerSie (approved):

PR #204, opened 2022-11-21
New complext type Offline.ClientOnlyType

bruce-qin-dev:


vocabularies/Offline.xml
```diff
@@ -49,7 +49,21 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
...
+        </Annotation>
+      </Term>
+
+      <Term Name="LocalDraft" Type="Core.Tag" DefaultValue="true" Nullable="false" AppliesTo="EntitySet">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The annotated model is draft and exists only on client devices. A draft entity
+        can be converted to non-draft entity for uploading"/>
+         <Annotation Term="Core.LongDescription">
+           <String>Typically for each entity type there is one corresponding entity set, however OData permits having multiple entity sets sharing the same entity type.
+            This can be taken advantage of to enable a facility for local (offline) drafts by defining an additional (draft) entity set for an existing entity type.
```
HeikoTheissen, line 61: But this differs from the existing draft concept, where drafts and active entities are in the same entity set.

smartev3: This Offline vocabulary is used to support Client Only entity types (entity types that don't exist in the backend) and Local Draft entity sets (the entity type exists in the backend but has additional entity set(s) that assist with the implementation of offline drafts). For online drafts, we wouldn't need additional entity sets.

vocabularies/Offline.xml
```diff
@@ -49,7 +49,21 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
...
+            For example, with existing entity type "Order" and existing entity set "Orders", an additional "DraftOrders" entity set could be defined, annotated with
+            this annotation. When a client application creates an entity instance, it could be created in the "DraftOrders" entity set. A subsequent offline entity
+            upload operation should omit the uploading of any local draft entities. The DataService.activateDraft API can be used to transition a draft entity
```
ralfhandl, line 64: Where is the `DataService.activateDraft` API defined/documented? Please add a link

ralfhandl (changes requested):

vocabularies/Offline.xml
```diff
@@ -49,7 +49,21 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
...
+            (and optionally its children) into non-draft entities so they will then become uploadable.</String>
```
ralfhandl, line 65:
```suggestion
            upload operation will not upload any local draft entities.
            Draft entities and their children need to be transitioned into non-draft entities to become uploadable.</String>
```

vocabularies/Offline.xml
HeikoTheissen: What can't you simply tag the entity sets that are used for local drafts as `Offline.ClientOnly`? Can there be a client-only entity set that is _not_ used for local drafts?

bruce-qin-dev: There could be client-only entity set which can't be used as local drafts.
The client only entity type  and client only entity set can be defined from the client side, and the server may don't event know the existence of some client-only entity set.

vocabularies/Offline.xml
```diff
@@ -49,7 +49,21 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
            attempt to upload any changes that are made to them locally.
            The service as implemented on the server is described by the CSDL document if
            all elements tagged as `Offline.ClientOnly` are removed together with all subsequently dangling references.</String>
-        </Annotation>        
+        </Annotation>
+      </Term>
+
+      <Term Name="LocalDraft" Type="Core.Tag" DefaultValue="true" Nullable="false" AppliesTo="EntitySet">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The annotated model is draft and exists only on client devices. A draft entity
```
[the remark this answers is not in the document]
bruce-qin-dev: No, LocalDraft does not imply ClientOnly.
LocalDraft does imply that there is a sever entity set(aka:non-draft) which reference to the same entity type as the draft.

vocabularies/Offline.xml
bruce-qin-dev: API reference removed as suggested, thanks a lot for the advice.

vocabularies/Offline.xml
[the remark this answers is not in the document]
ralfhandl: Your description says "The annotated model is draft and exists only on client devices".

This means that each draft model is client-only.

Please adapt the description if it is not correct.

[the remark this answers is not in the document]
smartev3: Here's a possible alternate Core.Description:

"Marks a local draft version of an entity set, which uses the same entity type as a non-draft entity set defined in the backend defined metadata."

Anyone needing more clarity can read the LongDescription.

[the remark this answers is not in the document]
bruce-qin-dev: Thanks a lot,  the short description is updated to avoid ambiguity.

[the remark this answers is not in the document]
HeikoTheissen: > "Marks a local draft version of an entity set, which uses the same entity type as a non-draft entity set defined in the backend defined metadata."

If two backend-defined entity sets use the same entity type and a third entity set tagged as `Offline.LocalDraft` uses the same entity type, it is unclear to which of the two other entity sets it corresponds. Should `Offline.LocalDraft` perhaps contain the name of the backend-defined entity set rather than being a boolean?

vocabularies/Offline.xml
```diff
@@ -49,7 +49,20 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
            attempt to upload any changes that are made to them locally.
            The service as implemented on the server is described by the CSDL document if
            all elements tagged as `Offline.ClientOnly` are removed together with all subsequently dangling references.</String>
-        </Annotation>        
+        </Annotation>
+      </Term>
+
+      <Term Name="LocalDraft" Type="Core.Tag" DefaultValue="true" Nullable="false" AppliesTo="EntitySet">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Marks a local draft version of an entity set, which uses the same entity type as a non-draft entity set defined in the backend defined metadata."/>
```
HeikoTheissen, line 57: If two backend-defined entity sets use the same entity type and a third entity set tagged as `Offline.LocalDraft` uses the same entity type, it is unclear to which of the two other entity sets it corresponds. Should `Offline.LocalDraft` perhaps contain the name of the backend-defined entity set rather than being a boolean?

HeikoTheissen (changes requested):

vocabularies/Offline.xml
smartev3: The API we are providing for clients when "activating" a draft allows the client to indicate the "active" entity set into which the draft will be activated (only necessary to indicate this if there is an ambiguity).

However your question/suggestion is a good one, perhaps we can change it to Type="Edm.String" Nullable="true" without a default value (the implicit default of null can then indicate that the draft set is linked with the only other entity set using the same entity type).

bruce-qin-dev: The term is now change to <Term Name="LocalDraft" Type="Edm.String" Nullable="true" AppliesTo="EntitySet">
thanks a lot for the suggestions.

ralfhandl: I like it

ralfhandl (approved):

vocabularies/Offline.xml
```diff
@@ -52,7 +52,7 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
         </Annotation>
       </Term>
 
-      <Term Name="LocalDraft" Type="Core.Tag" DefaultValue="true" Nullable="false" AppliesTo="EntitySet">
+      <Term Name="LocalDraft" Type="Edm.String" Nullable="true" AppliesTo="EntitySet">
```
HeikoTheissen, line 55:
```suggestion
      <Term Name="LocalDraft" Type="Edm.String" Nullable="false" AppliesTo="EntitySet">
```
What would `"@Offline.LocalDraft": null` mean? Isn't it sufficient to omit the annotation if there is no local draft?

HeikoTheissen (changes requested):

vocabularies/Offline.xml
bruce-qin-dev: If there is no local-draft, then the annotation should of course be removed.
Regarding null value or default value, here I quote Evan's reply from the conversation above:
"the implicit default of null can then indicate that the draft set is linked with the only other entity set using the same entity type."

HeikoTheissen (approved):

vocabularies/Offline.xml
ralfhandl: Actually this is brittle: adding another entity set with the same type is a compatible change, but it would break the annotation.

It also puts burden on the client: look for the only other set with the same type, instead of just writing down what the author of the annotation already knows: the name of that other entity set.

vocabularies/Offline.xml
```diff
@@ -49,7 +49,21 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
            attempt to upload any changes that are made to them locally.
            The service as implemented on the server is described by the CSDL document if
            all elements tagged as `Offline.ClientOnly` are removed together with all subsequently dangling references.</String>
-        </Annotation>        
+        </Annotation>
+      </Term>
+
+      <Term Name="LocalDraft" Type="Edm.String" Nullable="true" AppliesTo="EntitySet">
```
ralfhandl, line 55:
```suggestion
      <Term Name="LocalDraft" Type="Edm.String" Nullable="false" AppliesTo="EntitySet">
```

ralfhandl (changes requested):

vocabularies/Offline.xml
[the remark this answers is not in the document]
HeikoTheissen: Sorry, I still don't understand. You say that a `LocalDraft` entity set need not be `ClientOnly`. But then it exists also on the server, right? Can you give an example for this?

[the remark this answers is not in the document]
bruce-qin-dev: Hello Heiko,

Sorry for the confusion,  LocalDraft and ClientOnly are different concept here.
OfflineOData is basically client-server structure. Users download server data to client device, and then make all odata CRUD request against local dataset. (This is a big difference with online-odata which operates on back-end dataset directly).

The odata metadata for OfflineOData client device is a combination of metadata downloaded from back-end and metadata defined from the client side. ClientOnly and LocalDraft are two terms defined in the client side. ClientOnly means some data which exists in client device only, and should never upload to back-end. The entity type annotated with ClientOnly could be defined from the client side,  which back-end may not aware of. 
The LocalDraft indicates that the underlying entity type is recognizable by back-end, and the draft could be covered to non-draft and upload to back-end later.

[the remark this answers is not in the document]
HeikoTheissen: For me this means that
* The local dataset on the client can contain both draft and non-draft entities.
* Draft entities can exist in the local dataset on the client as well as on the back-end.

Correct?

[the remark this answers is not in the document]
bruce-qin-dev: That's correct, thanks. Just to mention that the draft entities are converted to non-draft entities from the client side  before upload to back-end, and the conversion is done by client app which is transparent to back-end. So technically back-end will only see non-draft entities, thanks.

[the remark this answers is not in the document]
HeikoTheissen: > So technically back-end will only see non-draft entities, thanks.

But why is the LocalDraft entity set (`SalesOrderDraft` in your example) then needed in the back-end?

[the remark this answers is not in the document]
bruce-qin-dev: Hi Heiko,

I though the sample is used to demonstrate the usage the term.
If I'm wrong I will delete all the samples, sorry for my misunderstanding.

vocabularies/Offline.xml
```diff
@@ -49,7 +49,21 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
...
+         <Annotation Term="Core.LongDescription">
+           <String>Typically for each entity type there is one corresponding entity set, however OData permits having multiple entity sets sharing the same entity type.
+            This can be taken advantage of to enable a facility for local (offline) drafts by defining an additional (draft) entity set for an existing entity type.
+            For example, with existing entity type "Order" and existing entity set "Orders", an additional "DraftOrders" entity set could be defined, annotated with
+            this annotation. When a client application creates an entity instance, it could be created in the "DraftOrders" entity set. A subsequent offline entity
+            upload operation will not upload any local draft entities.
+            Draft entities and their children need to be transitioned into non-draft entities to become uploadable.
+            The linked non-draft entity set using the same entity type can be defined in the string value, and the implicit default value of null indicates that the draft set is linked with the only other non-draft entity set using the same entity type.</String>
+        </Annotation>
```
ralfhandl, line 66:
```suggestion
        <Annotation Term="Core.Description" String="Marks a local draft version of an entity set, which uses the same entity type as a non-draft entity set defined in the backend defined metadata."/>
         <Annotation Term="Core.LongDescription">
           <String>Typically for each entity type there is one corresponding entity set, however OData permits having multiple entity sets sharing the same entity type.
            This can be taken advantage of to enable a facility for local (offline) drafts by defining an additional local draft entity set for an existing server-side entity set and entity type.
            For example, with a server-side entity set `Orders` and entity type `Order`, an additional `DraftOrders` entity set could be defined, annotated with this annotation. 
            The annotation value is the name of the non-draft entity set using the same entity type.
            When a client application creates an entity instance, it could be created locally in the `DraftOrders` entity set. 
            A subsequent offline entity upload operation will not upload any local draft entities.
            Draft entities and their children need to be transitioned into non-draft entities to become uploadable.</String>
        </Annotation>
```

vocabularies/Offline.xml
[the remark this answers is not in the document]
HeikoTheissen: My question is not only about your sample: If the back-end will only see non-draft entities, why is the LocalDraft entity set then needed in the back-end?

ralfhandl (dismissed):

vocabularies/Offline.xml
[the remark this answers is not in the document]
bruce-qin-dev: I don't think LocalDraft entity set is needed in the back-end. It is the client app which needs the LocalDraft entity set, thanks.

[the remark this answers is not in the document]
ralfhandl: > why is the LocalDraft entity set then needed in the back-end?

My understanding: the LocalDraft entity set exists only locally on the client, and the LocalDraft-sample shows the client-local schema.

@bruce-qin-dev @HeikoTheissen Should we have both `LocalDraft-sample-client` and `LocalDraft-sample-server` files to show the difference?

vocabularies/Offline.xml
ralfhandl: That's my understanding: local draft entity sets are a specific flavor of ClientOnly.

examples/Offline.LocalDraft-sample.xml
```diff
@@ -0,0 +1,50 @@
...
+        <!--SalesOrderDraft is sharing the SalesOrderType, and is annotated as LocalDraft-->
+        <EntitySet Name="SalesOrderDraft" EntityType="client.SalesOrderType">
+          <Annotation Term="Offline.LocalDraft" String="SalesOrder" />
+          <NavigationPropertyBinding Path="_Item" Target="SalesOrderItemDraft" />
+        </EntitySet>
+
+        <!--SalesOrderItemDraft is sharing the SalesOrderItemType, and is annotated as LocalDraft-->
+        <EntitySet Name="SalesOrderItemDraft" EntityType="client.SalesOrderItemType">
+          <Annotation Term="Offline.LocalDraft" String="SalesOrderItem" />
+        </EntitySet>
```
HeikoTheissen, line 46:
```suggestion
        <EntitySet Name="SalesOrderItemDraft" EntityType="client.SalesOrderItemType">
          <Annotation Term="Offline.LocalDraft" String="SalesOrderItem" />
          <Annotation Term="Offline.ClientOnly" />
        </EntitySet>
```

vocabularies/Offline.xml
```diff
@@ -49,7 +49,21 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
...
            all elements tagged as `Offline.ClientOnly` are removed together with all subsequently dangling references.</String>
-        </Annotation>        
+        </Annotation>
+      </Term>
+
+      <Term Name="LocalDraft" Type="Edm.String" Nullable="false" AppliesTo="EntitySet">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Marks a local draft version of an entity set, which uses the same entity type as a non-draft entity set defined in the backend defined metadata."/>
+         <Annotation Term="Core.LongDescription">
+           <String>Typically for each entity type there is one corresponding entity set, however OData permits having multiple entity sets sharing the same entity type.
```
HeikoTheissen, line 59:
```suggestion
           <String>A entity set annotated with this term must also be tagged as [`ClientOnly`](#ClientOnly).
           
           Typically for each entity type there is one corresponding entity set, however OData permits having multiple entity sets sharing the same entity type.
```

HeikoTheissen (changes requested):
LocalDraft requires ClientOnly. Make this requirement explicit.

examples/Offline.LocalDraft-sample.xml
HeikoTheissen: Alternatively, combine ClientOnly and LocalDraft into one annotation, see #214.

HeikoTheissen (changes requested):
Delete the LocalDraft examples.

vocabularies/Offline.xml
```diff
@@ -37,20 +34,42 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
...
-           The service as implemented on the server is described by the CSDL document if
-           all elements tagged as `Offline.ClientOnly` are removed together with all subsequently dangling references.</String>
-        </Annotation>        
+           <String>
+Occasionally a customer will want to store additional “client-only” entities in the same database.
+The easiest way to accommodate this is for the client to extend the backend-returned $metadata
+(upon which the local persistence is based) and to mark some entities as client-only so that the
+client won’t attempt to upload any changes that are made to them locally.
+The service as implemented on the server is described by the CSDL document if all elements tagged
+as `Offline.ClientOnly` are removed together with all subsequently dangling references.
```
HeikoTheissen, line 47:
```suggestion
The service as implemented on the server is described by the CSDL document if all elements annotated
with `Offline.ClientOnly` are removed together with all subsequently dangling references.
```
Because `ClientOnly` is no longer a tag.

HeikoTheissen (changes requested):

HeikoTheissen (approved):

ralfhandl (approved):
Are we ready to merge?

bruce-qin-dev:
> Are we ready to merge?

Yes, please help to merge it. Thanks a lot!

PR #205, opened 2022-11-25
Add padding flag to PDF features.

bernhardgr42:
Add padding flag to PDF. New boolean value to see if the server supports the "padding" option. 

"Sets the padding (left, right, bottom, top) in PDF units. The value must be > 0."

### DIFF
--- a/vocabularies/PDF.xml
+++ b/vocabularies/PDF.xml
@@ -77,7 +77,7 @@
         <Property Name="Padding" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
           <Annotation Term="Core.Description" String="Padding of document" />
           <Annotation Term="Core.LongDescription">
-            <String>Sets the padding (left, right, bottom, top) in PDF units</String>
+            <String>Is padding (left, right, bottom, top) supported?</String>
           </Annotation>
         </Property>
         <Property Name="ResultSizeDefault" Type="Edm.Int32">
--- a/vocabularies/PDF.xml
+++ b/vocabularies/PDF.xml
@@ -74,6 +74,12 @@
             <String>If this option is selected, the font size is automatically selected in such a way that all columns of a table fit on one page. Other layout options like margin, border and composite cell spacing are adapted accordingly, with respect to the chose scaling factor.</String>
           </Annotation>
         </Property>
+        <Property Name="Padding" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="Padding of document" />
+          <Annotation Term="Core.LongDescription">
+            <String>Sets the padding (left, right, bottom, top) in PDF units. The value must be > 0</String>
+          </Annotation>
+        </Property>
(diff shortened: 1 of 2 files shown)


bernhardgr42 (commented):
@ralfhandl Pls, can you review this?

bernhardgr42:
@ralfhandl Is this ok?

vocabularies/PDF.xml
```diff
@@ -74,6 +74,12 @@
             <String>If this option is selected, the font size is automatically selected in such a way that all columns of a table fit on one page. Other layout options like margin, border and composite cell spacing are adapted accordingly, with respect to the chose scaling factor.</String>
           </Annotation>
         </Property>
+        <Property Name="Padding" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
```
HeikoTheissen, line 77:
```suggestion
        <Property Name="Padding" Type="Edm.Decimal" Nullable="false" DefaultValue="0">
```
If the value must be > 0, this cannot be a boolean value.

HeikoTheissen (changes requested):

ralfhandl (changes requested):
Same comment as @HeikoTheissen

vocabularies/PDF.xml
bernhardgr42: It must be a boolean.

We can have a separate Frontend server. The annotation is to let the client know, if the server supports this option. This is the same alike FontName, FontSize and so on.

The actual type of the Padding is for sure an integer. But this is described in the separated OData service ($metadata) where you can set the format options like fontName, size and also padding

vocabularies/PDF.xml
```diff
@@ -77,7 +77,7 @@
         <Property Name="Padding" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
           <Annotation Term="Core.Description" String="Padding of document" />
           <Annotation Term="Core.LongDescription">
-            <String>Sets the padding (left, right, bottom, top) in PDF units. The value must be > 0</String>
+            <String>Sets the padding (left, right, bottom, top) in PDF units</String>
```
ralfhandl, line 80: Shouldn't the text rather be something like
```suggestion
            <String>Padding (left, right, bottom, top) is supported</String>
```

ralfhandl (changes requested):

vocabularies/PDF.xml
bernhardgr42: Yes right. 

==> Changed the description.

HeikoTheissen (approved):

ralfhandl (approved):

vocabularies/PDF.xml
ralfhandl, line 77: Is this property already shipped, or is it still experimental?

ralfhandl (changes requested):

ralfhandl (approved):

ralfhandl:
@bernhardgr42 Can we merge this PR?

bernhardgr42:
Yes please

PR #212, opened 2022-12-15
UpdateAction

HeikoTheissen:


ralfhandl (commented):
It would be nice to have a matching JSON example of an aggregation result with instance annotations.

vocabularies/Analytics.xml
```diff
@@ -167,6 +167,60 @@
...
+            </Annotation>
+          </Record>
+        </Annotation>
+      </Term>
+      <ComplexType Name="UpdateActionType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="ActionImport" Type="Core.SimpleIdentifier" Nullable="false">
+          <Annotation Term="Core.Description" String="Name of an action import" />
+        </Property>
+        <Property Name="Aggregate" Type="Edm.String" Nullable="true">
```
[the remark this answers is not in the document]
HeikoTheissen: Then the action has no parameter for identifying the aggregate entity, but encodes the entity in its action name. For very simple use cases, this might be feasible.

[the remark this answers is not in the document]
ralfhandl: Do we have concrete examples of this very simple use case?

If not, I'd start with not nullable, and we could relax that later.

examples/Analytics.UpdateAction-result.json
```diff
@@ -0,0 +1,75 @@
...
+      "Customer": {
+        "Country": "Netherlands"
+      },
+      "Product": {
+        "Name": "Paper"
+      },
+      "Amount": 3,
+      "@Analytics.UpdateAction": {
+        "ActionImport": "UpdateAggregateSales",
+        "Aggregate": "groupby((Customer/Country,Product/Name),aggregate(Amount))/filter(Customer/Country eq 'Netherlands' and Product/Name eq 'Paper')",
```
ralfhandl, line 14: I'd use something cryptic here to push the point that this is an opaque string.

Examples are easily misinterpreted as specifications.

```suggestion
        "Aggregate": "8C@FA3JWWrFDE@>6C^r@F?ECJ[!C@5F4E^}2>6X[288C682E6Wp>@F?EXX^7:=E6CWrFDE@>6C^r@F?ECJ 6B V}6E96C=2?5DV 2?5 !C@5F4E^}2>6 6B V!2A6CVX",
```

examples/Analytics.UpdateAction-result.json
```diff
@@ -0,0 +1,75 @@
+{
+  "@odata.context": "$metadata#SalesModel.SalesData/Sales(Customer(Country),Product(Name),Amount)",
```
HeikoTheissen, line 2: Is this the correct way to address an entity set from a referenced schema?

ralfhandl: I would expect

```suggestion
  "@odata.context": "$metadata#Sales(Customer(Country),Product(Name),Amount)",
```

because this response comes from the service that serves the `Sales` entity set.

Which means that the full example wouldn't work: if the response from `GET Sales?$apply=...` just mentions an action import name in the instance annotation, that action import must be in the same service / entity container as the entity set that is being aggregated.

examples/Analytics.UpdateAction-sample.xml
```diff
@@ -0,0 +1,35 @@
...
+        <Parameter Name="Changes" Type="SalesModel.Sale" Nullable="false">
+          <Annotation Term="Core.Description" String="Aggregate entity with nested properties that are to be updated, like the body of a PATCH request" />
+        </Parameter>
+        <ReturnType Type="SalesModel.Sale" Nullable="false">
+          <Annotation Term="Core.Description" String="Aggregate entity with updates applied" />
+        </ReturnType>
+      </Action>
+      <EntityContainer Name="container">
+        <ActionImport Name="UpdateAggregateSales" Action="this.UpdateAggregateSales" EntitySet="SalesModel.SalesData/Sales" />
+      </EntityContainer>
```
ralfhandl, line 29: This won't work, the entity set producing an aggregated response with instance annotations needs to be in the same service / container as the action import that can be called on its aggregated results.

examples/Analytics.UpdateAction-sample.xml
```diff
@@ -0,0 +1,35 @@
...
+        <ReturnType Type="SalesModel.Sale" Nullable="false">
+          <Annotation Term="Core.Description" String="Aggregate entity with updates applied" />
+        </ReturnType>
+      </Action>
+      <EntityContainer Name="container">
+        <EntitySet Name="Sales" EntityType="SalesModel.Sale">
+          <Annotation Term="Aggregation.CustomAggregate" Qualifier="Amount" String="Edm.Decimal" />
+        </EntitySet>
+        <ActionImport Name="UpdateAggregateSales" Action="this.UpdateAggregateSales" EntitySet="Sales" />
+      </EntityContainer>
```
ralfhandl, line 32: This won't work, the action import must be in the same service / entity container as the entity set being aggregated.

ralfhandl (changes requested):

examples/Analytics.UpdateAction-sample.xml
HeikoTheissen: Both entity set and action import are now in the entity container `this.container`. Only the entity type is in a different service.

ralfhandl: I'd prefer having the entity type also in this example, especially since the URL https://oasis-tcs.github.io/odata-vocabularies/examples/Org.OData.Aggregation.V1.SalesModel-sample.xml results in 404 Not Found.

ralfhandl (changes requested):

ralfhandl (approved):

examples/Analytics.UpdateAction-sample.xml
```diff
@@ -0,0 +1,58 @@
...
+        <Annotation Term="Core.Description" String="Updates an aggregate entity of type `SalesModel.Sale`" />
+        <Parameter Name="Aggregate" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="String identifying the aggregate entity to be updated" />
+        </Parameter>
+        <Parameter Name="Updates" Type="SalesModel.Sale" Nullable="false">
+          <Annotation Term="Core.Description" String="Aggregate entity with nested properties that are to be updated, like the body of a PATCH request" />
+        </Parameter>
+        <ReturnType Type="SalesModel.Sale" Nullable="false">
+          <Annotation Term="Core.Description" String="Aggregate entity with updates applied" />
+        </ReturnType>
```
HeikoTheissen, line 48: Non-nullable properties that have been aggregated away will be absent in the parameter `Updates` and in the return value. Does this contradict the specification?

PR #214, opened 2022-12-16
Complex-valued ClientOnly annotation

HeikoTheissen:
As discussed with Evan and Bruce

### DIFF
--- a/vocabularies/Offline.xml
+++ b/vocabularies/Offline.xml
@@ -53,7 +53,6 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
       </Term>
       <ComplexType Name="ClientOnlyType">
         <Annotation Term="Common.Experimental" />
-        <Property Name="IsClientOnly" Type="Edm.Bool" Nullable="false" DefaultValue="true"/>
       </ComplexType>
       <ComplexType Name="LocalDraft" BaseType="Offline.ClientOnlyType">
         <Annotation Term="Common.Experimental" />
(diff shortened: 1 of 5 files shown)


ralfhandl (dismissed):
This is cleaner than the two separate terms proposed in #204

vocabularies/Offline.xml
```diff
@@ -49,8 +49,26 @@ Please refer to the [Offline OData product home](https://help.sap.com/doc/f53c64
            attempt to upload any changes that are made to them locally.
            The service as implemented on the server is described by the CSDL document if
            all elements tagged as `Offline.ClientOnly` are removed together with all subsequently dangling references.</String>
-        </Annotation>        
+        </Annotation>
       </Term>
+      <ComplexType Name="ClientOnlyType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="IsClientOnly" Type="Edm.Bool" Nullable="false" DefaultValue="true"/>
```
HeikoTheissen, line 56: Can this property even be omitted? A `ClientOnly` annotation with an empty record as value would then mean "is client only".

ralfhandl: Correct. An empty record without type specification means: all default values of the properties of the term's declared type.

HeikoTheissen: I meant: Can the property be omitted from the complex type definition? That would syntactically rule out `"@Offline.ClientOnly": {"IsClientOnly": false, "ActiveEntitySet": "..."}`.

ralfhandl: Yes, that's also possible, a complex type can have zero properties and only serve as a tagging type and base type.

I'd prefer that.

ralfhandl (approved):

vocabularies/Offline.xml
HeikoTheissen: > Yes, that's also possible, a complex type can have zero properties and only serve as a tagging type and base type.

But that leads to a wrong JSON representation, see oasis-tcs/odata-csdl-schemas#75.

ralfhandl (approved):

HeikoTheissen:
Merged into #204.

PR #219, opened 2023-01-10
Hide properties from UI adaptation

BerSie:
Provide a possibility to exclude properties or entities from UI adaptation/configuration/personalization (at runtime). These properties or entities can only be used at design time (metadata, annotations, or code). The current name is just a base for discussion ... another proposal: UI.ConfigurationHidden.

### DIFF
--- a/vocabularies/UI.json
+++ b/vocabularies/UI.json
@@ -1306,6 +1306,7 @@
     "AdaptationHidden": {
       "$Kind": "Term",
       "$Type": "Core.Tag",
+      "$DefaultValue": true,
       "$AppliesTo": ["Property", "EntitySet", "EntityType"],
       "@Common.Experimental": true,
       "@Core.Description": "Properties or entities annotated with this term can't be used for UI adaptation/configuration/personalization",


vocabularies/UI.xml
```diff
@@ -1448,6 +1448,11 @@ or `204 No Content` if the request included a `Prefer` header with a value of `r
         </Annotation>
       </Term>
 
+      <Term Name="AdaptationHidden" Type="Core.Tag" Nullable="false" AppliesTo="Property EntitySet EntityType">
```
HeikoTheissen, line 1451:
```suggestion
      <Term Name="HiddenAdaptation" Type="Core.Tag" Nullable="false" AppliesTo="Property Parameter EntityType">
```
* Term name in analogy with `HiddenFilter`.
* What shall be the difference between `AppliesTo="EntitySet"` and `AppliesTo="EntityType"`? If we want container-bound annotation targets, we also need `AppliesTo="Collection"`.
* Applies also to parameters.

vocabularies/UI.xml
```diff
@@ -1448,6 +1448,11 @@ or `204 No Content` if the request included a `Prefer` header with a value of `r
...
+        <Annotation Term="Core.Description" String="Properties or entities annotated with this term can't be used for UI adaptation/configuration/personalization" />
```
HeikoTheissen, line 1452:
```suggestion
        <Annotation Term="Core.Description" String="Properties, parameters and entities annotated with this term can't be used for UI adaptation/configuration/personalization" />
```

vocabularies/UI.xml
```diff
@@ -1448,6 +1448,11 @@ or `204 No Content` if the request included a `Prefer` header with a value of `r
...
+        <Annotation Term="Core.LongDescription" String="The annotated properties or entities can only be used in UI based on metadata, annnotations or code" />
```
HeikoTheissen, line 1453:
```suggestion
        <Annotation Term="Core.LongDescription" String="The annotated elements can only be used in UI based on metadata, annnotations or code." />
```

HeikoTheissen (changes requested):

vocabularies/UI.xml
ralfhandl: `HiddenFilter` marks properties that actually used in a "hidden" filter.

The name "HiddenAdaptation" would imply that there actually is a "hidden" adaptation, whereas the term description implies that the annotated thingy is not available for adaptation.

Maybe "NoAdaptation"?

ralfhandl: I'm fine with "AdaptationHidden" because it is in line with "CreateHidden" etc.

HeikoTheissen: OK with `AdaptationHidden`.

vocabularies/UI.xml
BerSie: Why are you mentioning parameters there in the "description"? Not sure if this term makes sense for parameters, and if so, we have to add parameters to the AppliesTo attribute as well.

HeikoTheissen (approved):

ralfhandl (dismissed):

vocabularies/UI.xml
ralfhandl, line 1451:
```suggestion
      <Term Name="AdaptationHidden" Type="Core.Tag" Nullable="false" AppliesTo="Property EntitySet EntityType">
        <Annotation Term="Common.Experimental" />
```

Do we want to start as experimental?

ralfhandl: If we want to use this as a "tagging term" and annotate with `<Annotation Term="UI.AdaptationHidden" />`, then we need a `DefaultValue="true"`.

vocabularies/UI.xml
```diff
@@ -1448,6 +1448,12 @@ or `204 No Content` if the request included a `Prefer` header with a value of `r
         </Annotation>
       </Term>
 
+      <Term Name="AdaptationHidden" Type="Core.Tag" Nullable="false" AppliesTo="Property EntitySet EntityType">
```
ralfhandl, line 1451:
```suggestion
      <Term Name="AdaptationHidden" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property EntitySet EntityType">
```
Do we want to allow tagging syntax without explicit value?
~~~xml
<Annotation Term="UI.AdaptationHidden" />
~~~

BerSie: not sure who was faster ... but I committed exactly this change (as I like the short tagging syntax).

ralfhandl (approved):

PR #221, opened 2023-01-17
Entity type for notes

HeikoTheissen:
To be discussed in the [Albatross meeting](https://wiki.one.int.sap/wiki/display/odata/Workstream+OData+V4#WorkstreamODataV4-2023-01-10): "Reuse Components integration (Notes UI)"


BerSie:
What is the intent of this PR - is it just a copy of something already existing and use it as a basis of a discussion about a new term or a base entity type that has to be used for notes, or is that already something that has to be used and you want to discuss that in this PR?

HeikoTheissen:
> What is the intent of this PR - is it just a copy of something already existing and use it as a basis of a discussion about a new term or a base entity type that has to be used for notes, or is that already something that has to be used and you want to discuss that in this PR?

Services that support a re-use component for notes must contain an entity type with this structure. This vocabulary shall serve as the location where this entity type is published, so that all implementation frameworks can look it up there.

ralfhandl:
> Services that support a re-use component for notes must contain an entity type with this structure

Is this already the minimum structure, or can we strip it down?
- omit the key fields and make the entity type abstract to allow its use outside of the Fiori Draft use case?
- omit S/4-specific fields (NoteBasicObjectNodeType etc.)
- omit "management" fields CreatedBy*, LastChangedBy*
- omit RAP-specific Control fields
- omit Gateway-specific SAP__ fields

That would leave Title, MIMEType, and Content (Length is an implementation detail of platforms still using fixed-length strings).

HeikoTheissen:
> That would leave Title, MIMEType, and Content.

Then I suggest to delete this example and instead annotate the entity type with a new Term `UI.Note` as follows:
```xml
<EntityType Name="C_NoteBasicTPType">
  <Annotation Term="UI.Note">
    <Record>
      <PropertyValue Property="Title" Path="Title"/>
      <PropertyValue Property="MIMEType" Path="MIMEType"/>
      <PropertyValue Property="Content" Path="Content"/>
    </Record>
  </Annotation>
</EntityType>
```

vocabularies/UI.xml
```diff
@@ -1389,6 +1389,21 @@ The trend is
...
       </EnumType>
 
+      <Term Name="Note" Type="UI.NoteType" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Core.Description" String="The annotated entity type can be visualized in a reuse component for notes" />
+      </Term>
+      <ComplexType Name="NoteType">
+        <Property Name="Title" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Title of the note"/>
+        </Property>
+        <Property Name="MIMEType" Type="Edm.String" Nullable="false">
```
BerSie, line 1399:
```suggestion
        <Property Name="ContentType" Type="Edm.String" Nullable="false">
```

vocabularies/UI.xml
```diff
@@ -1389,6 +1389,21 @@ The trend is
...
+          <Annotation Term="Core.Description" String="MIME type of the note"/>
```
BerSie, line 1400:
```suggestion
          <Annotation Term="Core.Description" String="MIME type of the note"/>
          <Annotation Term="Core.IsMediaType" />
```

BerSie: This enforce the usage of a MIME type

vocabularies/UI.xml
BerSie: I would prefer using the same "ContentType" as in the term "MediaResource".

BerSie (commented):
I feel a bit unsure about the ByteSize ... if we should add it to the term or not. Same for CreatedAt, ChangedAt ... and furthermore CreatedByUserName, ChangedByUserName. Let's discuss pro and con ...

ralfhandl (dismissed):

ralfhandl:
> do we need `PropertyPath`s instead of `Path`s, because these properties must also be updatable?

No, `Path` is fine, it is just as invertable as `PropertyPath`.

HeikoTheissen:
> No, `Path` is fine, it is just as invertable as `PropertyPath`.

Yes, but instead of `Path`, someone could use
```
"$If": [{"$Path": "a"}, "b", "c"]
```
which is not invertible. But I also prefer this additional expressiveness over `Edm.PropertyPath`.

thomas-kaltenbach:
@HeikoTheissen thank you for the PullRequest.

I've looked at this and think we need some extra fields.
- `NoteBasicType` - required for grouping the notes by notetype key
- `NoteBasicTypeDescription` - required for displaying a logon language dependent notetype description
- `NoteBasicLanguage` - required for filtering
- `LanguageName` -  required for displaying a logon language dependent language name

Here a screenshot of the UI:
Screenshot 2023-02-01 at 16 22 47

I guess we need also the "management" fields like CreatedBy*, LastChangedBy* but maybe the idea is to use other annotations like `@Semantics.[user|systemDateTime]`, or?

In addition for grouping the notes the `NoteBasicType` points via association to another complex type. 
````
<NavigationProperty Name="_NoteBasicType" Type="com.sap.gateway.srvd.ui_nte_demo_m.v0001.I_NoteBasicTypeType"/>
````
Here the application developer can configure the ui and behaviour backend-driven. This is a DevelopmentObject in ABAP which is defined here [abap-file-formats/blob/main/file-formats/ntty/examples/z_aff_example_ntty.ntty.json](https://github.com/SAP/abap-file-formats/blob/main/file-formats/ntty/examples/z_aff_example_ntty.ntty.json)
So I assume we need an additional EntityType for this.

Followed the fields of current metadata.
````
<EntityType Name="I_NoteBasicTypeType">
  <Key>
      <PropertyRef Name="NoteBasicObjectNodeType"/>
      <PropertyRef Name="NoteBasicType"/>
  </Key>
  <Property Name="NoteBasicObjectNodeType" Type="Edm.String" Nullable="false" MaxLength="30"/>
  <Property Name="NoteBasicType" Type="Edm.String" Nullable="false" MaxLength="20"/>
  <Property Name="NoteBasicLanguageDependent" Type="Edm.Boolean" Nullable="false"/>
  <Property Name="NoteBasicMaximalLength" Type="Edm.Int32" Nullable="false"/>
  <Property Name="NoteBasicMultipleNotes" Type="Edm.Boolean" Nullable="false"/>
  <Property Name="NoteBasicHideTitle" Type="Edm.Boolean" Nullable="false"/>
  <Property Name="NoteBasicIsTextPlain" Type="Edm.Boolean" Nullable="false"/>
</EntityType>
`````

**Comment by HeikoTheissen** [comment]
> I guess we need also the "management" fields like CreatedBy*, LastChangedBy* but maybe the idea is to use other annotations like `@Semantics.[user|systemDateTime]`, or?

Are there corresponding OData annotations for these CDS annotations?

> In addition for grouping the notes the `NoteBasicType` points via association to another complex type.

What's the difference between `C_NoteBasicTPType.NoteBasicType` and `C_NoteBasicTPType._NoteBasicType.NoteBasicType`?

**Comment by thomas-kaltenbach** [comment]
> Are there corresponding OData annotations for these CDS annotations?

Yes there are some annotations i.e. `@Semantics.user.createdBy`, `@Semantics.user.lastchangedBy` ([Link](https://help.sap.com/doc/saphelp_nw75/7.5.5/en-US/fb/cd3a59a94148f6adad80b9c97304ff/content.htm?no_cache=true)) but unfortunately there are not rendered into the metadata. Or I was not able to find it.

> What's the difference between C_NoteBasicTPType.NoteBasicType and C_NoteBasicTPType._NoteBasicType.NoteBasicType?

This are the same fields. `C_NoteBasicTPType.NoteBasicType` this is the foreign key of a NoteType configuration and it associated to `C_NoteBasicTPType._NoteBasicType`. `C_NoteBasicTPType._NoteBasicType.NoteBasicType` is the primary key together with `NoteBasicObjectNodeType`.

File: examples/UI.Note-sample.xml [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+        <Property Name="NoteBasicLanguage" Type="Edm.String" Nullable="false" MaxLength="2" />
+        <Property Name="LanguageName" Type="Edm.String" Nullable="false" MaxLength="16" />
+        <Property Name="NoteBasicTitle" Type="Edm.String" Nullable="false" MaxLength="255" />
+        <Property Name="NoteBasicMIMEType" Type="Edm.String" Nullable="false" MaxLength="127" />
+        <Property Name="NoteBasicLength" Type="Edm.Int32" Nullable="false" />
+        <Property Name="CreatedByUser" Type="Edm.String" Nullable="false" MaxLength="12" />
+        <Property Name="CreationDateTime" Type="Edm.DateTimeOffset" Precision="7" />
+        <Property Name="LastChangedByUser" Type="Edm.String" Nullable="false" MaxLength="12" />
+        <Property Name="LastChangedByUserDescription" Type="Edm.String" Nullable="false" MaxLength="80" />
+        <Property Name="LastChangeDateTime" Type="Edm.DateTimeOffset" Precision="7" />
```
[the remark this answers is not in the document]
thomas-kaltenbach: Its currently not needed for the UI. I might add that for consistency.

HeikoTheissen:
We now have so many properties in the annotation that we must get clarify which of them are mandatory. Because the annotation can only be used if all mandatory properties are available in the underlying entity type. But the annotation should not be 1:1 with an entity type, only the least common denominator of all "Note" entity types (whether from RAP or CAP) must be mandatory.

thomas-kaltenbach:
Ok I understand. 
I had a look at the UI and I guess we need all the fields excepts `IsTextPlain`. This is just for convenience, we could do the same with `ContentType`.
For the `Created*` and `LastChanged*` fields it would be nice we we could use something generic.

vocabularies/UI.xml
```diff
@@ -1004,6 +1004,7 @@ The trend is
               <String>UI.KPI</String>
               <String>UI.LineItem</String>
               <String>UI.MediaResource</String>
+              <String>UI.Note</String>
```
thomas-kaltenbach, line 1007: I guess the purpose of the change is similar to `UI.LineItem` in CDS it's shown `#LINEITEM_REFERENCE`. With the addition of UI.Note, would the following annotation appear in CDS? 
`````
@UI.facet: [
{
      id:            'NotesSection',
      purpose:       #STANDARD,
      type:          #NOTE_REFERENCE,
      label:         'Note Section',
      position:      10,
      targetElement: '_NoteBasic'
  } ]
````
It seems that the singular is almost always used here. In the case of the Notes, it is usually a list. So shouldn't be called #NOTES or NOTELIST?

File: vocabularies/UI.xml [hunk 5]
```diff
@@ -1389,6 +1390,52 @@ The trend is
...
+        <Property Name="TypeDescription" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Description for the type of the note" />
+        </Property>
+        <Property Name="Language" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Language of the note" />
+        </Property>
+        <Property Name="LanguageName" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Name of the language of the note" />
+        </Property>
+        <Property Name="ObjectNodeType" Type="Edm.String" Nullable="false">
```
thomas-kaltenbach, line 1419: Similar to the `Created*` and `LastChanged*` fields, the `ObjectNodeType` refers to the `sapObjectNodeType`. This is not something specific to the Notes component.
In CDS we already have annotations like `@ObjectModel.sapObjectNodeType[Reference]`, maybe we should define something common here too.

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,52 @@ The trend is
...
+        </Property>
+        <Property Name="TypeDescription" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Description for the type of the note" />
+        </Property>
+        <Property Name="Language" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Language of the note" />
+        </Property>
+        <Property Name="LanguageName" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Name of the language of the note" />
+        </Property>
```
thomas-kaltenbach, line 1418: Looking at the common annotations, I found `Common.IsLanguageIdentifier`. `Language` is in our case i.e. 'EN', 'DE'. As I understand it correctly, it could be used here as well, right?
Only something for the LanguageIdentifier Name is needed. This is in our case the translated Language Name i.e. 'Englisch', 'German'.

thomas-kaltenbach:
Small update.
I found out that at the fields `NoteBasicType` and `NoteBasicLangugage` already the followed annotations can be found.
```
<Annotations Target="SAP__self.C_NoteBasicTPType/NoteBasicType">
  <Annotation Term="SAP__common.Text" Path="NoteBasicTypeDescription"/>
  ...
</Annotations>
...
<Annotations Target="SAP__self.C_NoteBasicTPType/NoteBasicLanguage">
  <Annotation Term="SAP__common.Text" Path="LanguageName"/>
  ...
</Annotations>
```
 So this should be sufficient and `NoteBasicTypeDescription` and `LanguageName` can be also removed.

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
...
+          [`Common.CreatedBy`](Common.md#CreatedBy),
+          [`Common.CreatedAt`](Common.md#CreatedAt),
+          [`Common.ChangedBy`](Common.md#ChangedBy),
+          [`Common.ChangedAt`](Common.md#ChangedAt)
+          on the same entity type.</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="NoteType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="Title" Type="Edm.String" Nullable="false">
```
BerSie, line 1407: Title is not nullable, but there is a property HideTitle? Where does the title come from if it's hidden?

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
         </Member>
       </EnumType>
 
+      <Term Name="Note" Type="UI.NoteType" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="The annotated entity type can be visualized in a reuse component for notes" />
```
BerSie, line 1395: I wouldn't write something about reuse components - the term definition should be more timeless - it's defined for the visualization of notes. What do the others think?

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
...
+          <Annotation Term="Core.Description" String="Title of the note" />
+        </Property>
+        <Property Name="Content" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Content of the note, as a string" />
+          <Annotation Term="Core.LongDescription">
+            <String>The reuse component also evaluates the [`Core.MediaType`](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.md#MediaType)
+            annotation on this property.</String>
+          </Annotation>
+        </Property>
+        <Property Name="Type" Type="Edm.String" Nullable="false">
```
BerSie, line 1417: What is the type, can you pls. mention an example?

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
...
+          <Annotation Term="Core.Description" String="Type of the note" />
+        </Property>
+        <Property Name="IsLanguageDependent" Type="Edm.Boolean" Nullable="false">
```
BerSie, line 1420: We try to reuse other terms as good as possible (which I think is good), but here we define a property instead of using Core.IsLanguageDependent. What's the reason?

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
...
+          <Annotation Term="Core.LongDescription">
+            <String>When this is true, the reuse component also evalutes the property of the same entity type
+            that is tagged with [`Common.IsLanguageIdentifier`](Common.md#IsLanguageIdentifier)
+            in order to determine the language of the note.</String>
+          </Annotation>
+        </Property>
+        <Property Name="MaximalLength" Type="Edm.Int32" Nullable="false">
+          <Annotation Term="Core.Description" String="MaximalLength of the text of the note" />
+        </Property>
+        <Property Name="MultipleNotes" Type="Edm.Boolean" Nullable="false">
```
BerSie, line 1431: Could that be an own term? In a collection of notes, you will have this on every instance.

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
...
+          <Annotation Term="Core.Description" String="Whether multiple notes are allowed" />
+        </Property>
+        <Property Name="HideTitle" Type="Edm.Boolean" Nullable="false">
+          <Annotation Term="Core.Description" String="Whether to hide the note title" />
+        </Property>
+        <Property Name="IsTextPlain" Type="Edm.Boolean" Nullable="false">
```
BerSie, line 1437: If Core.MediaType is evaluated, why do we need a IsTextPlain?

BerSie (commented):
Thanks for driving this @thomas-kaltenbach and @HeikoTheissen. For me it's (from a documentation point of view) too tightly coupled with the reuse component.

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
...
+        </Property>
+        <Property Name="IsLanguageDependent" Type="Edm.Boolean" Nullable="false">
+          <Annotation Term="Core.Description" String="Whether the note is language-dependent" />
+          <Annotation Term="Core.LongDescription">
+            <String>When this is true, the reuse component also evalutes the property of the same entity type
+            that is tagged with [`Common.IsLanguageIdentifier`](Common.md#IsLanguageIdentifier)
+            in order to determine the language of the note.</String>
+          </Annotation>
+        </Property>
+        <Property Name="MaximalLength" Type="Edm.Int32" Nullable="false">
```
ralfhandl, line 1428: Isn't this the `MaxLength` of the underlying property `Content` points to via a path expression?

ralfhandl (changes requested):

vocabularies/UI.xml
HeikoTheissen: Then what purpose does `_NoteBasic/NoteBasicMaximalLength` serve?

vocabularies/UI.xml
HeikoTheissen: Which target should be annotated with the term?

BerSie: The annotation UI.Note could be tagged with it?

vocabularies/UI.xml
ralfhandl: We should ask the author of the CDS view. To me it seems totally redundant, given that `NoteBasicContent` is of type `string` and thus has a "built-in" length.

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,47 @@ The trend is
...
+            another property of the same entity type that is tagged with [`Common.IsLanguageIdentifier`](Common.md#IsLanguageIdentifier)
+            determines the language of the note.</String>
+          </Annotation>
+        </Property>
+        <Property Name="Type" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Type of the note" />
+        </Property>
+        <Property Name="MaximalLength" Type="Edm.Int32" Nullable="true">
+          <Annotation Term="Core.Description" String="Maximal length of the text of the note" />
+        </Property>
```
ralfhandl, line 1425: This seems redundant, remove

ralfhandl (changes requested):

vocabularies/UI.xml
HeikoTheissen: But it depends on the instance whether there are multiple notes for the same business entity.

BerSie: It's not about "having multiple notes", but about  if "multiple notes are allowed" - and here I believe (!) that is either allowed or not for one business entity type.

vocabularies/UI.xml
thomas-kaltenbach: > Then what purpose does `_NoteBasic/NoteBasicMaximalLength` serve, @thomas-kaltenbach?

From our point of view we have two entities. Note & note type.
**Note** is the content itself, with title, language and content.
**NoteType** is the configuration of a group of notes. Like "Product Description" or "Comments".
Applications can define their own note types and assign them to a SAPObjectNodeType.

The following properties...
    <Property Name="NoteBasicLanguageDependent" Type="Edm.Boolean" Nullable="false"/>
    <Property Name="NoteBasicMaximalLength" Type="Edm.Int32" Nullable="false"/>
    <Property Name="NoteBasicMultipleNotes" Type="Edm.Boolean" Nullable="false"/>
    <Property Name="NoteBasicHideTitle" Type="Edm.Boolean" Nullable="false"/>
... are all part of our NoteType. This is needed so that in the frontend is known what is allowed or not.
So an application can decide to allow only 200 characters.

The two things are now in the UI.Note entity. The question comes up regularly. @HeikoTheissen should maybe separate it and add an additional type for it, maybe `UI.NoteType`? This would make it much clearer.

HeikoTheissen: The separation into note and note type is an implementation detail, it should not be reflected in the UI annotation. But since we already have the four properties in the annotation, we don't need the `Type` itself, thus answering https://github.com/SAP/odata-vocabularies/pull/221#discussion_r1114415917.

vocabularies/UI.xml
HeikoTheissen: The same is true for `MaximalLength`, `LanguageDependent` and `HideTitle`. These four are actually properties of the note _type_, but I don't want to expose the distinction between note and note type in this annotation.

vocabularies/UI.xml
thomas-kaltenbach: We need the type because we are grouping by this value. So we need to know by which field we should do this. In addition we need the "Text" of the NoteType this is description which is displayed in the UI. 
https://github.com/SAP/odata-vocabularies/pull/221#issuecomment-1412315655

vocabularies/UI.xml
thomas-kaltenbach: I described it here with and screenshot of the UI https://github.com/SAP/odata-vocabularies/pull/221#issuecomment-1412315655 and also here https://github.com/SAP/odata-vocabularies/pull/221#discussion_r1115311905

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,49 @@ The trend is
...
+            another property of the same entity type that is tagged with [`Common.IsLanguageIdentifier`](Common.md#IsLanguageIdentifier)
+            determines the language of the note.</String>
+          </Annotation>
+        </Property>
+        <Property Name="Type" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="A type used for grouping notes" />
+        </Property>
+        <Property Name="MaximalLength" Type="Edm.Int32" Nullable="true">
+          <Annotation Term="Core.Description" String="Type-specific maximal length of the content of the note" />
+        </Property>
```
ralfhandl, line 1427: What is the client supposed to do with this value?

Prevent users from entering overly long note texts? Are we targeting edit scenarios here?

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,49 @@ The trend is
...
+        <Property Name="MultipleNotes" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="Whether the type allows multiple notes for one object" />
+        </Property>
```
ralfhandl, line 1430: Again: is this for edit scenarios?

vocabularies/UI.xml
HeikoTheissen: @thomas-kaltenbach, can you answer this?

thomas-kaltenbach: Yes, exactly. This is for editing.

vocabularies/UI.xml
thomas-kaltenbach: Yes, exactly. This is for editing.

ralfhandl (commented):
No more questions

martin4all:
<Annotations Target="myservice.C_NoteBasicTPType">         
        <Annotation Term="Common.SAPObjectNodeType">
          <Record>
            <PropertyValue Property="Name" Path="_NoteBasic/NoteBasicObjectNodeType" />
          </Record>
        </Annotation>

Annotating the SAP object type like that will not work here, because here we mention the SAP object type of the "Note" business object, which is fixed:

      <Annotations Target="myservice.C_NoteBasicTPType">         
        <Annotation Term="Common.SAPObjectNodeType">
          <Record>
            <PropertyValue Property="Name" String="NoteBasic" />
          </Record>
        </Annotation>

But what we want here is the SAP object type of the _note's host object._ Therefore we should add an additional property to the UI annotation?

HeikoTheissen:
> Annotating the SAP object type like that will not work here, because here we mention the SAP object type of the "Note" business object.

I have corrected the example, the `Common.SAPObjectNodeType` annotation now points to `_NoteBasicType/NoteBasicObjectNodeType`, which is node type defined in the note type, and this is the node type of the host object.

vocabularies/UI.xml
BerSie: Thanks, @thomas-kaltenbach . Where does this type come from - the screenshot shows a dropdown. So there must a "ValueList EntitySet" exist that contains these value. And is this something the app defines, or the customer, or is it overarching (a central team defines it)?

vocabularies/UI.xml
BerSie: I'm still not convinced :-)
MaxLength is an attribute of a property - so this should come within the $metadata (http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_Toc453752530).
LanguageDependent is a Core Term (https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.xml#L306) which can be applied to a property or a term.
HideTitle can be maybe achieved by using UI.Hidden on the title property.
Wrt MultipleNotes I didn't immediately found something, that's why I proposed an additional term, but I checked again ... and it sounds maybe weird, but we have a UI.PresentationVariant with a property "MaxItems" - note: we don't have it as a property on UI.LineItem!, in addition I find there something like GroupBy (or SortBy) which could be applied to the Notes as well, right. Would the Notes term fit there as well? Just a thought...

vocabularies/UI.xml
thomas-kaltenbach: The application defines a notetype and assign it to applications SAPObjectNodeType. For ABAP Cloud we are also developing ADT UI for these entities (NTTY/NTTA). 
So we know which notetypes can be expected and also provide the create button for these notetypes.
Therefore we have the entityset currently called `I_NoteBasicType`.

vocabularies/UI.xml
thomas-kaltenbach: > MaxLength is an attribute of a property - so this should come within the $metadata (http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_Toc453752530).
The problem is we don't have a single "MaxLength". Notes is a generic reuse component where an application developer can define the note type and set the maximum length of one note type to 50 characters and another note type can be unlimited. So it's not possible to set this in the metadata unless we add all the note type definitions in the metadata and don't have an entityset for it anymore. 

The same applies to the other fields.

HeikoTheissen: > MaxLength is an attribute of a property

This does not work, because we have only _one_ property `C_NoteBasicTPType/NoteBasicContent` that will be used by _all_ note types. But the `NoteBasicMaximalLength` varies per note type.

We could express this through an annotation `{"@Validation.Pattern": ".{,128}"}`, which would have to be constructed dynamically:
```
<Annotations Target="service.C_NoteBasicTPType/NoteBasicContent">
  <Annotation Term="Validation.Pattern">
    <Apply Function="odata.concat">
      <String>.{,</String>
      <Path>_NoteBasicType/NoteBasicMaximalLength</Path>
      <String>}</String>
    </Apply>
  </Annotation>
</Annotations>
```
Do we want that?

> HideTitle can be maybe achieved by using UI.Hidden on the title property

Good idea, see https://github.com/SAP/odata-vocabularies/pull/221/commits/147ccccecfeb2afddb342e5ce89c6647074076de.

BerSie: Oh, OK - the MaxLength depends on the NoteType - now I understand the comment from @HeikoTheissen much better - the term _type_ is used in too many contexts :-).

examples/UI.Note-sample.xml
```diff
@@ -0,0 +1,58 @@
...
+        <Annotation Term="Common.CreatedBy" Path="CreatedByUser" />
+        <Annotation Term="Common.CreatedAt" Path="CreationDateTime" />
+        <Annotation Term="Common.ChangedBy" Path="LastChangedByUser" />
+        <Annotation Term="Common.ChangedAt" Path="LastChangeDateTime" />
+      </Annotations>
+      <Annotations Target="service.C_NoteBasicTPType/NoteBasicTitle">
+        <Annotation Term="UI.Hidden" Path="_NoteBasicType/NoteBasicHideTitle" />
+      </Annotations>
+      <Annotations Target="service.C_NoteBasicTPType/NoteBasicContent">
+        <Annotation Term="Core.MediaType" Path="NoteBasicMIMEType" />
```
HeikoTheissen, line 50:
```suggestion
        <Annotation Term="Core.MediaType" Path="NoteBasicMIMEType" />
        <Annotation Term="Validation.Pattern">
          <Apply Function="odata.concat">
            <String>.{,</String>
            <Path>_NoteBasicType/NoteBasicMaximalLength</Path>
            <String>}</String>
          </Apply>
        </Annotation>
```
This would allow us to omit the `@UI.Note/MaximalLength` property.

examples/UI.Note-sample.xml
```diff
@@ -0,0 +1,58 @@
...
+          <Record>
+            <PropertyValue Property="Name" Path="_NoteBasicType/NoteBasicObjectNodeType" />
+          </Record>
+        </Annotation>
+        <Annotation Term="Common.CreatedBy" Path="CreatedByUser" />
+        <Annotation Term="Common.CreatedAt" Path="CreationDateTime" />
+        <Annotation Term="Common.ChangedBy" Path="LastChangedByUser" />
+        <Annotation Term="Common.ChangedAt" Path="LastChangeDateTime" />
+      </Annotations>
+      <Annotations Target="service.C_NoteBasicTPType/NoteBasicTitle">
```
HeikoTheissen, line 46:
```suggestion
      <Annotations Target="service.C_NTE_DEMO_MType/_NoteBasic">
        <Annotation Term="Validation.MaxItems">
          <If>
            <Path>_NoteBasic/_NoteBasicType/NoteBasicMultipleNotes</Path>
            <Int>9223372036854775807</Int>
            <Int>1</Int>
          </If>
        </Annotation>
      </Annotations>
      <Annotations Target="service.C_NoteBasicTPType/NoteBasicTitle">
```
This would allow us to omit `@UI.Note/MultipleNotes`.

vocabularies/UI.xml
HeikoTheissen: Regarding `MultipleNotes`, see https://github.com/SAP/odata-vocabularies/pull/221/files#r1116658577.

examples/UI.Note-sample.xml
thomas-kaltenbach: could you give an example how this string would look like?
Would it be something like this: `{NOTETYPE1,200},{NOTETYPE2,50}`? 
If yes I guess, it would be ok for us.

martin4all:
> > Annotating the SAP object type like that will not work here, because here we mention the SAP object type of the "Note" business object.
> 
> I have corrected the example, the `Common.SAPObjectNodeType` annotation now points to `_NoteBasicType/NoteBasicObjectNodeType`, which is the node type defined in the note type, and this is the node type of the host object.

No, this is still wrong. This annotation describes SAP Object Node Type represented by the annotated entity type. And this would be "Notes" or "NotesBasic" (depending on the name of the SAP Object Note Type @thomas-kaltenbach will assign to the Notes business Object) and this is **not** the host objects. Maybe **SAPObjectNodeTypeReference** would match better:

     <Annotations Target="myservice.C_NoteBasicTPType/NoteBasicContent">
       <Annotation Term="Common.SAPObjectNodeTypeReference" Path="_NoteBasic/NoteBasicObjectNodeType" />
       <Annotation Term="Core.MediaType" Path="NoteBasicMIMEType" />
       <Annotation Term="Core.IsLanguageDependent" Path="_NoteBasic/NoteBasicLanguageDependent" />
     </Annotations>

HeikoTheissen:
> No, this is still wrong.

The annotation
https://github.com/SAP/odata-vocabularies/blob/147ccccecfeb2afddb342e5ce89c6647074076de/examples/UI.Note-sample.xml#L36-L40
on `C_NoteBasicTPType` works, because
```
/sap/opu/odata4/sap/ui_nte_demo_m_o4/srvd/sap/ui_nte_demo_m/0001
/C_NTE_DEMO_M(ObjectUUID=42010aef-3ca5-1edd-a4b1-0918e962f885,IsActiveEntity=true)
/_NoteBasic
?$expand=_NoteBasicType($select=NoteBasicObjectNodeType)
```
returns 
```
"NoteBasicObjectNodeType" : "DemoObject"
```

examples/UI.Note-sample.xml
HeikoTheissen: `{"@Validation.Pattern": ".{,128}"}`

thomas-kaltenbach: ok, now I got it. This validation pattern is assigned to the NoteContent.
This means for the content I have to follow the path `_NoteBasicType/NoteBasicMaximalLength` and I get the maxLength, right?
The question is only how can I define this in CDS?

HeikoTheissen: > This means for the content I have to follow the path `_NoteBasicType/NoteBasicMaximalLength` and I get the maxLength, right?

Well, you would have to ensure that the `<Apply>` element otherwise looks like in this example. I find this tooo tricky.

martin4all (approved):

thomas-kaltenbach (approved):

ralfhandl (changes requested):

examples/UI.Note-sample.xml
```diff
@@ -0,0 +1,53 @@
...
+            </Record>
+          </Collection>
+        </Annotation>
+      </Annotations>
+      <Annotations Target="service.C_NoteBasicTPType">
+        <Annotation Term="UI.Note">
+          <Record>
+            <PropertyValue Property="Title" Path="NoteBasicTitle">
+              <Annotation Term="UI.Hidden" Path="_NoteBasicType/NoteBasicHideTitle" />
+            </PropertyValue>
```
ralfhandl, line 31: Does UI5 already support this? So far we seem to only have
- annotations directly within annotations (ApplicableTerms annotated on a Term)
- annotations within annotation records (ApplicableTerms annotated on a ComplexType)

This seems to be the first time we have annotations within an annotation property value.

vocabularies/UI.xml
ralfhandl: @thomas-kaltenbach Is this comment still relevant? If not, please resolve this conversation.

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,55 @@ The trend is
...
+          [`Common.ChangedBy`](Common.md#ChangedBy),
+          [`Common.ChangedAt`](Common.md#ChangedAt)
+          on the same entity type.</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="NoteType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="Title" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.Description" String="Title of the note" />
+          <Annotation Term="Validation.ApplicableTerms">
```
ralfhandl, line 1409: This appears to be our first use of `Validation.ApplicableTerms` on a property, other uses are on complex types or on terms.

Is this already supported by UI5?

HeikoTheissen: What must UI5 support here? Are you referring to the annotation modeling tools?

vocabularies/UI.xml
thomas-kaltenbach: It's not relevant anymore

vocabularies/UI.xml
ralfhandl: UI5 OData Model to consume it at runtime.

And the annotation modeling tools at design-time, thanks for pointing them out!

examples/UI.Note-sample.xml
```diff
@@ -0,0 +1,54 @@
...
+            <PropertyValue Property="MaximalLength" Path="_NoteBasicType/NoteBasicMaximalLength" />
+            <PropertyValue Property="MultipleNotes" Path="_NoteBasicType/NoteBasicMultipleNotes" />
+          </Record>
+        </Annotation>
+        <Annotation Term="Common.CreatedBy" Path="CreatedByUser" />
+        <Annotation Term="Common.CreatedAt" Path="CreationDateTime" />
+        <Annotation Term="Common.ChangedBy" Path="LastChangedByUser" />
+        <Annotation Term="Common.ChangedAt" Path="LastChangeDateTime" />
+      </Annotations>
+      <Annotations Target="service.C_NoteBasicTPType/NoteBasicTitle">
```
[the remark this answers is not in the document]
ralfhandl: We had problems in other cases with "cascading hidden", and I advise against running into the same problem again here.

vocabularies/UI.xml
```diff
@@ -1406,6 +1406,10 @@ The trend is
         <Annotation Term="Common.Experimental" />
         <Property Name="Title" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Title of the note" />
+          <Annotation Term="Core.LongDescription">
+            <String>The title of a note is hidden with an annotation `@UI.Note/Title/@UI.Hidden`, not with
+            an annotation of the property targeted by `@UI.Note/Title`.</String>
```
ralfhandl, line 1411:
```suggestion
            an annotation on the property targeted by `@UI.Note/Title`.</String>
```
Not sure whether "of" or "on" is better here.

ralfhandl (dismissed):

examples/UI.Note-sample.xml
thomas-kaltenbach: > @thomas-kaltenbach, can UI5 consume this annotation?

I cannot say it for sure but I guess it should be no problem to read the annotation with the MetadataModel.

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,59 @@ The trend is
...
+          </Annotation>
+        </Property>
+        <Property Name="Type" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="A type used for grouping notes" />
+        </Property>
+        <Property Name="MaximalLength" Type="Edm.Int32" Nullable="true">
+          <Annotation Term="Core.Description" String="Type-specific maximal length of the content of the note" />
+        </Property>
+        <Property Name="MultipleNotes" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="Whether the type allows multiple notes for one object" />
```
BerSie, line 1442: This description left me with the question: which type - entity type, or the Type property in UI.Note?

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,59 @@ The trend is
...
+            <Collection>
+              <String>UI.Hidden</String>
+            </Collection>
+          </Annotation>
+        </Property>
+        <Property Name="Content" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Content of the note, as a string" />
+          <Annotation Term="Core.LongDescription">
+            <String>The property targeted by `@UI.Note/Content` must be annotated with `Core.MediaType` and
+            `Common.SAPObjectNodeTypeReference`. When it is tagged with `Core.IsLanguageDependent`,
```
BerSie, line 1423: I feel that this requirement is to hard - why can I use notes only in combination with a "SAPObjectNodeType" (btw. it would be good to link a reference here, what this term means and where I can find the list of these types). When I checked the Common vocabulary I recognized that the SAPObjectNode is a complex type with only one property - by intent? (but that's off-topic)

vocabularies/UI.xml
```diff
@@ -1389,6 +1390,59 @@ The trend is
...
+          <Annotation Term="Validation.ApplicableTerms">
+            <Collection>
+              <String>Core.MediaType</String>
+              <String>Core.IsLanguageDependent</String>
+              <String>Common.SAPObjectNodeTypeReference</String>
+            </Collection>
+          </Annotation>
+        </Property>
+        <Property Name="Type" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="A type used for grouping notes" />
```
BerSie, line 1436: I still cannot grap this - who defines these types (assuming a CAP application wants to use notes)? And why is it nullable: false - why should I always want to group my notes.

BerSie (commented):
We are coming closer imo - but there are (at least for me) too many assumptions about the backend. We should also enable with this annotation simple use cases, where the end user can simply put one or more notes to the current object instance (and this also on CAP).

vocabularies/UI.xml
HeikoTheissen: `Common.SAPObjectNodeTypeReference` _is_ linked under "Can be annotated with", see the markdown file.

@thomas-kaltenbach, please answer whether a SAP object node type reference is mandatory when working with notes.

vocabularies/UI.xml
HeikoTheissen: @thomas-kaltenbach, please answer.

vocabularies/UI.xml
BerSie: Sounds good - and as I claimed that (at least imo) the note type is not mandatory - can I expect that this property is per entity type instance?

vocabularies/UI.xml
BerSie: Maybe a misunderstanding - the link to the other term is good - but step into the shoes of a CAP app developer, who wants to implement a notes reuse service - what do you expect form him (and the description there "The name of the SAP Object Node Type" will not be very helpful for him).

HeikoTheissen: I have changed the description of `SAPObjectNodeTypeReference`. But the list of these types is only available SAP-internally, so it makes no sense to document it in a public GitHub repo.

vocabularies/UI.xml
thomas-kaltenbach: > I still cannot grap this - who defines these types (assuming a CAP application wants to use notes)? And why is it nullable: false - why should I always want to group my notes.

In the followed comment you can see a screenshot of the notes component https://github.com/SAP/odata-vocabularies/pull/221#issuecomment-1412315655.
A Note Type is defines a group of notes for example "Product Description" or "Comments". The application can reuse an existing one or define a new one. So far, applications and UX have required us to display them by default in groups. Which also makes sense from my point of view.
In addition you can define at the note type the content-type, if a title should be displayed and so on.

vocabularies/UI.xml
thomas-kaltenbach: > @thomas-kaltenbach, please answer whether a SAP object node type reference is mandatory when working with notes.

For an generic note reuse service it is needed to know which object node type is using the component. I guess this is not needed if a application is defining their own notes and reusing the frontend. But for a generic backend service we do some additional http calls with the application objectID and the sapObjectNodeType for the language/notetype combobox and for the create menu button to display them correct.

BerSie: Thanks, @thomas-kaltenbach! For me this means. that the annotation with "Common.SAPObjectNodeTypeReference" should not be a MUST.

vocabularies/UI.xml
BerSie: Thanks, @thomas-kaltenbach! The property "Type defines a group, but if an app only has a plain list of notes, this property doesn't really matter, right. If so, we shouldn't define it as Nullable=false.

thomas-kaltenbach: The "type" also defines how the group should be displayed. We don't have a default behavior how to handle the UI when no configuration is available. From our side, the note type and the configuration attached to it are required.
Also, the note type is persisted as a foreignkey at the note content in the database because we are a reuse component and need it for the where/on condition. If we say that this could be nullable, we would be making the behaviour very complex. To be honest, I don't even know if that fits into the whole concept.
if there are any further doubts, we can also discuss it in a meeting.

vocabularies/UI.xml
thomas-kaltenbach: For me it is currently difficult to say exactly because we still have to carry out the implementation. But it may well not be a MUST.
As soon as I know more details, I can also open a PR with an adjustment if that is desired.

BerSie (approved):
looks good to me

PR #224, opened 2023-01-25
Add Header and Footer

bernhardgr42:
Adding Header and Footer

### DIFF
--- a/vocabularies/PDF.xml
+++ b/vocabularies/PDF.xml
@@ -81,6 +81,7 @@
           </Annotation>
         </Property>
         <Property Name="HeaderFooter" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Common.Experimental" />
           <Annotation Term="Core.Description" String="Page header and footer supported" />
           <Annotation Term="Core.LongDescription">
             <String>Headers and footers are areas in the top and the bottom page margins, where you can add page number and date information</String>
--- a/vocabularies/PDF.xml
+++ b/vocabularies/PDF.xml
@@ -81,7 +81,7 @@
           </Annotation>
         </Property>
         <Property Name="HeaderFooter" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
-          <Annotation Term="Core.Description" String="Header and Footer of document" />
+          <Annotation Term="Core.Description" String="Whether the document contains header and footer" />
           <Annotation Term="Core.LongDescription">
             <String>Headers and footers are areas in the top and the bottom page margins, where you can add page number and date information</String>
           </Annotation>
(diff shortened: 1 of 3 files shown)


vocabularies/PDF.xml
```diff
@@ -80,6 +80,12 @@
             <String>Is padding (left, right, bottom, top) supported?</String>
           </Annotation>
         </Property>
+        <Property Name="HeaderFooter" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="Header and Footer of document" />
```
HeikoTheissen, line 84:
```suggestion
          <Annotation Term="Core.Description" String="Whether the document contains Header and Footer" />
```
This boolean property only controls whether header and footer exist, not how they look, right?

HeikoTheissen (changes requested):

vocabularies/PDF.xml
bernhardgr42: Hi Heiko,

this annotation tells the client if the backend is capable to add a header or a footer. A requirement is that the "PDF API" is able to do this and this is only supported with a certain Backend version.

To control the page number, data/time stuff in a header or footer the client has to do a Create request on 
<EntityType Name="DocumentDescription"> of our common service in OData.

HeikoTheissen: Will you then rephrase the description as I suggested?

HeikoTheissen (approved):

HeikoTheissen (dismissed):

vocabularies/PDF.xml
```diff
@@ -80,6 +80,12 @@
             <String>Is padding (left, right, bottom, top) supported?</String>
           </Annotation>
         </Property>
+        <Property Name="HeaderFooter" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="A header and Footer of the document is supported" />
```
ralfhandl, line 84:
```suggestion
          <Annotation Term="Core.Description" String="Page header and footer supported" />
```

ralfhandl (commented):
Rephrase description

ralfhandl (dismissed):

HeikoTheissen (approved):

vocabularies/PDF.xml
```diff
@@ -80,6 +80,12 @@
             <String>Is padding (left, right, bottom, top) supported?</String>
           </Annotation>
         </Property>
+        <Property Name="HeaderFooter" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
```
ralfhandl, line 83:
```suggestion
        <Property Name="HeaderFooter" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
          <Annotation Term="Common.Experimental" />
```

HeikoTheissen (approved):

PR #230, opened 2023-02-01
New type UI.ActionName

ralfhandl:


BerSie (approved):
Looks good to me, seems to hit all cases

PR #233, opened 2023-02-24
Proposal for a Common.ValidationFunction on parameters and action

nlunets:
You might want to invest in a file formatter since it seems that the line ending are not consistent :(


ralfhandl:
We are recommending a VS Code plugin for XML that has nice pretty-printing: https://github.com/SAP/odata-vocabularies/blob/aceec18ca8d302a90caa432a7ea233c781b9270a/.vscode/extensions.json#L9

nlunets:
I'm not usually using vscode though so that's my problem.
Fixed it here

vocabularies/Common.xml
```diff
@@ -1425,6 +1425,14 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <Term Name="ValidationFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="Parameter Action Function">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Function that validates whether the parameter value, or parameter values (in case of Action and Function) is correct or not" />
```
HeikoTheissen, line 1430: How would a generic client know how this to pass the parameter value(s) to this function? Must it have a particular signature?

nlunets: In case of a parameter the function would require you to pass as input parameter the current parameter value so the type would have to match.
If the function is defined globally for an action or function i guess we would need to define a ComplexType mapping to the original function parameters, or at least a subset of it if some parameter need to be excluded from validation (some that would be validated through value helps for instance)

ralfhandl:
> I'm not usually using vscode though so that's my problem.

We could try out https://www.npmjs.com/package/@prettier/plugin-xml and force-format XML files during `npm run build`. Need to try out whether this conflicts with the Red Hat XML plugin.

The Red Hat XML plugin provides XML schema validation and schema-based auto-complete, so maybe you want to switch to VS Code when contributing to this repository?

vocabularies/Common.xml
HeikoTheissen: Such rules should be part of the long description of the new term.

nlunets:
Having prettier running could also help for sure but that's really up to you !
Otherwise yes i'm now using vscode for this one :)

vocabularies/Common.xml
nlunets: yes that was my assumption as well, is the way i formulated them ok for this or is there something i should keep in mind ?

vocabularies/Common.xml
```diff
@@ -1425,6 +1425,14 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+        <Annotation Term="Core.LongDescription">
+          <String>The qualified name of an unbound function that validates whether the parameter value, or parameter values (in case of Action and Function) is correct or not.</String>
```
HeikoTheissen, line 1432:
```suggestion
          <String>This annotation must not target a binding parameter. When it targets a non-binding parameter, the validation function has one parameter to which the to-be-validated parameter value must be passed. When this annotation targets an action or function, the validation function has one parameter per non-binding action/function parameter to which the to-be-validated value must be passed. The validation function returns true or false.</String>
```

vocabularies/Common.xml
HeikoTheissen: See my [suggestion](https://github.com/SAP/odata-vocabularies/pull/233#discussion_r1117228872).

nlunets: thanks, at least i understand it so that might help :)

vocabularies/Common.json
```diff
@@ -1282,6 +1282,14 @@
...
       "@Core.IsURL": true,
       "@Core.Description": "Base URL for WebSocket connections"
+    },
+    "ValidationFunction": {
+      "$Kind": "Term",
+      "$Type": "Common.QualifiedName",
+      "$AppliesTo": ["Parameter", "Action", "Function"],
+      "@Common.Experimental": true,
+      "@Core.Description": "Function that validates whether the parameter value, or parameter values (in case of Action and Function) is correct or not",
+      "@Core.LongDescription": "This annotation must not target a binding parameter. When it targets a non-binding parameter, the validation function has one parameter to which the to-be-validated parameter value must be passed. When this annotation targets an action or function, the validation function has one parameter per non-binding action/function parameter to which the to-be-validated value must be passed. The validation function returns true or false."
```
uhlmannm, line 1292: Would it not also require the binding parameter to have the right context? E.g. if the bound action changes the status of an object, the allowed new statuses might depend on the current status of the object.
From another angle: What about overloading? Or would overloading not be allowed because the ValidationFunction is unbound?

vocabularies/Common.json
```diff
@@ -1282,6 +1282,14 @@
       "@Common.Experimental": true,
       "@Core.IsURL": true,
       "@Core.Description": "Base URL for WebSocket connections"
+    },
+    "ValidationFunction": {
+      "$Kind": "Term",
+      "$Type": "Common.QualifiedName",
+      "$AppliesTo": ["Parameter", "Action", "Function"],
```
uhlmannm, line 1289: Does it make sense to introduce a ValidationFunction for a Function? After all, the function can simply be executed and it would return whether parameter values are good (together with the respective error messages).

vocabularies/Common.xml
```diff
@@ -1425,6 +1425,14 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <Term Name="ValidationFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="Parameter Action Function">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Function that validates whether the parameter value, or parameter values (in case of Action and Function) is correct or not" />
+        <Annotation Term="Core.LongDescription">
+          <String>This annotation must not target a binding parameter. When it targets a non-binding parameter, the validation function has one parameter to which the to-be-validated parameter value must be passed. When this annotation targets an action or function, the validation function has one parameter per non-binding action/function parameter to which the to-be-validated value must be passed. The validation function returns true or false.</String>
```
HeikoTheissen, line 1432:
```suggestion
          <String>When this annotation targets a non-binding parameter, the validation function must have an unbound overload with one parameter to which the to-be-validated parameter value must be passed. When this annotation targets an action or function, the validation function must have an overload with the same parameter types (including binding parameter, if applicable) as the annotation target. The validation function returns true or false.</String>
```

vocabularies/Common.json
HeikoTheissen: This annotation can be used to validate individual function parameters even before all parameters have been entered.

vocabularies/Common.xml
nlunets: You lost me as soon as you introduced the word overload here.
Why would overload matter in this case ? We're talking about a function such as

```myValidationFunction(parameter:any): 200OK | Error with message which probably need to be defined / aligned`

File: vocabularies/Common.json (hunk 4 above)
Comment (nlunets) [reply vocabularies/Common.json:1289]: Yes that's the whole point :)

File: vocabularies/Common.xml (hunk 5 above)
Comment (HeikoTheissen) [reply vocabularies/Common.xml:1432]: An action or function can have several overloads with different parameters. And you could annotate individual overloads with different validation functions, or with different overloads of the same validation function.

> `myValidationFunction(parameter:any): 200OK | Error with message which probably need to be defined / aligned`

This is a validation function for _one_ parameter. But as @uhlmannm pointed out, the result of the validation can also depend on other parameters, including the binding parameter.

Comment (nlunets) [reply vocabularies/Common.xml:1432]: Ok in which case if you're targeting a bound action/ function you also need the validation function to exist in relation to the existing binding parameter, make sense

File: vocabularies/Common.xml [hunk 6]
```diff
@@ -1425,6 +1425,14 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <Term Name="ValidationFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="Parameter Action Function">
```
HeikoTheissen, line 1428: `Common.QualifiedName` identifies a function. Do we also want to be able to identify a particular function overload?

vocabularies/Common.xml
HeikoTheissen: A validation function needs a binding parameter, if it shall validate more than one parameter of a bound action or function. For validating only _one_ parameter, an unbound validation function is needed.

uhlmannm: I see only a limited use in a validation without the binding parameter. A validation from business perspective would often require to know the context.
Also for the UI design I would not see "bound validation functions" as a big restriction. Normally, the user selects the object / binding parameter, then presses the action button and finally fills all the other parameters in a popup. So the binding parameter is available when calling a validation function.

HeikoTheissen: > I see only a limited use in a validation without the binding parameter.

I have updated my suggestion accordingly.

vocabularies/Common.xml
ThomasChadzelek: I don't get how you call this function at all. You need the qualified name for a bound function, but what would be its binding parameter? I assume an unbound function is sufficient, but then you would need the name of the function import to call it. Once this is clarified, we can look at the rules for overloading (https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_FunctionOverloads)

Also: for a single non-binding parameter, the overload should be clear from its type. For an operation, I wonder if the validation function uses the same sequence of parameters or if there is a correlation based on names? Depending on that, there may again be no freedom w.r.t. overloads.

PR #234, opened 2023-03-02
Instance annotation for 24:00:00

HeikoTheissen:


vocabularies/Common.xml
```diff
@@ -131,6 +131,12 @@
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="Annotated property or parameter is a time zone" />
       </Term>
+      <Term Name="Time24h" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property Parameter">
```
ralfhandl, line 134: How about
```suggestion
      <Term Name="IsEndOfDay" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property Parameter">
```

PR #242, opened 2023-03-21
Fix: allow some UI annotations on type definitions

ralfhandl:
Allow some UI annotations also on type definitions - CAP CDS supports type definitions

Intentionally do not add `TypeDefinition` for "hidden" or property-/parameter-specific annotations.


HeikoTheissen (commented):
What about
- [ ] `Hidden`
- [ ] `HiddenFilter`
- [ ] `AdaptationHidden`
- [ ] `DataFieldDefault`
- [ ] `ParameterDefaultValue`
- [ ] `RecommendationState`
- [ ] `ExcludeFromNavigationContext`?

ralfhandl:
Those are the ones meant by

> Intentionally do not add `TypeDefinition` for "hidden" or property-/parameter-specific annotations.

Thanks for explicitly listing them

ralfhandl:
RecommendationState does not have an `AppliesTo` list.

@BerSie: can you check where it is expected/considered?

BerSie:
> @BerSie: can you check where it is expected/considered?

I don't know - I have to ask team members who is using it and how... if it's used.

HeikoTheissen:
The case of `RecommendationState` is still open.

HeikoTheissen (dismissed):

vocabularies/UI.xml
```diff
@@ -1400,12 +1400,12 @@ The trend is
...
       </Term>
 
-      <Term Name="MultiLineText" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property PropertyValue Parameter">
+      <Term Name="MultiLineText" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property PropertyValue Parameter TypeDefinition">
         <Annotation Term="Core.Description" String="Properties and parameters annotated with this annotation should be rendered as multi-line text (e.g. text area)" />
         <Annotation Term="Core.RequiresType" String="Edm.String" />
       </Term>
 
-      <Term Name="Placeholder" Type="Edm.String" Nullable="false" AppliesTo="Property Parameter">
+      <Term Name="Placeholder" Type="Edm.String" Nullable="false" AppliesTo="Property Parameter TypeDefinition">
```
BerSie, line 1408: Isn't that a property-/parameter-specific annotation? MultiLineText above looks similar.

BerSie (commented):
DoNotCheckScaleOfMeasuredQuantity and RecommendationList look also like property-and parameter-specific annotations. Your initial remarks said that those will be excluded, right?

ralfhandl:
> DoNotCheckScaleOfMeasuredQuantity and RecommendationList look also like property-and parameter-specific annotations.

I think these make sense on a reusable data type that can be used in multiple properties or parameters.

vocabularies/UI.xml
ralfhandl: MultiLineText makes sense to me on a reuse data type.

Not sure about Placeholder, that probably is usage-specific and not reusable.

ralfhandl, line 1408:
```suggestion
      <Term Name="Placeholder" Type="Edm.String" Nullable="false" AppliesTo="Property Parameter">
```

HeikoTheissen (approved):

BerSie (approved):

PR #245, opened 2023-04-04
Fix: ODM.root on entity type, wording aligned with SAP/openapi-specification

ralfhandl:
Aligned with https://github.com/SAP/openapi-specification/pull/11
- long description
- applies to entity type === structure definition, not entity set === resource collection


swennemers (approved):
Good catch. Thanks for the fix.

PR #246, opened 2023-04-04
Chore: ODM.entity name and ODM.oid no longer experimental

ralfhandl:
These two annotations are now translated by CAP's OpenAPI generator into the corresponding `x-sap-odm-` extension keywords defined in https://github.com/SAP/openapi-specification/tree/main/sap-schemas/v3.0#x-sap-odm-entity-name


HeikoTheissen (approved):

ralfhandl:
@swennemers We could park this PR until RAP has implemented these annotations.

Developers of CAP applications may be reluctant to use these terms because they will see this tooltip:

image

swennemers (approved):
I had a quick discussion with the ODM team and avoiding the warning in the CAP tooling is a good argument to remove the experimental flag. Changes to the terms are anyhow not easy anymore.

PR #247, opened 2023-04-05
Actions for creating and moving hierarchy nodes

HeikoTheissen:
We restrict ourselves to the single-parent use case for now.

For multiple parents, the ChangeNext/PreviousSibling actions need an additional parameter `Parent` that takes the key of the parent node.


vocabularies/Hierarchy.xml
```diff
@@ -228,6 +245,91 @@ the following collections of hierarchy nodes are distinguished:
...
+          <Annotation Term="Core.OptionalParameter">
+            <Record/>
+          </Annotation>
+        </Parameter>
+        <Parameter Name="Parent" Type="Edm.PrimitiveType" Nullable="true">
+          <Annotation Term="Core.Description" String="Node identifier of the new node's parent (null if the new node shall be a root)" />
+        </Parameter>
+        <Parameter Name="SiblingRank" Type="Edm.Int64" Nullable="true">
+          <Annotation Term="Core.Description" String="Optional integer rank of the new node among its siblings" />
+        </Parameter>
```
HeikoTheissen, line 272: Or rather the node identifier of the `RightSibling` (null means new node becomes the rightmost sibling)?

vocabularies/Hierarchy.json
```diff
@@ -140,6 +155,118 @@
...
+            "$Name": "Parent",
+            "$Type": "Edm.PrimitiveType",
+            "$Nullable": true,
+            "@Core.Description": "Node identifier of the new node's parent (null if the new node shall be a root)"
+          },
+          {
+            "$Name": "SiblingRank",
+            "$Type": "Edm.Int64",
+            "$Nullable": true,
+            "@Core.Description": "Optional integer rank of the new node among its siblings"
```
uhlmannm, line 192: Would that not be a @Core.OptionalParameter as well?

vocabularies/Hierarchy.json
```diff
@@ -140,6 +155,118 @@
...
+            "$Nullable": true,
+            "@Core.Description": "Node identifier of a parent from which the node shall be removed (null if the node shall no longer be a root)",
+            "@Core.LongDescription": "This parameter is ignored if `KeepParents` is true.\n            Otherwise, for a node with only one parent, it defaults to that parent's node identifier (null if the node is only a root).",
+            "@Core.OptionalParameter": {}
+          },
+          {
+            "$Name": "SiblingRank",
+            "$Type": "Edm.Int64",
+            "$Nullable": true,
+            "@Core.Description": "Optional integer rank of the node among its siblings under the new parent"
```
uhlmannm, line 240: Would that not be a @Core.OptionalParameter as well?

vocabularies/Hierarchy.json
```diff
@@ -140,6 +155,118 @@
...
+            "$Name": "Entity",
+            "$Type": "Edm.EntityType",
+            "@Core.Description": "The entity to become a new node",
+            "@Core.LongDescription": "If the entity exists already without a node identifier for the recursive hierarchy,\n            it is upserted with PATCH semantics.\n            It is an error if the entity exists already with a node identifier for the recursive hierarchy."
+          },
+          {
+            "$Name": "Node",
+            "$Type": "Edm.PrimitiveType",
+            "@Core.Description": "The node identifier, defaults to the node property of the `Entity`",
+            "@Core.LongDescription": "To insert an existing entity with key property `ID` as a new node with server-determined node identifier,\n            provide the parameter `\"Entity\": {\"ID\": \"...\"}` and omit `Node`.",
```
uhlmannm, line 179: What is the purpose of "Node"? When would it ever be provided?

HeikoTheissen: You can always specify the node identifier property of `Entity` instead. But the `Node` parameter offers convenience if you don't know the name of/path to that property and only know the value.

vocabularies/Hierarchy.xml
uhlmannm: Would specifying the node before or after it in the list, mean that the client does not need to know about the SiblingRank? If so, I am in favor of this simplification.

vocabularies/Hierarchy.json
uhlmannm: But then `Entity` still needs to be provided?
I would assume that our client knows where the node property is in `Entity`.

vocabularies/Hierarchy.json
```diff
@@ -140,6 +155,115 @@
...
+            "@Core.Description": "Node identifier of the node to be moved"
+          },
+          {
+            "$Name": "Parent",
+            "$Type": "Edm.PrimitiveType",
+            "$Nullable": true,
+            "@Core.Description": "Node identifier of the node's new parent (null if the node shall become a root)"
+          },
+          {
+            "$Name": "KeepParents",
```
ThomasChadzelek, line 218: We could avoid both KeepParents and RemoveParent if we would accept a 2nd action call to a RemoveAction. Keeping would be the default and removing done via that 2nd call. What do you think?

vocabularies/Hierarchy.json
```diff
@@ -140,6 +155,115 @@
...
+            "@Core.LongDescription": "If the entity exists already without a node identifier for the recursive hierarchy,\n            it is upserted with PATCH semantics.\n            To insert an existing entity with key property `ID` as a new node with server-determined node identifier,\n            provide the parameter `\"Entity\": {\"ID\": \"...\"}` without node identifier property.\n            It is an error if the entity exists already with a node identifier for the recursive hierarchy."
+          },
+          {
+            "$Name": "Parent",
+            "$Type": "Edm.PrimitiveType",
+            "$Nullable": true,
+            "@Core.Description": "Node identifier of the new node's parent (null if the new node shall be a root)"
+          },
+          {
+            "$Name": "RightSibling",
```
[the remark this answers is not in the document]
ThomasChadzelek: OK, I see that you come from SiblingRank, but I'm too stupid to use this tool, sorry. I cannot see the discussions leading to this decision.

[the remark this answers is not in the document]
HeikoTheissen: @uhlmannm would prefer right sibling over sibling rank. And right sibling = null can be used for the (common?) case where the new node is the rightmost sibling.

vocabularies/Hierarchy.json
HeikoTheissen: In a hierarchy where only one parent is allowed, keeping as default would not work. Moreover, I don't want two requests for the common case where a node is moved from one parent to another.

uhlmannm (approved):

vocabularies/Hierarchy.json
```diff
@@ -86,6 +86,31 @@
         "$Type": "Edm.PropertyPath",
         "$Nullable": true,
         "@Core.Description": "Integer property of type `Edm.Int64` holding the the number of [matching](#MatchCount) descendants a node has in the unlimited hierarchy"
+      },
+      "CreateAction": {
+        "$Type": "Core.QualifiedActionName",
+        "$Nullable": true,
+        "@Core.Description": "Action that creates a new node, following [this template](#Template_CreateAction)"
+      },
+      "CreateChildAction": {
```
uhlmannm, line 95: And then the CreateAction is needed only for creating root nodes.

HeikoTheissen: No, it is also needed for creating child nodes in a separate request. Because the instance-bound action is meant for batch requests only: https://github.com/SAP/odata-vocabularies/blob/ba3cb87f7b7f8fcafd46a36864ca99ebd57021cf/vocabularies/Hierarchy.xml#L272-L277

vocabularies/Hierarchy.json
```diff
@@ -146,6 +181,188 @@
...
+            "$Type": "Edm.EntityType",
+            "$Nullable": true,
+            "@Core.Description": "Key of the node's next sibling under the new parent (null if the node shall become the last sibling)",
+            "@Core.LongDescription": "In the absence of this parameter, the sibling position is determined by the service.",
+            "@Core.OptionalParameter": {}
+          }
+        ]
+      }
+    ],
+    "Template_MoveChildAction": [
```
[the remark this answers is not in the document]
HeikoTheissen: Yes. Actions come in two flavors:
- one bound to the child entity (or child entity collection, in the case of create)
- one bound to the parent entity for cases where the parent has been created as part of the same batch request.

(The remove action does not need the second flavor.)

vocabularies/Hierarchy.json
```diff
@@ -86,6 +86,26 @@
...
+      },
+      "ChangePreviousSiblingAction": {
+        "$Type": "Core.QualifiedActionName",
+        "$Nullable": true,
+        "@Core.Description": "Action that moves a node before a newly-created sibling, following [this template](#Template_ChangePreviousSiblingAction)"
+      },
+      "InsertAction": {
+        "$Type": "Core.QualifiedActionName",
+        "$Nullable": true,
+        "@Core.Description": "Action that inserts a new root into a recursive hierarchy, following [this template](#Template_InsertAction)"
```
uhlmannm, line 103: "... inserts an additional root ..." ?

HeikoTheissen: "inserts a root", because this can also be the first node at all in the hierarchy.

vocabularies/Hierarchy.json
```diff
@@ -86,6 +86,26 @@
...
+      },
+      "ChangePreviousSiblingAction": {
+        "$Type": "Core.QualifiedActionName",
+        "$Nullable": true,
+        "@Core.Description": "Action that moves a node before a newly-created sibling, following [this template](#Template_ChangePreviousSiblingAction)"
+      },
+      "InsertAction": {
+        "$Type": "Core.QualifiedActionName",
+        "$Nullable": true,
+        "@Core.Description": "Action that inserts a root into a recursive hierarchy, following [this template](#Template_InsertAction)"
```
uhlmannm, line 103: Discussion on 30.05.23: Should it not just insert the entity into the hierarchy? It will be a root if it does not have a parent.

HeikoTheissen:
To be finally reviewed in V4 tree meeting on 2023-06-30.

vocabularies/Hierarchy.xml
```diff
@@ -145,6 +145,34 @@ the following collections of hierarchy nodes are distinguished:
...
+        <Annotation Term="Core.LongDescription">
+          <String>The qualified action names identify actions for maintaining nodes in the recursive hierarchy,
+which are specific for the given annotation qualifier.
+These actions MUST have the same signature as the template actions linked below, with
+`Edm.EntityType` replaced with the entity type on which the recursive hierarchy is defined.
+In parameters that are defined as "Key of …", only the key properties are relevant.
+
+The template actions themselves cannot be invoked.</String>
+        </Annotation>
+        <Property Name="ChangeNextSiblingAction" Type="Core.QualifiedActionName" Nullable="true">
```
HeikoTheissen, line 162: Add `ChangePreviousSiblingAction`.

vocabularies/Hierarchy.json
```diff
@@ -146,6 +183,74 @@
...
+    "Template_ChangePreviousSiblingAction": [
+      {
+        "$Kind": "Action",
+        "$IsBound": true,
+        "@Common.Experimental": true,
+        "@Core.Description": "Template for actions that move a node among its siblings and are named in [`RecursiveHierarchyActions/ChangePreviousSiblingAction`](#RecursiveHierarchyActionsType)",
+        "$Parameter": [
+          { "$Name": "Node", "$Type": "Edm.EntityType", "@Core.Description": "The node T to be moved" },
+          {
+            "$Name": "NextSibling",
```
uhlmannm, line 214: PreviousSibling

uhlmannm (approved):

uhlmannm (approved):

vocabularies/Hierarchy.xml
```diff
@@ -220,6 +254,93 @@ the following collections of hierarchy nodes are distinguished:
...
+      <Action Name="Template_MakeStartNodeAction" IsBound="true">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Template for actions that make a node a start node and are named in [`RecursiveHierarchyActions/MakeStartNodeAction`](#RecursiveHierarchyActionsType)" />
+        <Annotation Term="Core.LongDescription">
+          <String>Depending on the definition of start node in the recursive hierarchy, this action may fail for certain nodes.
+          If the [`RecursiveHierarchy/ParentNavigationProperty`](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.html#RecursiveHierarchy)
+          is single-valued, the new start node loses a parent it may have had.</String>
+        </Annotation>
+        <Parameter Name="Node" Type="Edm.EntityType" Nullable="false">
+          <Annotation Term="Core.Description" String="The node to be made a root" />
```
ThomasChadzelek, line 266: "root" vs. "start node"?

ThomasChadzelek (approved):

vocabularies/Hierarchy.xml
```diff
@@ -145,6 +145,44 @@ the following collections of hierarchy nodes are distinguished:
...
+        <Annotation Term="Core.LongDescription">
+          <String>The qualified action names identify actions for maintaining nodes in the recursive hierarchy,
+which are specific for the given annotation qualifier.
+These actions MUST have the same signature as the template actions linked below, with
+`Edm.EntityType` replaced with the entity type on which the recursive hierarchy is defined.
+In parameters that are defined as "Key of …", only the key properties are relevant.
+
+The template actions themselves cannot be invoked.</String>
+        </Annotation>
+        <Property Name="MakeStartNodeAction" Type="Core.QualifiedActionName" Nullable="true">
```
HeikoTheissen, line 166: Use a local action name (from the same schema) instead of a qualified action name?

vocabularies/Hierarchy.xml
```diff
@@ -145,6 +145,44 @@ the following collections of hierarchy nodes are distinguished:
...
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Actions for maintaining the recursive hierarchy defined by the [base term](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.html#RecursiveHierarchy)" />
+        <Annotation Term="Core.LongDescription">
+          <String>When this annotation is present, the [`ParentNavigationProperty`](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.html#RecursiveHierarchyType)
+          in the base term must not have a collection-valued segment prior to its last segment.</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="RecursiveHierarchyActionsType">
+        <Annotation Term="Core.LongDescription">
+          <String>The qualified action names identify actions for maintaining nodes in the recursive hierarchy,
```
HeikoTheissen, line 158:
```suggestion
          <String>The action names identify actions from the schema of the annotated entity type for maintaining nodes in the recursive hierarchy,
```

vocabularies/Hierarchy.xml
ThomasChadzelek: No, that's the wrong direction. This is an annotation - what would "same schema" mean? You should really provide the qualified action name, incl. the full namespace. Do not use an alias unless you don't need this to work with UI5 ;-)

HeikoTheissen, line 166:
```suggestion
        <Property Name="MakeStartNodeAction" Type="Edm.String" Nullable="true">
```
Since we disallow aliases here, we cannot use the type `Core.QualifiedActionName`.

vocabularies/Hierarchy.xml
HeikoTheissen, line 158:
```suggestion
          <String>The properties of this type are namespace-qualified action names that identify actions for maintaining nodes in the recursive hierarchy,
```

uhlmannm (approved):

ThomasChadzelek (approved):

uhlmannm (approved):

vocabularies/Hierarchy.xml
```diff
@@ -167,13 +167,17 @@ The template actions themselves cannot be invoked.</String>
         <Property Name="IsNode" Type="Edm.Boolean" Nullable="true">
           <Annotation Term="Core.Description" String="Whether the entity is a node in the recursive hierarchy" />
           <Annotation Term="Core.LongDescription">
-            <String>This property is null if membership in the recursive hierarchy is not exposed in the metadata.</String>
+            <String>This property is null if membership in the recursive hierarchy is not exposed in the metadata.
+            Otherwise, if it is given as a dynamic expression based on an entity property, that property
+            may be changed move an entity into or out of the hierarchy.</String>
```
ThomasChadzelek, line 172: "may be changed move an entity" - is there a "to" missing right before "move"?

PR #248, opened 2023-04-11
Allow setting initial strings to ValueListConstants

BerSie:
We have already a property "InitialValueIsSignificant" for ValueListParameterIn and ValueListParameterInOut. With this change this property will be added to ValueListParameterConstant as well. 
Besides that the experimental tag is removed from the above mentioned properties as well as from ValueListParameterConstant itself.


ralfhandl (approved):

ralfhandl:
Hi Bernhard,

Wie hast Du denn diesen Branch angelegt:

***@***.***

99 commits behind 🤔

Hab mal main gemergt und die neue Property mit “Experimental” markiert: bitte gegenprüfen.

Vielen Dank im Voraus
Ralf

BerSie:
Looks like I used for fetch and rebase a branch with a name that must not be mentioned ... (will not happen again - I removed it locally as well now). @ralfhandl: thanks for correcting it

PR #254, opened 2023-05-23
Node actions corresponding to RAP actions

HeikoTheissen:


vocabularies/Hierarchy.xml
```diff
@@ -158,26 +156,14 @@ The template actions themselves cannot be invoked.
           <Annotation Term="Core.Description" String="Whether the service provides the [`StartNodeCounts`](#StartNodeCounts) instance annotation" />
         </Property>
         -->
-        <Property Name="CreateAction" Type="Core.QualifiedActionName" Nullable="true">
-          <Annotation Term="Core.Description" String="Action that creates a new node, following [this template](#Template_CreateAction)" />
+        <Property Name="LinkAction" Type="Core.QualifiedActionName" Nullable="true">
+          <Annotation Term="Core.Description" String="Action that creates a new node, following [this template](#Template_LinkAction)" />
```
ralfhandl, line 160: Adjust text to match property name

vocabularies/Hierarchy.xml
```diff
@@ -158,26 +156,14 @@ The template actions themselves cannot be invoked.
...
         </Property>
-        <Property Name="CreateChildAction" Type="Core.QualifiedActionName" Nullable="true">
-          <Annotation Term="Core.Description" String="Instance-bound action that creates a new non-root node, following [this template](#Template_CreateChildAction)" />
+        <Property Name="UnlinkAction" Type="Core.QualifiedActionName" Nullable="true">
+          <Annotation Term="Core.Description" String="Action that creates a new node, following [this template](#Template_UnlinkAction)" />
```
ralfhandl, line 163: Adjust text to match property name

ralfhandl (changes requested):

ralfhandl (commented):
If these actions are RAP-specific, should we move them to a RAP vocabulary?

Or are we confident that CAP backends would also be able to implement them?

vocabularies/Hierarchy.xml
```diff
@@ -256,184 +242,72 @@ The template actions themselves cannot be invoked.
...
+      <Action Name="Template_LinkAction" IsBound="true">
         <Annotation Term="Common.Experimental" />
-        <Annotation Term="Core.Description" String="Template for instance-bound actions that create a new non-root node and are named in [`RecursiveHierarchy/CreateChildAction`](#RecursiveHierarchyType)" />
+        <Annotation Term="Core.Description" String="Template for actions that link a node to a parent and are named in [`RecursiveHierarchy/LinkAction`](#RecursiveHierarchyType)" />
         <Annotation Term="Core.LongDescription">
-          <String>This action can be used when creating several nodes in one JSON batch request like:
+          <String>In single-parent hierarchies, the node is first unlinked from its current parent, if any.
+
+To create a new node as child of an existing parent, this action can be used in a JSON batch request like:
 ```
```
ralfhandl, line 252:
```suggestion
```json
```

HeikoTheissen: Not valid JSON, because the code contains ...

HeikoTheissen:
> If these actions are RAP-specific, should we move them to a RAP vocabulary?

No, because the properties `RecursiveHierarchy/LinkAction` etc. are optional anyway, CAP backends could omit them (and perhaps use other optional CAP-specific properties).

PR #256, opened 2023-05-25
FIORITECHP1-27226

HeikoTheissen:
https://jira.tools.sap/browse/FIORITECHP1-27226


vocabularies/UI.xml
```diff
@@ -1837,6 +1837,31 @@ This switches off the validation of user input with respect to decimals.</String
...
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Rules for hierarchy nodes that a UI can guide the user towards" />
+        <Annotation Term="Core.LongDescription">
+          <String>For example, dropping a node on another node can be prevented if the dropped node cannot
+          be a child of the target node.</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="RecursiveHierarchyType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="ChildRules" Type="Collection(UI.RecursiveHierarchyPasteType)" Nullable="false">
```
BerSie, line 1850: Do you mean "UI.RecursiveHierarchyChildType", which is defined below?

vocabularies/UI.xml
```diff
@@ -1837,6 +1837,31 @@ This switches off the validation of user input with respect to decimals.</String
...
+          <Annotation Term="Core.Description" String="Which nodes can be children of this node" />
+        </Property>
+      </ComplexType>
+      <ComplexType Name="RecursiveHierarchyChildType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Rule for nodes of a given type (specified by the [`Hierarchy.RecursiveHierarchy/NodeTypeProperty`](Hierarchy.md#RecursiveHierarchyType))" />
```
BerSie, line 1856: I didn't find a RecursiveHierarchy/NodeTypeProperty (neither in the SAP Hierarchy vocabulary nor in the OASIS TCS Aggregation vocabulary. Is this something to be defined then?

BerSie (commented):
Thank you very much, @HeikoTheissen , starting with that.

vocabularies/UI.xml
HeikoTheissen: It is in the release candidate branch:
https://github.com/SAP/odata-vocabularies/blob/9500b9cc688cae3eb5ca1b71ac8c9b0d83f1d407/vocabularies/Hierarchy.xml#L89

vocabularies/UI.xml
```diff
@@ -1837,6 +1837,31 @@ This switches off the validation of user input with respect to decimals.</String
...
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Rules for hierarchy nodes that a UI can guide the user towards" />
+        <Annotation Term="Core.LongDescription">
+          <String>For example, dropping a node on another node can be prevented if the dropped node cannot
+          be a child of the target node.</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="RecursiveHierarchyType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="ChildRules" Type="Collection(UI.RecursiveHierarchyChildType)" Nullable="false">
```
GeraldKrause, line 1850: Given an entity type that
- has a a NodeType property NT
- and is annotated with UI.RecursiveHierarchy where the Hierarchy.RecursiveHierarchy/NodeTypeProperty points to NT

Imagine NT would take three values A, B and C. How would I define that B and C can be children of A, but does not apply the other way around?

I think a ParentType is also needed, which would lead to: ChildRules is a collection of (ParentType, collection of (RecursiveHierarchyChildType))

HeikoTheissen:
```json
"ChildRules": [
  {"ChildType": "B", "Allowed": {"$Eq": [{"$Path": "NodeType"}, "A"]}},
  {"ChildType": "C", "Allowed": {"$Eq": [{"$Path": "NodeType"}, "A"]}}
]
```

GeraldKrause: With this proposal above, a single lookup is required to determine the answer for a _given child node_.

We can also look at it from the opposite direction like this:
```
"ChildRules": [
  { "ParentType": "A", "AllowedChildren": ["$In": [   { "$Path": "NodeType"  }, [ "B", "C" ] ] ] } ]
``` 

With this approach, a single lookup is required to determine the answer for a _given parent target node_.

If the use cases at hand have a "preferred" direction, we could select the better fitting approach.  

Maybe a question to @BerSie if there is such a preference.

HeikoTheissen: But then it should be `ParentRules`, not `ChildRules`.

BerSie: You're so fast ... I' ve to ask at least one silly question: you are writing "{"$Path": "NodeType"}" - shouldn't it be _NT_ instead of _NodeType_?
Wrt the question of direction - for the drag'n'drop use case we have the child type and with that we know which parents are possible targets (and which not). The same is valid if there is per type a create function. If there is only one create and the child type is selected later on, we have the parent type and the user can choose from the valid child types. The scenario I currently know has the first two use cases.
And there is another question - how to express the extreme cases: all children can be created under one parent, nothing can be created under one parent and all parents can serve as parent for this child type, none of the parents can serve as a parent for this child type (depending on the type of direction we choose).

PR #260, opened 2023-06-20
HTML5.LinkTarget

HeikoTheissen:


vocabularies/HTML5.xml
```diff
@@ -49,6 +49,16 @@
         </Property>
       </ComplexType>
 
+      <Term Name="LinkTarget" Type="Edm.String" Nullable="true">
```
ralfhandl, line 52: How about making this a type definition with AllowedValues, describing (a subset of) the generic link targets `_self`, `_blank`, `_parent`, and `_top`?

ralfhandl (changes requested):

vocabularies/HTML5.xml
HeikoTheissen: This is the HTML5 vocabulary, why not use the HTML5 terms here?

ralfhandl (changes requested):

vocabularies/HTML5.json
```diff
@@ -36,22 +39,18 @@
...
     },
     "LinkTargetType": {
       "$Kind": "TypeDefinition",
       "$UnderlyingType": "Edm.String",
+      "@Core.Description": "The values are interpreted like the [target attribute](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-target) in HTML",
       "@Validation.AllowedValues": [
-        { "Value": "_self", "@Core.Description": "Open link in the current browsing context" },
-        { "Value": "_blank", "@Core.Description": "Open link in a new browsing context" },
-        { "Value": "_parent", "@Core.Description": "Open link in the parent browsing context" },
-        { "Value": "_top", "@Core.Description": "Open link in the top browsing context" }
```
ralfhandl: I liked these descriptions, the table of allowed values looks empty without them.

vocabularies/UI.xml
```diff
@@ -1018,6 +1023,11 @@ The trend is
         <Property Name="Url" Type="Edm.String" Nullable="false">
           <Annotation Term="Core.IsURL" />
           <Annotation Term="Core.Description" String="URL of referenced information" />
+          <Annotation Term="Validation.ApplicableTerms">
```
BerSie, line 1026: what should happen to the facet if you annotate it with "LinkTarget" - how should the UI behave?

HeikoTheissen: When the UI contains a link to that URL, the link shall open according to the `HTML5.LinkTarget`. In other words: The `sap.m.Link` element shall get a corresponding `target` property.

ralfhandl (approved):

BerSie (approved):

PR #265, opened 2023-07-07
PDF: IANA time zone-support

fb0d3nheimer:


vocabularies/PDF.xml
```diff
@@ -99,6 +99,12 @@
             <String>Max result size for PDF documents. Used if $top has been provided and $top > ResultSizeMaximum</String>
           </Annotation>
         </Property>
+        <Property Name="IANATimezoneFormat" Type="Edm.Boolean">
```
HeikoTheissen, line 102:
```suggestion
        <Property Name="IANATimezoneFormat" Type="Edm.Boolean" Nullable="false">
```
There is no difference between false and absent in this case.

vocabularies/PDF.xml
```diff
@@ -99,6 +99,12 @@
...
+          <Annotation Term="Core.Description" String="Formatting DateTimeOffset in IANA-time zone supported" />
```
HeikoTheissen, line 103: Don't understand what this means. Can you give an example of an `Edm.DateTimeOffset` value that is "formatted in a IANA-time zone"?

Is `DateTimeOffset` another property? Where does it come from?

HeikoTheissen (changes requested):

vocabularies/PDF.xml
fb0d3nheimer: Hi Heiko,

here is the setup:
The timestamp contains value  "2017-04-20T13:37:00Z"
My time zone (IANA) is Europe/Berlin

The result in PDF will be "20.04.2017 15:37:00, 0000000 Europe/Berlin"

Let me know if you need further information or examples.

The description could maybe be a little bit more precise. What is actually meant is "Formatting properties with type EDM.DateTimeOffset in IANA-time zone supported". So maybe it is enough to add the "EDM." before "DateTimeOffset"? 

Best regards,

Florian

fb0d3nheimer: Hi Heiko,

there are two possibilities:

1. By defaults time stamp fields shall be shown in the users time zone (SU3/SU01)
2. There shall also be a support for showing a field in a dedicated IANA time zone

I'll send you the link to our one-pager with more information.

Best regards,

Florian

vocabularies/PDF.xml
```diff
@@ -99,6 +99,12 @@
             <String>Max result size for PDF documents. Used if $top has been provided and $top > ResultSizeMaximum</String>
           </Annotation>
         </Property>
+        <Property Name="IANATimezoneFormat" Type="Edm.Boolean" Nullable="false">
+          <Annotation Term="Core.Description" String="Formatting DateTimeOffset in IANA-time zone supported" />
+          <Annotation Term="Core.LongDescription">
+            <String>DateTimeOffset can be formatted in user's IANA time zone or by a defined time zone-property.</String>
+          </Annotation>
+        </Property>
```
HeikoTheissen, line 107:
```suggestion
        <Property Name="TimezoneAware" Type="Edm.Boolean" Nullable="false">
          <Annotation Term="Core.Description" String="Whether points in time are represented in the time zone specified by the [`Common.Timezone`](Common.md#Timezone) annotation" />
        </Property>
```

HeikoTheissen (changes requested):

vocabularies/PDF.xml
HeikoTheissen, line 107:
```suggestion
        <Property Name="IANATimestampFormat" Type="Edm.Boolean" Nullable="false">
          <Annotation Term="Core.Description" String="If this is true, the PDF format supports the value `IANA-Timestamp` in `com.sap.gateway.default.iwbep.common.v0001.TableColumn/Format/DisplayFormat`" />
          <Annotation Term="Core.LongDescription">
            <String>When a table column has `Format/DisplayFormat = 'IANA-Timestamp'`, it is formatted the in the time zone contained in the `Format/IANATimezoneProperty` or by default in the user's time zone.</String>
          </Annotation>
        </Property>
```

HeikoTheissen (changes requested):

vocabularies/PDF.xml
fb0d3nheimer: Hi Heiko,

I discussed your suggestion with the team and we think it is too detailed. With the annotation we just want to announce, that we support IANA time stamp formatting on a server.
Also notice that there is at least one different OData-Implementation at SAP. I am pretty sure, that they don't use our common service for describing their document description.

Best regards,

```suggestion
  <Property Name="IANATimestampFormat" Type="Edm.Boolean" Nullable="false">
          <Annotation Term="Core.Description" String="If this is true, the PDF format supports formatting columns in IANA-time zone" />
          <Annotation Term="Core.LongDescription">
            <String>A table column can be formatted using user's IANA-time zone or an IANA-time zone from a specified column.</String>
          </Annotation>
        </Property>
```

fb0d3nheimer: Hi Heiko,
meanwhile we had to introduce a third way to set the IANA timezone. Therefore, I adjusted my previous suggested change slightly.
```suggestion
        <Property Name="IANATimezoneFormat" Type="Edm.Boolean" Nullable="false">
          <Annotation Term="Core.Description" String="If this is true, the PDF format supports formatting columns in IANA-time zone" />
          <Annotation Term="Core.LongDescription">
            <String>>A table column can be formatted using a fixed IANA-time zone, user's IANA-time zone or an IANA-time zone from a specified column.</String>
          </Annotation>
        </Property>
```
Best regards,
Florian

HeikoTheissen: Two things should be made clearer:
- Only columns of type `Edm.DateTimeOffset` can be formatted in this way.
- Where is the IANA time zone taken from? "Fixed IANA time zone" and "IANA time zone from a specified column" leave that unclear.

HeikoTheissen:
```suggestion
        <Property Name="IANATimezoneFormat" Type="Edm.Boolean" Nullable="false">
          <Annotation Term="Core.Description" String="If this is true, the PDF format supports formatting columns of type `Edm.DateTimeOffset` in a IANA time zone given in the document description" />
        </Property>
```
If the LongDescription does not contain additional information, let's omit it.

HeikoTheissen (changes requested):

HeikoTheissen (approved):

PR #270, opened 2023-07-25
Remove todo

HeikoTheissen:


ralfhandl (changes requested):
Please replace TODO with its outcome, that is either a positive or a negative statement on whether "double annotating" is necessary.

PR #272, opened 2023-08-02
Pdf treeview

sap-mrein:
Added treeview feature to PDF vocabulary (cf https://jira.tools.sap/browse/GATEWAY-731)

### DIFF
--- a/vocabularies/PDF.xml
+++ b/vocabularies/PDF.xml
@@ -102,6 +102,12 @@
         <Property Name="IANATimezoneFormat" Type="Edm.Boolean" Nullable="false">
           <Annotation Term="Core.Description" String="If this is true, the PDF format supports formatting columns of type `Edm.DateTimeOffset` in a IANA time zone given in the document description" />
         </Property>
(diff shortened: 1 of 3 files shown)


vocabularies/PDF.xml
```diff
@@ -102,6 +102,12 @@
         <Property Name="IANATimezoneFormat" Type="Edm.Boolean" Nullable="false">
           <Annotation Term="Core.Description" String="If this is true, the PDF format supports formatting columns of type `Edm.DateTimeOffset` in a IANA time zone given in the document description" />
         </Property>
+        <Property Name="Treeview" Type="Edm.Boolean" Nullable="false">
```
HeikoTheissen, line 105:
```suggestion
        <Property Name="Treeview" Type="Edm.Boolean" Nullable="false">
          <Annotation Term="Common.Experimental" />
```

HeikoTheissen (changes requested):
Shouldn't this be experimental at first?

HeikoTheissen (approved):

PR #273, opened 2023-08-03
Remove rendering proposal from DataFieldWithAction

BerSie:
The proposal for a DataFieldWithAction (to render it as a link) might be misleading, as actions are usually connected with buttons or similar.

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1678,7 +1678,7 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
 
       <ComplexType Name="DataFieldWithAction" BaseType="UI.DataField">
         <Annotation Term="Core.Description" String="A piece of data that allows triggering an OData action" />
-        <Annotation Term="Core.LongDescription" String="The action is tied to a data value. This is in contrast to [DataFieldForAction](#DataFieldForAction)) which is not tied to a specific data value." />
+        <Annotation Term="Core.LongDescription" String="The action is tied to a data value. This is in contrast to [DataFieldForAction](#DataFieldForAction) which is not tied to a specific data value." />
         <Property Name="Value" Type="Edm.PrimitiveType" Nullable="false" />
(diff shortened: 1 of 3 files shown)


vocabularies/UI.xml
```diff
@@ -1678,7 +1678,7 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
 
       <ComplexType Name="DataFieldWithAction" BaseType="UI.DataField">
         <Annotation Term="Core.Description" String="A piece of data that allows triggering an OData action" />
-        <Annotation Term="Core.LongDescription" String="The action is tied to a data value which should be rendered as a hyperlink. This is in contrast to [DataFieldForAction](#DataFieldForAction)) which is not tied to a specific data value." />
+        <Annotation Term="Core.LongDescription" String="The action is tied to a data value. This is in contrast to [DataFieldForAction](#DataFieldForAction)) which is not tied to a specific data value." />
```
HeikoTheissen, line 1681:
```suggestion
        <Annotation Term="Core.LongDescription" String="The action is tied to a data value which could be rendered as a button. This is in contrast to [DataFieldForAction](#DataFieldForAction)) which is not tied to a specific data value." />
```
Make a better proposal rather than remove it?

BerSie: No, since I believe usually a term cannot /should not decide that. That's a decision of the renderer and the guideline he has to work with (e.g. the Fiori Design Guidelines - which say in that case it should be usually a button).

HeikoTheissen (dismissed):

vocabularies/UI.xml
ralfhandl, line 1681:
```suggestion
        <Annotation Term="Core.LongDescription" String="The action is tied to a data value. This is in contrast to [DataFieldForAction](#DataFieldForAction) which is not tied to a specific data value." />
```
Remove redundant closing parenthesis

ralfhandl (changes requested):
Minor nit, otherwise fine

ralfhandl (approved):

PR #274, opened 2023-08-11
New format proposal for TopLevels' Expand parameter

HeikoTheissen:
The suggested format allows specifying how many levels deep a node shall be expanded.

For compatibility reasons, a new parameter has been introduced and two old ones deprecated.


HeikoTheissen:
Converted to draft, because this will not be implemented in the first phase.

bernhardgr42:
Hello Heiko,

Old:
 <Parameter Name="Expand" Type="Collection(Edm.String)" Nullable="false">
      
  New:     
       <Parameter Name="Expand" Type="Collection(Hierarchy.TopLevelsExpandType)" Nullable="false">
       
       This is incompatible. Could you create v2 for the new version?

HeikoTheissen:
Would a new function name `Hierarchy.ExpandLevels` instead of `Hierarchy.TopLevels` be sufficient? Then I'd rename the current function in the vocabulary and you can continue to parse the old one (which no-one will ever use)?

ralfhandl (approved):

PR #276, opened 2023-08-17
Actions/functions for recommendations

HeikoTheissen:
See [these diagrams](../blob/Common.Recommendations/docs/recommendations.md).


vocabularies/Common.json
```diff
@@ -1288,6 +1288,48 @@
...
+        "@Core.LongDescription": "The template action itself cannot be invoked.",
+        "$Parameter": [
+          {
+            "$Name": "EntityKeys",
+            "$Type": "Edm.ComplexType",
+            "@Core.Description": "Nested structure containing the keys of entities for which recommendations shall be retrieved",
+            "@Core.LongDescription": "The nested complex type of this parameter has the same structure as the entity type bearing the [`GetRecommendationsAction`](#GetRecommendationsAction) annotation\n            and its related entity types, but contains only their key properties.\n            If this parameter can be included in a URL, a function can be used instead of an action."
+          }
+        ],
+        "$ReturnType": { "$Collection": true, "$Type": "Common.RecommendationPropertyType" }
```
SNairVinod, line 1312: Can we name the type as "PropertyRecommendationType". The reasoning behind is that we are trying to get the recommendations for a property.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,48 @@
...
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "value": {},
+      "text": {},
+      "alternatives": { "$Collection": true, "$Type": "Common.AlternativeRecommendationType" }
+    },
+    "AlternativeRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "value": {},
```
SNairVinod, line 1331: Can we have a description

```json
"value": {
    "@Core.Description": "Recommended value for the property, if it meets a certain criteria."
}
```

vocabularies/Common.json
```diff
@@ -1288,6 +1288,48 @@
       "@Common.Experimental": true,
       "@Core.IsURL": true,
       "@Core.Description": "Base URL for WebSocket connections"
+    },
+    "GetRecommendationsAction": {
```
SNairVinod, line 1292: Can we name this as "RecommendationsAction"

vocabularies/Common.json
HeikoTheissen: "If it meets certain criteria" is too vague for a description. What do you want to express?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,51 @@
...
+            "@Core.LongDescription": "The nested complex type of this parameter has the same structure as the entity type bearing the [`GetRecommendationsAction`](#GetRecommendationsAction) annotation\n            and its related entity types, but contains only their key properties.\n            If this parameter can be included in a URL, a function can be used instead of an action."
+          }
+        ],
+        "$ReturnType": { "$Collection": true, "$Type": "Common.PropertyRecommendationType" }
+      }
+    ],
+    "PropertyRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "target": { "@Core.Description": "Target for this recommendation, like the `target` in [`Messages`](#Messages)" },
```
uhlmannm, line 1318: Is it really sufficient to refer to the target used in the message property? The message property is part of the entity data which the recommendation is not.
Also as the recommendations for several entities may be requested, the key predicates have to be part of the target. So e.g. when fetching recommendations for sales orders 1, 2 and 3, the targets would need to be something like (id=1)/soldToParty?

uhlmannm: Or would the target be relative to the request URL? This does not seem to be helpful. So the targets are relative to the entity set for which the action is provided. Only that the action is annotated at the entity type which might be used for several entity sets (in theory).
An example for targeting a property of a related entity would not hurt either.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,51 @@
...
+    },
+    "Template_RecommendationsAction": [
+      {
+        "$Kind": "Action",
+        "@Common.Experimental": true,
+        "@Core.Description": "Template for actions that retrieve AI-based recommendations for an entity and its related entities",
+        "@Core.LongDescription": "The template action itself cannot be invoked.",
+        "$Parameter": [
+          {
+            "$Name": "EntityKeys",
```
uhlmannm, line 1306: So the action will return recommendations for all properties of the entities for which recommendations are calculated. But the underlying assumption of not providing a mean to limit the properties is that the coupling between service and UI is tightly enough so that there won't be (many) recommendations generated for unavailable fields?

vocabularies/Common.json
HeikoTheissen: We can define the target to be relative to the entity container, if that is preferred.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,51 @@
       "@Common.Experimental": true,
       "@Core.IsURL": true,
       "@Core.Description": "Base URL for WebSocket connections"
+    },
+    "RecommendationsAction": {
+      "$Kind": "Term",
+      "$AppliesTo": ["EntityType"],
```
uhlmannm, line 1294: See below: ... So the targets are relative to the entity set for which the action is provided. Only that the action is annotated at the entity type which might be used for several entity sets (in theory).

vocabularies/Common.json
uhlmannm: Not sure whether this is helpful for FE. I guess relative to the entityset should be good.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,51 @@
...
+        "$Kind": "Action",
+        "@Common.Experimental": true,
+        "@Core.Description": "Template for actions that retrieve AI-based recommendations for an entity and its related entities",
+        "@Core.LongDescription": "The template action itself cannot be invoked.",
+        "$Parameter": [
+          {
+            "$Name": "EntityKeys",
+            "$Type": "Edm.ComplexType",
+            "@Core.Description": "Nested structure containing the keys of entities for which recommendations shall be retrieved",
+            "@Core.LongDescription": "The nested complex type of this parameter has the same structure as the entity type bearing the [`GetRecommendationsAction`](#GetRecommendationsAction) annotation\n            and its related entity types, but contains only their key properties.\n            If this parameter can be included in a URL, a function can be used instead of an action."
```
SNairVinod, line 1309: - GetRecommendationsAction should be changed to RecommendationsAction (text as well as reference
- Current approach increases the effort for the application developer by mandating the type definition for parameter. Is there a way this could be generated on the fly by the Framework

HeikoTheissen: > * Current approach increases the effort for the application developer by mandating the type definition for parameter.

We can try to reuse the entity type.

vocabularies/Common.json
HeikoTheissen: Annotation now targets the entity set.

vocabularies/Common.xml
```diff
@@ -1450,6 +1450,68 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <TypeDefinition Name="NamespaceQualifiedActionName" UnderlyingType="Edm.String">
```
ralfhandl, line 1453: Why namespace-qualified? CSDL XML allows to use namespace or alias as qualifier, and CSDL JSON _requires_ using the alias if one is defined.

vocabularies/Common.xml
```diff
@@ -1450,6 +1450,68 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+      <Action Name="Template_RecommendationsAction" IsBound="true">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Template for actions that retrieve AI-based recommendations for the bound entity and its related entities" />
+        <Annotation Term="Core.LongDescription">
+          <String>The template action itself cannot be invoked.</String>
+        </Annotation>
+        <Parameter Name="Entity" Type="Edm.EntityType" Nullable="false">
+          <Annotation Term="Core.Description" String="Entity relative to which the `EntityKeys` are interpreted" />
+        </Parameter>
+        <Parameter Name="EntityKeys" Type="Edm.EntityType" Nullable="false">
```
ralfhandl, line 1481: What's the purpose of this parameter? The action is instance-bound, so how do these key values relate to the key values of the binding parameter?

vocabularies/Common.xml
```diff
@@ -1450,6 +1450,68 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+        <Property Name="target" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Target for this recommendation, format tbd" />
+        </Property>
+        <Property Name="value" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Recommended value, converted to string" />
+        </Property>
+        <Property Name="text" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.Description" String="Description of the recommended value" />
+        </Property>
+        <Property Name="alternatives" Type="Collection(Common.AlternativeRecommendationType)" Nullable="false">
```
ralfhandl, line 1502: Can the alternatives also appear as items of the returned collection?

vocabularies/Common.xml
HeikoTheissen: `EntityKeys` can contain _nested_ entity keys_, for example
```
{"SalesOrder": "A", "_Items": [{
  "SalesOrderItem": "010"
}, {
  "SalesOrderItem": "020"
}]]
```
if only recommendations for header and items 010 and 020 are desired.

vocabularies/Common.xml
ralfhandl: This is problematic as it violates CSDL rules and is inconsistent with other annotations that reference actions, for example Common.SideEffects in this vocabulary.

vocabularies/Common.xml
HeikoTheissen: There is _one_ recommendation and zero or more alternatives in the return value. What do you mean by "returned collection"?

vocabularies/Common.xml
ralfhandl: Please add this to the description.

ralfhandl: And can the "root" key differ from the key of the binding instance?

vocabularies/Common.xml
ralfhandl: Line 1489 reads
~~~xml
 <ReturnType Type="Collection(Common.PropertyRecommendationType)" Nullable="false" />
~~~

Should the "Collection" be removed?

HeikoTheissen: > Should the "Collection" be removed?

No, there is one `PropertyRecommendationType` per targeted property.

vocabularies/Common.xml
HeikoTheissen: Added to Albatross agenda.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,77 @@
...
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "Function": {
+        "$Type": "Common.QualifiedName",
+        "$Nullable": true,
+        "@Core.Description": "Qualified name of a function that retrieves recommendations (see [template](#Template_RecommendationsFunction))"
+      },
+      "Dependencies": {
+        "$Collection": true,
+        "$Type": "Common.RecommendationsDependency",
```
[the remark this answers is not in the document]
HeikoTheissen: I haven't defined that yet. You did not miss anything.

HeikoTheissen:
Converted to draft because there are still so many questions open:
- Alternative approach: unbound function with a parameter `Entities` that takes a collection of [property-addressing URLs](https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_AddressingaProperty). This could even be a service-independent function, because it does not depend on the entity type.
- The `target` of a recommendation could also be such a property-addressing URL.
- Even more fundamental alternative: service-independent unbound function that takes as parameter a collection of name-value pairs as understood by the AI service. This could also be used to retrieve recommendations for parameters (not just entity properties).

This also makes clear that the desired function is not part of the individual RAP or CAP service. It should  be an OData service of its own, similar to the value help services.

Finally, the recommendations retrieval must not happen in the same sticky session as the main OData service (SD sales order, say), because it would then block other OData requests (sticky requests are executed sequentially by the ABAP backend).

vocabularies/Common.xml
```diff
@@ -1450,6 +1450,81 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+        </Property>
+      </ComplexType>
+      <ComplexType Name="RecommendationsDependency">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="TargetProperty" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Property targeted by a recommendation" />
+        </Property>
+        <Property Name="Predictors" Type="Collection(Edm.PropertyPath)" Nullable="false">
+          <Annotation Term="Core.Description" String="Properties on which the recommendations for the `TargetProperty` depend" />
+        </Property>
```
HeikoTheissen, line 1476: Is a property path sufficient here? Or can a recommendation also depend on properties from the parent entity?

ralfhandl: It is if the annotation targets the parent entity, or if it targets the child entity and the child entity has an "uplink" to its parent, which is often the case for Fiori APIs.

drktfl:
> But then the recommendations can be overtaken by subsequent changes to the values on which they were based

Probably I missed something, but:
I would say this can happen irrespective of the way the function is defined (inside the same service / bound / whatever). Thinking of collaborative editing, where the second PATCH is done by another user. The client needs to be prepared to handle that somehow.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,106 @@
       "@Common.Experimental": true,
       "@Core.IsURL": true,
       "@Core.Description": "Base URL for WebSocket connections"
+    },
+    "Recommendations": {
+      "$Kind": "Term",
+      "$Type": "Common.QualifiedName",
+      "$AppliesTo": ["EntityType"],
+      "@Common.Experimental": true,
+      "@Core.Description": "Qualified name of a function that computes AI-based recommendations for entities of the annotated type (see [template](#Template_RecommendationsFunction))"
```
uhlmannm, line 1297: Will the name of the retrieval function be deduced (and thus need to follow a pattern) or do we need another annotation?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,106 @@
...
+    },
+    "Template_RecommendationsFunction": [
+      {
+        "$Kind": "Function",
+        "$IsBound": true,
+        "@Common.Experimental": true,
+        "@Core.Description": "Template for functions that compute AI-based recommendations for the bound entity and its related entities",
+        "@Core.LongDescription": "If the function is invoked repeatedly in a stateful session, it SHOULD omit\n          recommendations whose predictors have not changed since the previous invocation.\n\n          The template function itself cannot be invoked.",
```
uhlmannm, line 1305: But it should still return the (unchanged) recommendations for the targets whose predictors have not changed?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,106 @@
...
+            "@Core.Description": "URLs (relative to the value of the binding parameter) that address entities, properties or action/function parameters for which recommendations shall be retrieved"
+          }
+        ],
+        "$ReturnType": {
+          "@Core.IsURL": true,
+          "@Core.Description": "URL for retrieving the recommendations (which are computed asynchronously)",
+          "@Core.Example": {
+            "@odata.type": "https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.xml#Core.PrimitiveExampleValue",
+            "Description": "Relative URL for invoking a function import of the [retrieval function](#Template_RecommendationsRetrievalFunction)",
+            "Value": "../RetrieveRecommendations?ID=af43c2"
```
uhlmannm, line 1325: Should this be relative to the binding parameter? Would it then not be "RetrieveRecommendations?ID=af43c2"?

uhlmannm, line 1325: To what is the URL relative? The retrieval function is not bound. So should the retrieval URL then be relative to the service URL? Would it then not be "RetrieveRecommendations?ID=af43c2"?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,106 @@
...
+          "@Core.Description": "A collection of recommendations per targeted property"
+        }
+      }
+    ],
+    "PropertyRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "target": {
+        "@Core.IsURL": true,
+        "@Core.Description": "URL addressing a property [OData-URL, section 4.6] that is targeted by this recommendation"
```
uhlmannm, line 1354: And if it is relative, it is relative to the binding parameter of the previous RecommendationsFunction call?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,106 @@
...
+      },
+      "value": { "@Core.Description": "Recommended value, converted to string" },
+      "text": { "$Nullable": true, "@Core.Description": "Description of the recommended value" },
+      "alternatives": {
+        "$Collection": true,
+        "$Type": "Common.AlternativeRecommendationType",
+        "@Core.Description": "Alternative recommendations"
+      }
+    },
+    "AlternativeRecommendationType": {
```
uhlmannm, line 1364: There is no "text" here (unlike in the PropertyRecommendationsType)?

vocabularies/Common.json
HeikoTheissen: No, they should be omitted.

vocabularies/Common.json
HeikoTheissen: The returned URL is relative to the request URL `SalesOrder(...)/ns.Recommendations`. The retrieval URL is a function import `RetrieveRecommendations?ID=...`, so the relative URL must go one step back.

vocabularies/Common.json
HeikoTheissen: Yes, see my new commit.

vocabularies/Common.json
HeikoTheissen: @SNairVinod, can you answer this?

vocabularies/Common.json
uhlmannm: So whoever calculates this relative URL needs to know the length of the path with which it was called, i.e. SalesOrderItem(...)/ns.Recommendations vs. SalesOrder(...)/_Item(...)/ns.Recommendation. Would it not be easier if it is relative to the service URL? I guess that would make also the binding for FE simpler.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,117 @@
...
+        "@Core.IsURL": true,
+        "@Core.Description": "URL (relative to the URL of the recommendations function request) addressing a property [OData-URL, section 4.6] that is targeted by this recommendation",
+        "@Core.LongDescription": "The request `SalesOrder('A')/ns.RecommendationsFunction?Targets=...`\n            might retrieve a recommendation with `\"target\": \"Items(10)/Product\"`."
+      },
+      "value": { "@Core.Description": "Recommended value, converted to string" },
+      "text": { "$Nullable": true, "@Core.Description": "Description of the recommended value" },
+      "alternatives": {
+        "$Collection": true,
+        "$Type": "Common.AlternativeRecommendationType",
+        "@Core.Description": "Alternative recommendations"
```
SNairVinod, line 1372: Description can be "A list of alternative values, sorted by confidence score in descending order. If a value is recommended via property 'value', it must be the first entry in this list."

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+        "@Core.Description": "Template for functions that compute AI-based recommendations for the bound entity and its related entities",
+        "@Core.LongDescription": "If the function is invoked repeatedly in a stateful session, it SHOULD NOT\nrecompute recommendations whose predictors have not changed since the previous invocation.\n\nThe template function itself cannot be invoked.",
+        "$Parameter": [
+          { "$Name": "Entity", "$Type": "Edm.EntityType" },
+          {
+            "$Name": "Targets",
+            "$Collection": true,
+            "@Core.IsURL": true,
+            "@Core.Description": "URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved",
+            "@Core.LongDescription": "The request `SalesOrder('A')/ns.RecommendationsFunction?Targets=[\"Items(10)\"]`\n            retrieves recommendations for the entity `SalesOrder('A')/Items(10)`."
```
srbsing, line 1313: If recommendations are to be fetched to Sales Order A itself then I assume we will make a request like below? or may be have no targets at all.

SalesOrder('A')/ns.RecommendationsFunction?Targets=[\"SalesOrder(A)\"]

and for navigation properties on Item the calls will look like below.

SalesOrder('A')/ns.RecommendationsFunction?Targets=[\"_Item(10)/_ItemPartner(20)\"]

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+    },
+    "Template_RecommendationsFunction": [
+      {
+        "$Kind": "Function",
+        "$IsBound": true,
+        "@Common.Experimental": true,
+        "@Core.Description": "Template for functions that compute AI-based recommendations for the bound entity and its related entities",
+        "@Core.LongDescription": "If the function is invoked repeatedly in a stateful session, it SHOULD NOT\nrecompute recommendations whose predictors have not changed since the previous invocation.\n\nThe template function itself cannot be invoked.",
+        "$Parameter": [
+          { "$Name": "Entity", "$Type": "Edm.EntityType" },
```
srbsing, line 1307: If we are giving targets do we still need the Entity as a parameter?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+      "@Core.Description": "Role of this property or parameter regarding AI-based recommendations"
+    },
+    "RecommendationsRoleType": {
+      "$Kind": "EnumType",
+      "$IsFlags": true,
+      "@Common.Experimental": true,
+      "Input": 1,
+      "Input@Core.Description": "The property is input for some recommendation",
+      "Output": 2,
+      "Output@Core.Description": "The property or parameter is the target of some recommendation",
```
SNairVinod, line 1414: We can also have a non-mandatory property for recommendation call marked as input & output role. In runtime if the value for this property is passed it will be used as input param (known value) else AI service may predict the value (will be treated as unknown/output param). Therefore we need to have "InputOutput" also as an enum value

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+            "@Core.LongDescription": "The request `SalesOrder('A')/ns.RecommendationsFunction?Targets=[\"Items(10)\"]`\n            retrieves recommendations for the entity `SalesOrder('A')/Items(10)`."
+          }
+        ],
+        "$ReturnType": {
+          "@Core.IsURL": true,
+          "@Core.Description": "URL (relative to the service URL) for retrieving the recommendations (which are computed asynchronously)",
+          "@Core.Example": {
+            "@odata.type": "https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.xml#Core.PrimitiveExampleValue",
+            "Description": "Relative URL for invoking a function import of the [retrieval function](#Template_RecommendationsRetrievalFunction)",
+            "Value": "RetrieveRecommendations?ID=af43c2"
```
srbsing, line 1322: @uhlmannm - I believe we will have split this and then make a call via context.execute() or we can directly make a call to this URL via the model?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+          }
+        ],
+        "$ReturnType": {
+          "$Collection": true,
+          "$Type": "Common.PropertyRecommendationType",
+          "@Core.Description": "A collection of recommendations per targeted property"
+        }
+      }
+    ],
+    "Template_AsyncRecommendationsFunction": [
```
SNairVinod, line 1346: Even thought it is clear to me, we could ignore till the framework implement the support for recommendations correctly, isn't it?

vocabularies/Common.json
HeikoTheissen: > If recommendations are to be fetched to Sales Order A itself then I assume we will make a request like below? or may be have no targets at all.

Either no `Targets` or `Targets=["."]`.

vocabularies/Common.json
HeikoTheissen: Yes, because without the `Entity` parameter, the function would be unbound and relative URLs would not be possible.

vocabularies/Common.json
HeikoTheissen: Yes, this template becomes relevant only when Gateway supports this type of asynchronous requests.

vocabularies/Common.json
HeikoTheissen: Instead of a separate `InputOutput` value, you can give both flags:
```
<Annotation Term="Common.RecommendationsRole"
  EnumMember="Common.RecommendationsRoleType/Input
              Common.RecommendationsRoleType/Output"/>
```

vocabularies/Common.json
uhlmannm: Good point. But I would still expect that relative to the service URL is easier to do for everybody.

vocabularies/Common.json
uhlmannm: As this is a function, you could directly bind the URL. It would not be an operation binding then and `execute` could not be called. It would rather behave like a normal context binding, i.e. `requestObject` should do the trick. Refreshing via `refresh` would also be possible.
You could also split the Url and use an operation binding. But that should be unnecessary effort.

vocabularies/Common.json
nlunets: still says $AppliesTo: EntityType

vocabularies/Common.json
uhlmannm: There is only the `RecommendationsFunction` annotation and it will either list a function following the `` `Template_RecommendationsFunction` or the `Template_AsyncRecommendationsFunction`. FE will need to decide which one is called by looking at the return type? Or will there be another way of distinguishing the two?

uhlmannm (dismissed):

vocabularies/Common.json
uhlmannm: An empty string, i.e. `Targets=[""]`, for targeting the binding parameter?

vocabularies/Common.json
nlunets: Aren't the name for the template incorrect ?

The Template_RecommendationsFunction returns a url where the data can finally be retrieved while this one
Template_AsyncRecommendationsFunction returns the data 

For me the `async` component of all this relates to the asynchronous requests part....
Or are you saying that at no point in time do you consider that the recommendation service can directly return you the recommendation data ?
In case the data is cached this should be a pretty fast call...

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+    ],
+    "PropertyRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "target": {
+        "@Core.IsURL": true,
+        "@Core.Description": "URL (relative to the URL of the recommendations function request) addressing a property [OData-URL, section 4.6] that is targeted by this recommendation",
+        "@Core.LongDescription": "The request `SalesOrder('A')/ns.RecommendationsFunction?Targets=...`\n            might retrieve a recommendation with `\"target\": \"Items(10)/Product\"`."
+      },
+      "value": { "@Core.Description": "Recommended value, converted to string" },
```
nlunets, line 1381: Why would we convert the recommended value to string ?

vocabularies/Common.json
nlunets: Considering that so far we had InOut enums (see valueList) isn't that linked to framework limitation on RAP/CAP side that this cannot be annotated with multiple values and as such we should avoid it ?

vocabularies/Common.json
HeikoTheissen: If the return type is `Edm.String`, FE calls the function and polls the result at the URL returned by the function call.
Otherwise, FE calls the function with `Prefer: respond-async` and polls the result at the `Location` header returned by that request.
Is that acceptable?

vocabularies/Common.json
HeikoTheissen: You are right, the empty string is correct.
```
> new URL("", "http://server/service/SalesOrder('A')").href
"http://server/service/SalesOrder('A')"
> new URL(".", "http://server/service/SalesOrder('A')").href
'http://server/service/'
```

nlunets: From a modeling perspective does it really make sense to have SalesOrder('A')/ns.RecommendationsFunction?Targets=["_Item(10)/_ItemPartner(20)"] when you could just have SalesOrder('A')/_Item(10)/_ItemPartner(20)/ns.RecommendationsFunction ?

If you only consider the bindingParameter case does it really make sense to even specify the targets ? The AI Model generally generate recommendation for a full Tuple regardless of what you would specify (or at least they would have an opinion), as such i'm not even sure Targets are really relevant, or at least i miss the use case

vocabularies/Common.json
nlunets: That i understand yes, but why not also support a simpler case where you already get the `Collection(Common.PropertyRecommendationType)`
In repeat case where the response would already be the same and cached (as indicated in your recommendation), we would avoid an extra call

vocabularies/Common.json
HeikoTheissen: Because `Collection(Edm.PrimitiveType)` is not allowed in OData.

vocabularies/Common.json
HeikoTheissen: It now targets the entity type again, because the entity set can be derived from the binding parameter.

vocabularies/Common.json
nlunets: Isn't that a good use case for Edm.Untyped then ?

HeikoTheissen:
@nlunets:
> Template_AsyncRecommendationsFunction returns the data
The "Async" prefix shall express that this function can be called with `Prefer: respond-async`.

> framework limitation on RAP/CAP side that this cannot be annotated with multiple values
If such a restriction exists, we can change the `EnumType` accordingly. Just let me know.

vocabularies/Common.json
uhlmannm: The targets allow you to bundle the calls for the main object and several items into one function call instead of calling the bound function for each individually. That may help the backend to have a better performance.

vocabularies/Common.json
uhlmannm: @nlunets Using asynchronous requests would also allow that use case, at least in theory. The server may or may not apply the preference, i.e. it could decide whether the request can be responded to quickly and synchronously or not. However, my understanding is that the way GW works, such a distinction will not be possible.

To add another function that always returns the results synchronously would be possible. But how would FE know that it can call this function? The solution sketched in the spec for asynchronous requests, i.e. that the server decides whether to respond asynchronously, seem superior to me.

nlunets:
> The "Async" prefix shall express that this function can be called with `Prefer: respond-async`.

A Capabilities annotation would be nicer in my opinion :)

> If such a restriction exists, we can change the EnumType accordingly. Just let me know.

We have only a handful of enum type defined as flags and i couldn't find example where this is defined in CDS so i would tend to be on the conservative side here :)

vocabularies/Common.json
HeikoTheissen: And bundling is essential for the asynchronous behavior we want to achieve.

vocabularies/Common.json
nlunets: In the http world, asynchronous request with 202 are mostly designed for _long_ running jobs, which can range from tens of second to tens of minutes.
If we already consider that recommendation are going to take that long to be returned, we might as well not work on them :)

I'm not against considering that exceptionally, a service might decide to return them as an async call, but i would consider that most case should be implemented as a direct response.

For FE that's rather easy i'd say:

- If the function has a returnType of recommendation Common.PropertyRecommendationType -> it's a direct response one
- If the function has a returnType of type String -> by convention we can poll
- If the function is annotated with something to indicate it's an asynchronous process on the backend -> Async Request handling

vocabularies/Common.json
HeikoTheissen: Sorry, this is blast from the past, when `alternatives` as still a collection of primitives. But now it is a collection of a complex type, therefore we can use `Edm.PrimitiveType`. Thanks for making me aware of this!

vocabularies/Common.json
HeikoTheissen: The simpler case will be supported once we switch to `Prefer: respond-async`. Then the server can decide whether to respond synchronously or asynchronously. But this is not yet supported by Gateway.

Until then, an additional second request is better than a slower first request.

vocabularies/Common.json
nlunets: I would be very careful with asynchronous request for something like this, especially if you consider bundling is important.
Let me take an example as to what we will be faced with now

User make change on MainObject and realize we need to fetch recommendation for multiple target

First we have

GET MainObject/ns.GetReco(Targets=["","_Item(10)","_Item(20)"]) --> Handle1

This gives us back Handle1 to fetch the recommendation at some point but this is long running

Meanwhile the frontend makes another change and end up doing a call with less or different targets

GET MainObject/ns.GetReco(Targets=["", "_Item(30)]) --> Handle2


Now you can't really delete Handle1, you still need to fetch it because it contains data that Handle2 won't but if by some chance Handle2 comes back before Handle1 (and this will happen because it can), then what do we do ? 

Do we overwrite the result when Handle1 come back ? Do we need to keep track of all request and have the client needing to determine per target what is relevant ? 

The whole asynchronous operations makes everything more complicated and I fail to see any benefit highlighted (beside the singular sticky case which I imagine has other possible solution)

nlunets:
> Capabilities.AsynchronousRequestsSupported can only be applied to the entire service container, not to individual functions.

if only we had a way of defining additional annotations to make our life easier...

vocabularies/Common.json
HeikoTheissen: The client knows that the recommendations obtained via Handle1 must be processed before the ones obtained via Handle2. For now recommendations are too slow to be processed synchronously.

vocabularies/Common.json
uhlmannm: The server can decide. However, the recommendations call has to be requested only after the application of the PATCH. There are two ways how to achieve that. Either it is a sequential request after the $batch with the PATCH or it is part of the same $batch. In the latter case, there is typically also a side effects GET. So if fetching the recommendations takes several seconds, the user would see this as a slower application of side effects.
And with sticky there seems to be the restriction that only one request is processed at a time. And the recommendations call needs to use the sticky session. 
So overall, becoming asynchronous, even if it is for just a few seconds, seems to make sense to not slow down the normal processing with PATCHes and side effects.

ralfhandl:
> > Capabilities.AsynchronousRequestsSupported can only be applied to the entire service container, not to individual functions.
> 
> if only we had a way of defining additional annotations to make our life easier...

[`Prefer: respond=async`](https://www.rfc-editor.org/rfc/rfc7240.html#section-4.1) is just a preference that the server may accept or ignore per request. Even if we could annotate a single function, the server may still choose to process some requests to it synchronously if it decides to do so for whatever reasons.

So if the client finds this annotation on the entity container, it knows that the server may at least sometimes process some requests asynchronously, and the client knows that it makes sense to send this prefer header, hoping for the best.

ralfhandl:
> we can change the `EnumType` accordingly.

Unfortunately this is considered a breaking change that requires a major version bump for an OData service. Which is why we stopped adding new enum types to vocabularies and instead use type definitions annotated with `Validation.AllowedValues`.

nlunets:
> So if the client finds this annotation on the entity container, it knows that the server may at least sometimes process some requests asynchronously, and the client knows that it makes sense to send this prefer header, hoping for the best.

That's my main point, from a client perspective, especially a UI client, it is important to have an idea beforehand if a call might be asynchronous or not.

We also can't assume that all calls may be asynchronous, in my opinion even if the service is annotation with `AsynchronousRequestsSupported` i would like the send the prefer header only when it make sense.

In some case having an asynchronous call means that you might want to have a UI to keep track of the status (think running jobs for instance), that's why i would prefer to have a specific annotation for the Action/Function itself and not just rely on the naming convention as proposed by @HeikoTheissen

vocabularies/Common.json
ralfhandl: > the entity set can be derived from the binding parameter

As long as we only have one entity set per type.

nlunets:
> Unfortunately this is considered a breaking change that requires a major version bump for an OData service. Which is why we stopped adding new enum types to vocabularies and instead use type definitions annotated with Validation.AllowedValues.

Yes but this is a new enum, so that won't apply here, either that or i didn't understand your comment

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+        "@Common.Experimental": true,
+        "@Core.Description": "Template for functions that compute AI-based recommendations for the bound entity and its related entities",
+        "@Core.LongDescription": "If the function is invoked repeatedly in a stateful session, it SHOULD NOT\nrecompute recommendations whose predictors have not changed since the previous invocation.\n\nThe function computes the recommendations asynchronously and returns a URL where its result can be retrieved.\n\nThe template function itself cannot be invoked.",
+        "$Parameter": [
+          { "$Name": "Entity", "$Type": "Edm.EntityType" },
+          {
+            "$Name": "Targets",
+            "$Collection": true,
+            "@Core.IsURL": true,
+            "@Core.Description": "URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved",
```
drktfl, line 1312: > URLs (relative to the request URL)

sounds recursive to me: If the client requests `SalesOrder('A')/ns.RecommendationsFunction?Targets=["","Items(10)"]`, that's the "request URL", isn't it? I do understand that only `SalesOrder('A')` is meant by the term, it becomes clear by the description. I just wonder if there is a better name for it?

vocabularies/Common.json
```diff
@@ -1288,6 +1288,132 @@
...
+    "RecommendationsFunction": {
+      "$Kind": "Term",
+      "$Type": "Common.QualifiedName",
+      "$AppliesTo": ["EntityType"],
+      "@Common.Experimental": true,
+      "@Core.Description": "Qualified name of a function that computes AI-based recommendations for entities of the annotated type (see [this template](#Template_RecommendationsFunction) and [this alternative](#Template_RecommendationsFunction2))"
+    },
+    "Template_RecommendationsFunction": [
+      {
+        "$Kind": "Function",
```
drktfl, line 1301: Probably discussed elsewhere already: Is this one really a Function? It creates some state on the server that can be read at a different (!) URL (the returned one)

vocabularies/Common.json
HeikoTheissen: It's not recursive:
```
> new URL("Items(10)",`http://server/service/SalesOrder('A')/ns.RecommendationsFunction?Targets=["","Items(10)"]`).href
"http://server/service/SalesOrder('A')/Items(10)"
```
The part after the `?` of the "relative-to-URL" is ignored.

vocabularies/Common.json
HeikoTheissen: But the state is not really persistent on the server. It is deleted after it has been read (or when it is older than one minute or so). And it can only be accessed by the returned URL, so it is invisible to everyone else.

ralfhandl: It doesn't change the "business" data of its defining service, which is good enough for me.

Most/all HTTP calls create state somewhere, for example in access logs.

drktfl: thanks! ✅

vocabularies/Common.json
drktfl: > It's not recursive:

I know it isn't. I just wondered whether there is a better name, but maybe there isn't. Let's keep it.

bernhardgr42:
Some ideas:
1. instead of call function ... starting new task use bgPF https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=3328971542
2. instead of doing the loop on server side you could to this on client side
3. Why is GET RetrieveRecommendations?ID=. not part of the $batch? Then you could to a sync call to Recommendations

HeikoTheissen:
1. Good idea
2. But that leads to more network traffic.
3. The most important aspect of the whole exercise is that recommendations must be retrieved _asynchronously_, because computing them takes so long.

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,132 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+        <Annotation Term="Core.Description" String="Template for functions that retrieve AI-based recommendations for the bound entity and its related entities" />
+        <Annotation Term="Core.LongDescription">
+          <String>The template function itself cannot be invoked.</String>
+        </Annotation>
+        <Parameter Name="ID" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="ID that occurs in the URL returned by a prior invocation of the [recommendations function](#Template_RecommendationsFunction)" />
+        </Parameter>
+        <ReturnType Type="Collection(Common.PropertyRecommendationType)" Nullable="false">
+          <Annotation Term="Core.Description" String="A collection of recommendations per targeted property" />
+        </ReturnType>
```
HeikoTheissen, line 1503: Instead of a table with generic return type

target|value|text|alternatives
---|---|---|---
`Items(10)/Product`|M10|Milk|...
`Items(20)/Product`|M11|Cheese|...

wouldn't the following return value be easier to consume for the client?

```
{"Items": [
  {"Item": "010", "Product": {
    "value": "M10", "text": "Milk", "alternatives": ...
  }},
  {"Item": "020", "Product": {
    "value": "M11", "text": "Cheese", "alternatives": ...
  }}
]}
```

The return type of the recommendations retrieval function would then be entity-specific, but could be generated by the SADL designtime. And the `"value"` could have the correct type (for example, `Edm.Date`), not just `Edm.String`.

vocabularies/Common.json
```diff
@@ -1288,6 +1288,138 @@
...
+          "$Collection": true,
+          "$Type": "Common.PropertyRecommendationType",
+          "@Core.Description": "See return type in [`Template_RecommendationsRetrievalFunction`](#Template_RecommendationsRetrievalFunction)"
+        }
+      }
+    ],
+    "PropertyRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "target": {
```
drktfl, line 1376: Should this probably be `Target` (capital `T`)? I think this would be more consistent with the rest of the vocabulary. Maybe the other properties of this type and `AlternativeRecommendationType`, too.

vocabularies/Common.xml
HeikoTheissen: To be discussed in Albatross meeting.

vocabularies/Common.json
HeikoTheissen: To be discussed in Albatross meeting.

HeikoTheissen:
Short term orchestration:
```mermaid
sequenceDiagram
  actor client as Fiori client
  box ABAP
    participant server as OData service
    participant task as New task
  end
  participant AI as BTP recommendations service
  client ->>+ server: POST $batch
  server ->> server: PATCH SalesOrder('A')
  server ->> server: GET SalesOrder('A')?$expand=... for side effects
  server -->>- client: $batch response
  client ->>+ server: GET SalesOrder('A')/ns.Recommendations?Targets=...
  server ->>+ AI: compute recommendations
  AI -->>- server: return recommendations
  server ->>+ task: call function ... starting new task
  server -->>- client: return recommendations
  task ->> task: write top recommendation to database
  deactivate task
```

uhlmannm:
> Short term orchestration:
> 
> ```mermaid
> sequenceDiagram
>   actor client as Fiori client
>   box ABAP
>     participant server as OData service
>     participant task as New task
>   end
>   participant AI as BTP recommendations service
>   client ->>+ server: POST $batch
>   server ->> server: PATCH SalesOrder('A')
>   server ->> server: GET SalesOrder('A')?$expand=... for side effects
>   server -->>- client: $batch response
>   client ->>+ server: GET SalesOrder('A')/ns.Recommendations?Targets=...
>   server ->>+ AI: compute recommendations
>   AI -->>- server: return recommendations
>   server ->>+ task: call function ... starting new task
>   server -->>- client: return recommendations
>   task ->> task: write top recommendation to database
>   deactivate task
> ```
>
The "write top recommendations to database" seems to happen after returning the recommendations in the response of the function call.

HeikoTheissen:
> The "write top recommendations to database" seems to happen after returning the recommendations in the response of the function call.

Good point, I have adapted my diagram accordingly. However, this makes it even slower.

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,132 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <Term Name="RecommendationsFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="EntityType">
```
HeikoTheissen, line 1457: This annotation is not needed for the short-term orchestration, because the OData function is called in an app-specific code extension.

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,132 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+      <ComplexType Name="AlternativeRecommendationType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="value" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Alternatively recommended value, converted to string" />
+        </Property>
+        <Property Name="score" Type="Edm.Decimal" Scale="variable" Nullable="true">
+          <Annotation Term="Core.Description" String="Confidence score of the alternatively recommended value" />
+        </Property>
+      </ComplexType>
+      <Term Name="RecommendationsRole" Type="Common.RecommendationsRoleType" Nullable="false" AppliesTo="Property Parameter">
```
HeikoTheissen, line 1556: Do we need this annotation in the short-term orchestration? Does the app-specific code extension evaluate it?

vocabularies/Common.xml
HeikoTheissen: No need for this change.

vocabularies/Common.json
HeikoTheissen: Similarity with messages - keep lower case.

vocabularies/Common.xml
HeikoTheissen: Needed.

ralfhandl (dismissed):
Please add better documentation as requested in my previous review.

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,132 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <Term Name="RecommendationsFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Qualified name of a function that computes AI-based recommendations for entities of the annotated type (see [this template](#Template_RecommendationsFunction) and [this alternative](#Template_RecommendationsFunction2))" />
```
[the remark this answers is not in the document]
HeikoTheissen: Short-term, we will have a function like the second alternative, but call it synchronously.
Mid-term, I prefer the first alternative, which makes the recommendations retrieval asynchronous.
Long-term, we can use `Prefer: respond-async` for _one_ request in a $batch, but only when Gateway supports that.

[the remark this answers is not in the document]
ralfhandl: I would skip the mid-term thingy because it is structurally different.

Just go for sync single bound function now and then evolve to async single bound function.

vocabularies/Common.xml
```diff
@@ -1557,18 +1557,29 @@ The template function itself cannot be invoked.</String>
...
+            <Record>
+              <PropertyValue Property="Value" Int="Output" />
+              <Annotation Term="Core.Description" String="The property or parameter is the target of some recommendation" />
+            </Record>
+            <Record>
+              <PropertyValue Property="Value" Int="InputOutput" />
+              <Annotation Term="Core.Description" String="The property is input for and the target of some recommendation" />
+            </Record>
+            <Record>
+              <PropertyValue Property="Value" Int="Required" />
```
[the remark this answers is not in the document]
ralfhandl: So we have RequiredInput, (Optional)Input, (Optional)Input(And)Output, Output(Only)?

vocabularies/Common.xml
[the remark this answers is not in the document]
HeikoTheissen: We might need the mid-term thingy for 2408, if Gateway cannot implement the `Prefer: respond-async` in a $batch soon enough.

vocabularies/Common.xml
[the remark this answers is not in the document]
HeikoTheissen: Yes, and @SNairVinod said that "RequiredInput" should be called "Required" only.

vocabularies/Common.xml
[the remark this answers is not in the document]
ralfhandl: The mid-term thingy is a detour with a different choreography. The cost of teaching this workaround to clients and then teaching them to _no longer use it_ should not be underestimated.

Better wait for the real async solution.

vocabularies/Common.xml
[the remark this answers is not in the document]
ralfhandl: Which already confused one reader...

HeikoTheissen:
Discussion with Guru:
* The `text` in the return structure cannot be determined by BTP, is this done by SADL or even the application?
* Is the `text` really required in the return structure? Can't it be determined by the UI model?
* Is `value` only filled if the confidence score is high enough?
* Short-term, the top recommendation can be saved in the sticky session, not in the database.

vocabularies/Common.xml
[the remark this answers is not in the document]
HeikoTheissen: I have removed any mention of the workaround from the vocabulary.

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,93 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+      <Term Name="RecommendationsFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Qualified name of a function that computes AI-based recommendations for entities of the annotated type (see [this template](#Template_RecommendationsFunction))" />
+      </Term>
+      <Function Name="Template_RecommendationsFunction" IsBound="true">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Template for functions that compute AI-based recommendations for the bound entity and its related entities" />
+        <Annotation Term="Core.LongDescription">
+          <String>If the function is invoked repeatedly in a stateful session, it SHOULD NOT
+recompute recommendations whose predictors have not changed since the previous invocation.
```
ralfhandl, line 1466: Is this an instruction for the implementors of a recommendation function, or is it the function's desired behavior to not return recommendations that haven't changed since the last stateful call?

ralfhandl:
Can we move these nice sequence diagrams into the vocabulary markdown or an accompanying markdown file in the `examples` folder, similar to what we did with https://github.com/oasis-tcs/odata-vocabularies/blob/main/examples/Org.OData.Core.V1.GeometryFeature-sample.md?

vocabularies/Common.xml
HeikoTheissen: Both. But since this is an implementation detail, I'll remove the sentence.

ralfhandl:
Please add a link from the term definition to the new markdown file.

HeikoTheissen:
> Please add a link from the term definition to the new markdown file.

Done, but note that mermaid diagrams are not rendered on github.io.

ralfhandl:
> Done, but note that mermaid diagrams are not rendered on github.io.

We probably can make that work, see 
* https://github.com/ralfhandl/mermaid-diagrams
* https://ralfhandl.github.io/mermaid-diagrams/

vocabularies/Common.xml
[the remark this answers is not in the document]
SNairVinod: It was implicit that it is an input, so thought it doesn't really add any value. As this feedback is coming from second person on this commit, I don't mind making this as RequiredInput as well

docs/recommendations.md
```diff
@@ -0,0 +1,64 @@
...
+    server ->> server: GET SalesOrder('A')?$expand=... for side effects
+  and
+    server ->> server: GET SalesOrder('A')/ns.Recommendations?Targets=...<br>Prefer: respond-async
+    server ->>+ task: call function ... starting new task
+  end
+  server -->>- client: $batch response with<br>Location header for recommendations
+  task ->>+ AI: compute recommendations
+  client ->>+ server: GET «Location header for recommendations»
+  activate client
+  note right of client: outside sticky session
```
uhlmannm: It is only necessary to be outside of the session if the call waits for the results. On the client this would, however, mean that we introduce some special mechanism to do this polling call outside of the session. This would not be very nice.

HeikoTheissen: But doesn't repeated calling into the session make things slower overall, even if each individual request is rather short?

vocabularies/Common.json
```diff
@@ -1289,6 +1289,88 @@
...
+        "$Kind": "Function",
+        "$IsBound": true,
+        "@Common.Experimental": true,
+        "@Core.Description": "Template for functions that compute AI-based recommendations for the bound entity and its related entities",
+        "@Core.LongDescription": "The function SHOULD compute the recommendations asynchronously, see [this diagram](../docs/recommendations.md).\n\nThe template function itself cannot be invoked.",
+        "$Parameter": [
+          { "$Name": "Entity", "$Type": "Edm.EntityType" },
+          {
+            "$Name": "Targets",
+            "$Collection": true,
```
SNairVinod, line 1311: @HeikoTheissen, I guess we discussed in the past that complex types can't be passed as input parameters to OData Function. You recommended in that case to stringify and sent it. Today while looking at the PR we realized that we still have a collection here and RAP can't support this. Can we change this?

HeikoTheissen: Change it to what? We need a collection of target URLs, how else to represent this than as a collection of strings? A whitespace-separated list of strings?

ralfhandl: > complex types can't be passed as input parameters to OData Function

Which SAP tool introduces this restriction?

HeikoTheissen: SADL, see https://jira.tools.sap/browse/FIORITECHP1-22187.

HeikoTheissen: Looking into it...

docs/recommendations.md
uhlmannm: Beside the time the requests block the session? I hope not.

vocabularies/Common.json
```diff
@@ -1289,6 +1289,95 @@
...
+        "@Core.Description": "URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved",
+        "@Core.LongDescription": "The request `SalesOrder('A')/ns.GetRecommendations?Targets={\"URLs\":[\"\",\"Items(10)\"]}`\n            retrieves recommendations for the entities `SalesOrder('A')` and `SalesOrder('A')/Items(10)`."
+      }
+    },
+    "PropertyRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "AIRecmddFldURL": {
+        "@Core.IsURL": true,
+        "@Core.Description": "URL (relative to the URL of the [`GetRecommendations`](#GetRecommendations) function request) addressing a property [OData-URL, section 4.6] that is targeted by this recommendation",
```
SNairVinod, line 1333: Not sure how does the relative Url work in the following cases

- Editing an instance which has items, when the list is loaded there is one function call made to retrieve the recommendation & pass all the items in the UI as targets, for an example in a request like below

```js
SalesOrder('A')/ns.GetRecommendations?Targets={\"URLs\":[\"\",\"Items(10)\",\"Items(20)\",\"Items(30)\"]}
```

We will expect the results for SalesOrder A, Item(10), Item(20), Item(30) to return. So what is the expectation here

``` json
[
        {
            "target": "/CustomerPaymentTerms",
            "value": "",
            "text": "Cash only",
            "alternatives": [
                {
                    "value": "",
                    "score": 98
                },
                {
                    "value": "0009",
                    "score": 85
                }
            ]
        },
        {
            "target": "/_ShipToParty/BusinessPartner",
            "value": "69406910",
            "text": "Domestic NZ Supplier 1",
            "alternatives": [
                {
                    "value": "69406910",
                    "score": 99
                },
                {
                    "value": "45404510",
                    "score": 90
                },
                {
                    "value": "62100010",
                    "score": 85
                }
            ]
        },
        {
            "target": "/_Item(10)/Material",
            "value": "RMPI002",
            "text": "Input 1 AGENT 2",
            "alternatives": [
                {
                    "value": "RMPI002",
                    "score": 99
                },
                {
                    "value": "RMPI004",
                    "score": 90
                },
                {
                    "value": "RMPI003",
                    "score": 85
                }
            ]
        },
        {
            "target": "/_Item(20)/Material",
            "value": "RMPI002",
            "text": "Input 1 AGENT 2",
            "alternatives": [
                {
                    "value": "RMPI002",
                    "score": 99
                },
                {
                    "value": "RMPI004",
                    "score": 90
                },
                {
                    "value": "RMPI003",
                    "score": 85
                }
            ]
        }
    ]
```
_**Note: Here Item(30) doesn't have any recommendation. CustomerPaymentTerms is coming from Header node, BusinessPartner coming from 1..1 association _ShipToParty & Material comes from 1..n association _Item**_

Of course the names represent initial proposal and show updated names. @HeikoTheissen, Can you confirm?

vocabularies/Common.json
```diff
@@ -1289,6 +1289,95 @@
...
+        "$Collection": true,
+        "@Core.IsURL": true,
+        "@Core.Description": "URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved",
+        "@Core.LongDescription": "The request `SalesOrder('A')/ns.GetRecommendations?Targets={\"URLs\":[\"\",\"Items(10)\"]}`\n            retrieves recommendations for the entities `SalesOrder('A')` and `SalesOrder('A')/Items(10)`."
+      }
+    },
+    "PropertyRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "AIRecmddFldURL": {
```
SNairVinod, line 1331: Can we use non-abbreviated names proposed by Costa in the last mail like "AIRecommendedFieldURL"?

vocabularies/Common.json
HeikoTheissen: Costa's mail suggested a different format: `/SalesOrder('A')/_Item(20)/Material`.

I will now change the vocabulary to reflect the format in your example, although I don't understand the leading slash.

ralfhandl: Why don't we use the exact same format for `target` that we are using in warning and info messages for `target` and `additionalTargets`?

ralfhandl: And while we are at it, we could document this format here in addition to the internal wiki page where we currently hide it 😄

vocabularies/Common.xml
SNairVinod: Can we add this as well even in case application may not use short term?

HeikoTheissen: Why? We anyway plan to use a different annotation (`Core.ImplementsSignature`) long-term.

vocabularies/Common.json
SNairVinod: @HeikoTheissen, 

- Currently we have coded based on the absolute path. But we could make this to relative URL for consistency & lesser repetition. So relative URL is Ok
- Leading slash is a mistake in my sample data.  So it would be "CustomerPaymentTerms" and not "/CustomerPaymentTerms". It applies in all other cases as well
- @ralfhandl, relative Url is something which we use even in Messages. It is then consistent

docs/recommendations.md
```diff
@@ -0,0 +1,88 @@
+Invocation flow for a recommendations function annotated in [`Common.RecommendationsFunction`](../vocabularies/Common.md#RecommendationsFunction)
```
uhlmannm, line 1: This function is no longer part of the proposal.

docs/recommendations.md
```diff
@@ -0,0 +1,88 @@
...
+  and
+    server ->> server: GET SalesOrder('A')/ns.Recommendations?Targets=...<br>Prefer: respond-async
+    server ->>+ task: call function ... starting new task
+  end
+  server -->> client: $batch response with<br>Location header for recommendations
+  task ->>+ AI: compute recommendations
+  client ->>+ server: GET «Location header for recommendations»
+  note right of server: outside sticky session
+  par
+    loop until successful lookup
```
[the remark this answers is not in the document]
HeikoTheissen: > Will applications have a choice whether to use "long polling" or "repeated poll requests"? If so, it would be good for the client to know which is used (at least if a sticky session is used as well).

This document explains both methods. Which ones we support and how we inform the client is not part of this PR.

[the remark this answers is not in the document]
HeikoTheissen: > Will applications have a choice whether to use "long polling" or "repeated poll requests"?

The server could use long polling if called outside a sticky session (that is, without `sap-contextid` header) and short polling otherwise. Then it's up to the client whether to poll inside or outside the session.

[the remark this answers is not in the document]
uhlmannm: But these SAP specific additions will have to be defined somewhere. So essentially, we will need to see which way we will go in the productive implementation. Then there may be the need for some additional specification.

uhlmannm (approved):

uhlmannm (approved):

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+      <!--
+      <Term Name="RecommendationsFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Qualified name of a function that computes AI-based recommendations for entities of the annotated type (see [this template](#GetRecommendations))" />
+      </Term>
+      -->
+      <Function Name="GetRecommendations" IsBound="true">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.IsSignature" />
+        <Annotation Term="Core.Description" String="Computes AI-based recommendations for the bound entity and its related entities" />
```
ralfhandl, line 1466:
```suggestion
        <Annotation Term="Core.Description" String="Computes recommendations for the bound entity and its related entities" />
```
AI-based is cool, but do we want to restrict use of this term to special implementations?

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+          <Annotation Term="Core.Description" String="URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved" />
+          <Annotation Term="Core.LongDescription">
+            <String>The request `SalesOrder('A')/ns.GetRecommendations?Targets={"URLs":["","Items(10)"]}`
+            retrieves recommendations for the entities `SalesOrder('A')` and `SalesOrder('A')/Items(10)`.</String>
+          </Annotation>
+        </Property>
+      </ComplexType>
+      <ComplexType Name="PropertyRecommendationType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="AIRecommendedFieldPath" Type="Edm.String" Nullable="false">
```
ralfhandl, line 1498:
```suggestion
        <Property Name="RecommendedFieldPath" Type="Edm.String" Nullable="false">
```
I would keep it implementation-neutral

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+          <Annotation Term="Core.IsURL" />
+          <Annotation Term="Core.Description" String="URL (relative to the URL of the [`GetRecommendations`](#GetRecommendations) function request) addressing a property [OData-URL, section 4.6] that is targeted by this recommendation" />
```
ralfhandl, line 1500:
```suggestion
          <Annotation Term="Core.Description" String="URL (relative to the URL of the [`GetRecommendations`](#GetRecommendations) function request) addressing a property [OData-URL, section 4.6](https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_AddressingaProperty) that is targeted by this recommendation" />
```
Was this intended as a hyperlink?

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+          </Annotation>
+        </Property>
+        <Property Name="AIRecommendedFieldValue" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Recommended value, converted to string" />
+          <Annotation Term="Common.Text" Path="text" />
+        </Property>
+        <Property Name="AIRecommendedFieldDescription" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.Description" String="Description of the recommended value" />
+        </Property>
+        <Property Name="AIAltvRecmddFldVals" Type="Collection(Common.AlternativeRecommendationType)" Nullable="false">
```
ralfhandl, line 1513:
```suggestion
        <Property Name="RecommendedFieldValue" Type="Edm.String" Nullable="false">
          <Annotation Term="Core.Description" String="Recommended value, converted to string" />
          <Annotation Term="Common.Text" Path="text" />
        </Property>
        <Property Name="RecommendedFieldDescription" Type="Edm.String" Nullable="true">
          <Annotation Term="Core.Description" String="Description of the recommended value" />
        </Property>
        <Property Name="AlternativeValues" Type="Collection(Common.AlternativeRecommendationType)" Nullable="false">
```
Implementation-neutral, avoid cryptic abbreviations

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+            <String>If a value is recommended via property `AIRecommendedFieldValue`, it must be the first entry in this list.</String>
+          </Annotation>
+        </Property>
+      </ComplexType>
+      <ComplexType Name="AlternativeRecommendationType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="AIRecommendedFieldValue" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Alternatively recommended value, converted to string" />
+        </Property>
+        <Property Name="AIRecommendedFieldScore" Type="Edm.Decimal" Scale="variable" Nullable="true">
```
ralfhandl, line 1525:
```suggestion
        <Property Name="RecommendedFieldValue" Type="Edm.String" Nullable="false">
          <Annotation Term="Core.Description" String="Alternatively recommended value, converted to string" />
        </Property>
        <Property Name="RecommendedFieldScore" Type="Edm.Decimal" Scale="variable" Nullable="true">
```

ralfhandl (changes requested):

vocabularies/Common.xml
HeikoTheissen: The names are aligned with the Global Field Name governance, they insisted on the AI prefix to distinguish them from existing names (for example, for product recommendations).

vocabularies/Common.xml
HeikoTheissen: Our AI ethics demand that we let people know whether they are talking to AI or not.

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <!--
+      <Term Name="RecommendationsFunction" Type="Common.QualifiedName" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Qualified name of a function that computes AI-based recommendations for entities of the annotated type (see [this template](#GetRecommendations))" />
+      </Term>
+      -->
+      <Function Name="GetRecommendations" IsBound="true">
```
ralfhandl, line 1463:
```suggestion
      <Function Name="AIRecommendations" IsBound="true">
```

ralfhandl: Then let's consistently prefix everything here with AI, and don't use HTTP verbs in function names.

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+        <Property Name="URLs" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.IsURL" />
+          <Annotation Term="Core.Description" String="URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved" />
+          <Annotation Term="Core.LongDescription">
+            <String>The request `SalesOrder('A')/ns.GetRecommendations?Targets={"URLs":["","Items(10)"]}`
+            retrieves recommendations for the entities `SalesOrder('A')` and `SalesOrder('A')/Items(10)`.</String>
+          </Annotation>
+        </Property>
+      </ComplexType>
+      <ComplexType Name="PropertyRecommendationType">
```
ralfhandl, line 1496:
```suggestion
      <ComplexType Name="PropertyAIRecommendationType">
```

vocabularies/Common.xml
ralfhandl, line 1513: Please avoid GFN abbreviations in OData property names

ralfhandl (changes requested):

vocabularies/Common.xml
ralfhandl: Or get a readable GF name

vocabularies/Common.xml
```diff
@@ -1454,6 +1454,105 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+          <Annotation Term="Core.Description" String="A collection of recommendations that differ in their `AIRecommendedFieldPath` property" />
+        </ReturnType>
+      </Function>
+      <ComplexType Name="RecommendationsTargetsType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="URLs" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.IsURL" />
+          <Annotation Term="Core.Description" String="URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved" />
+          <Annotation Term="Core.LongDescription">
+            <String>The request `SalesOrder('A')/ns.GetRecommendations?Targets={"URLs":["","Items(10)"]}`
```
ralfhandl, line 1491:
```suggestion
            <String>The request `SalesOrder('A')/ns.AIRecommendations?Targets={"URLs":["","Items(10)"]}`
```

vocabularies/Common.xml
HeikoTheissen: We now go into the other direction: Add the AI prefix everywhere.

vocabularies/AI.xml
ralfhandl, line 1: AI is too generic, and it doesn't convey that this vocabulary is tailored along the needs of a single (if very important) LoB

ralfhandl:
`AI` as a vocabulary name is too generic, and it doesn't convey that this vocabulary is tailored along the needs of a single (if very important) LoB:
- Do you plan to have other ABAP or S/4 terms in this new vocabulary?
- Or is it exclusively for S/4 AI Recommendations?

HeikoTheissen:
Do you suggest to name the vocabulary not after AI, but after S/4HANA (the LoB)?

vocabularies/AI.xml
```diff
@@ -0,0 +1,134 @@
...
+            <PropertyValue Property="href" String="https://sap.github.io/odata-vocabularies/vocabularies/AI.json" />
+          </Record>
+          <Record>
+            <PropertyValue Property="rel" String="describedby" />
+            <PropertyValue Property="href" String="https://github.com/sap/odata-vocabularies/blob/main/vocabularies/AI.md" />
+          </Record>
+        </Collection>
+      </Annotation>
+
+      <Function Name="GetRecommendations" IsBound="true">
```
ralfhandl, line 38:
```suggestion
      <Function Name="Recommendations" IsBound="true">
```

ralfhandl (changes requested):

ralfhandl (approved):

uhlmannm (dismissed):

vocabularies/AI.json
```diff
@@ -0,0 +1,117 @@
...
+      }
+    ],
+    "RecommendationTargetsType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "URLs": {
+        "$Collection": true,
+        "@Core.IsURL": true,
+        "@Core.Description": "URLs (relative to the request URL) that address entities, properties or action/function parameters for which recommendations shall be retrieved",
+        "@Core.LongDescription": "The request `SalesOrder('A')/AI.Recommendations?AIRecommendationTargets={\"URLs\":[\"\",\"Items(10)\"]}`\n            retrieves recommendations for the entities `SalesOrder('A')` and `SalesOrder('A')/Items(10)`."
```
SNairVinod, line 56: We have discussion and agreed to send the full path instead of the relative path. Can we please update the comment

vocabularies/AI.json
```diff
@@ -0,0 +1,117 @@
...
+      }
+    },
+    "PropertyRecommendationType": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "AIRecommendedFieldPath": {
+        "@Core.IsURL": true,
+        "@Core.Description": "URL (relative to the URL of the [`Recommendations`](#Recommendations) function request) addressing a property [OData-URL, section 4.6](https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_AddressingaProperty) that is targeted by this recommendation",
+        "@Core.LongDescription": "The request `SalesOrder('A')/AI.Recommendations?AIRecommendationTargets=...`\n            might retrieve a recommendation with `\"AIRecommendedFieldPath\": \"Items(10)/Product\"`."
+      },
```
SNairVinod, line 66: For consistency reason, can we keep this comment also to indicate that the backend needs to send the full path back to the client and not relative

vocabularies/AI.json
```diff
@@ -0,0 +1,117 @@
...
+        "$Kind": "Function",
+        "$IsBound": true,
+        "@Common.Experimental": true,
+        "@Core.IsSignature": true,
+        "@Core.Description": "Computes AI-based recommendations for the bound entity and its related entities",
+        "@Core.LongDescription": "The function SHOULD compute the recommendations asynchronously, see [this diagram](../docs/recommendations.md).",
+        "$Parameter": [
+          { "$Name": "Entity", "$Type": "Edm.EntityType" },
+          {
+            "$Name": "Targets",
```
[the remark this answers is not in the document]
HeikoTheissen: But this is a collection, shouldn't it then be `AIRecommendedFieldPaths`?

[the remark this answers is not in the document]
HeikoTheissen: We need _two_ global field names:
- one for the parameter (which has a complex type)
- one for the property in this complex type (which is a collection of URLs).

PR #279, opened 2023-08-30
Core.Example values have the wrong type

HeikoTheissen:
In the SAP vocabularies, `Core.Example` annotations have values that are not of the correct type `Core.ExampleValue`.

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -171,8 +171,8 @@ a collection of business object instances, e.g. as a list or table.</String>
       <Term Name="ConnectedFields" Type="UI.ConnectedFieldsType" Nullable="false" AppliesTo="EntityType">
         <Annotation Term="Core.Description" String="Group of semantically connected fields with a representation template and an optional label" />
         <Annotation Term="Core.Example">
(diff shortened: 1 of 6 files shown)


vocabularies/UI.xml
```diff
@@ -171,8 +171,8 @@ a collection of business object instances, e.g. as a list or table.</String>
       <Term Name="ConnectedFields" Type="UI.ConnectedFieldsType" Nullable="false" AppliesTo="EntityType">
         <Annotation Term="Core.Description" String="Group of semantically connected fields with a representation template and an optional label" />
         <Annotation Term="Core.Example">
-          <Record>
-            <Annotation Term="com.sap.vocabularies.UI.v1.ConnectedFields" Qualifier="Material">
+          <Record Type="Core.ComplexExampleValue">
```
ralfhandl, line 174: The example wants to show an _annotation_ with term `ConnectedFields`, which it did.

Please revert this change.

ralfhandl (changes requested):
Please close this PR without action, the examples show annotations and are correct.

PR #281, opened 2023-09-06
Require namespace-qualified names in string values

ralfhandl:


ThomasChadzelek (approved):

HeikoTheissen (commented):
I'd prefer if the UI5 model implements alias resolution with the help of a "hardcoded" list of term (property) names. Such a list can be built automatically (during UI5 build) with the help of the Node.js function from #280.

uhlmannm (approved):

PR #284, opened 2023-09-14
Added Source Event for Side Effects

martin4all:
Start discussion for Side Effects triggered via Events. Events are transported via WebSocket Connection (ABAP APC).

## Modelling in RAP
- Side effect events are model in behavior definitions
```
event NumberOfSeatsUpdated for side effects;
```

- A side effect can be attached to an side effect event
 ```
side effects
{
   event NumberOfSeatsUpdated affects field _Root.TotalSeats;
}
```

- Via web socket connection (APC) the side effect events are issued. E.g.

PCP Field | Value 
-- | --
serverAction | RaiseSideEffect
sideEffectEventName | NumberOfSeatsUpdated
sourceEntityKey1 | 0000001000
sourceEntityKey2 | 0000000030

## OData Annotations
The following information is needed
1. The name of the side effect event name
2.  The UI needs the source instance from which the side effect should be executed. This must be sent via web socket. A mapping between the entity key fields in backend and on OData side is needed
3. Channel name 

- OData Annotation with new PropertyValue _SourceEvents_

```
<Annotations Target="SAP__self.SaloonType">
  <Annotation Term="SAP__common.SideEffects" Qualifier="SAP__Event_NumberOfSeatsUpdated">
    <Record Type="SAP__common.SideEffectsType">
      <PropertyValue Property="SideEffectKeyBinding">
        <Collection>
          <Record>
            <PropertyValue Property="PCPFieldName" String="sourceEntityKey1" />
            <PropertyValue Property="KeyPropertyName" PropertyPath="Cinema" />
          </Record>
          <Record>
            <PropertyValue Property="PCPFieldName" String="sourceEntityKey2" />
            <PropertyValue Property="KeyPropertyName" PropertyPath="Saloon" />
          </Record>
        </Collection>
      </PropertyValue>
      <PropertyValue Property="SideEffectEventNames">
        <Collection>
	  <String>NumberOfSeatsUpdated</String>
	</Collection>
      </PropertyValue>
      <PropertyValue Property="TargetProperties">
	<Collection>
	  <String>_Root/TotalSeats</String>
	</Collection>
      </PropertyValue>
    </Record>
  </Annotation>
```

## Open Questions
1. The UI has to connect to the right channel. The channel is given via URL parameter when opening the web socket connection. How to inform the UI about the URL parameter? There is already an annotation to get the web socket URL: `<Annotation Term="SAP__common.WebSocketBaseURL" String="/sap/bc/apc/sap/odata"/>`
### Example
````
/sap/bc/apc/sap/odata/?relatedService=/sap/opu/odata4/sap/b_apw_cinema_v4/srvd/sap/s_apw_cinema/0001/&application=cinema

### CODE REVIEW

**Comment by RaizoC** [comment]
Not sure we should add a constraint, that either source properties and entities XOR source events can be defined on source side.

**Comment by RaizoC** [comment]
Is it really an "event" that triggers the side effect from UI side or a "notification" (-> amc://notification/notify)?

**Comment by martin4all** [comment]
> Is it really an "event" that triggers the side effect from UI side or a "notification" (-> amc://notification/notify)?

We use the term "event" in backed (BDEF). Therefore it's named _SourceEvents_. To make it consistent we can also rename the AMC to amc://event/raise. But I'm also fine if we use notification ...

**Comment by MarcelWaechter** [comment]
Hi,
may I ask if we really need this separation of having events that triggers a side effect which then contains the target(s) to be refreshed... or if this is not like a server driven event that already contains the targets to be refreshed (not the data but the targets so we send GET requests same as we would do for side effects)
In this case the server could be even more precise as it knows what fields were changed so the client need to request them only.
Otherwise apps annotate a side effect with 5 targets that might be affected. Once the event is triggered the client then need to refresh all 5 properties.

**Comment by martin4all** [comment]
> Hi, may I ask if we really need this separation of having events that triggers a side effect which then contains the target(s) to be refreshed... or if this is not like a server driven event that already contains the targets to be refreshed (not the data but the targets so we send GET requests same as we would do for side effects) In this case the server could be even more precise as it knows what fields were changed so the client need to request them only. Otherwise apps annotate a side effect with 5 targets that might be affected. Once the event is triggered the client then need to refresh all 5 properties.

**Comment by HeikoTheissen** [comment]
Since you closed this, does it mean that we don't need the `WebSocketChannel` annotation anymore?

**Comment by martin4all** [comment]
> > Hi, may I ask if we really need this separation of having events that triggers a side effect which then contains the target(s) to be refreshed... or if this is not like a server driven event that already contains the targets to be refreshed (not the data but the targets so we send GET requests same as we would do for side effects) In this case the server could be even more precise as it knows what fields were changed so the client need to request them only. Otherwise apps annotate a side effect with 5 targets that might be affected. Once the event is triggered the client then need to refresh all 5 properties.

**Comment by martin4all** [comment]
There are 2 issues to my understanding:
1. Here we have the same issue as we have it for the key properties: The server does not speak OData which means it does not know entity sets, entity types, or navigation properties. This is only known when creating the $metadata of the service.
2. The servicer does not really know which data has been changed. We have to rely on the correct implementation of the business object that the event is raised correctly. (But for non-event driven side effects we already have possibility that a change of one fields can affect several other fields.)

**Comment by martin4all** [comment]
> Since you closed this, does it mean that we don't need the `WebSocketChannel` annotation anymore?
Sorry, closed by accident ...

**Review by HeikoTheissen — APPROVED** [review]

**Review by MarcelWaechter — APPROVED** [review]
the initial description of the PR isn't up-to-date anymore, still contains the key binding etc. you might also want to update this?
But anyhow, looks good to me, but for sure good if Martin/Renzo also have a look

Thanks

**Comment by ralfhandl** [comment]
How about "rescuing" the still relevant parts of the PR description into a Markdown file in the `examples` folder similar to the document [Annotating capabilities](https://github.com/oasis-tcs/odata-vocabularies/blob/main/examples/Org.OData.Capabilities.V1.capabilities.md)?

Side effects aren't the most straight-forward topic and deserve a little more documentation than the pure term description.

**Review by ralfhandl — CHANGES_REQUESTED** [review]
Please add more documentation

**Comment by MarcelWaechter** [comment]
Hi,
our UX came up with the idea to have different interaction types:
- notification: the User is just notified, the data is refreshed immediately (our default)
- confirmation: the User will be asked if data shall be refreshed or not
- none: the data is refreshed without any kind of interaction

My initial idea was that this shall be an app configuration (manifest setting), but I was asked if this can't be also part of the annotation.
What is your thought on this?

**Comment by HeikoTheissen** [comment]
If different SourceEvents can have different interaction types, I would put this into the annotation.

**Comment by MarcelWaechter** [comment]
yes, every event event could have a different interaction type

**Comment by RaizoC** [comment]
> yes, every event event could have a different interaction type

having it as annotation might be ok, but not sure it would be correct in the side effects annotation as how the UI reacts on one and the same side effect depends on the UI, so maybe this has to be added rather to the UI-annotations domain (especially as I doubt that the setting is specific for different side effects, but a global setting for the UI overall)

**Comment by MarcelWaechter** [comment]
> > yes, every event event could have a different interaction type
> 
> having it as annotation might be ok, but not sure it would be correct in the side effects annotation as how the UI reacts on one and the same side effect depends on the UI, so maybe this has to be added rather to the UI-annotations domain (especially as I doubt that the setting is specific for different side effects, but a global setting for the UI overall)

Yes, I agree they would not make so much sense in the Common but only in the UI vocabulary, and then I'm not sure on how we could have an understandable annotation with a connection to the side effects events. 
I personally think apps will anyway use only the default (notification).
 
I'd therefore start with a manifest setting, we could still provide it later if really needed

File: vocabularies/Common.xml [hunk 1]
```diff
@@ -1475,6 +1479,23 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Channel for WebSocket connections" />
+        <Annotation Term="Core.LongDescription">
+          <String>Messages sent over the channel follow a protocol whose documentation is listed below per qualifier of this annotation.
+          To consume a channel, the client opens a web socket connection at the [`WebSocketBaseURL`](#WebSocketBaseURL)
+          followed by a URL parameter `«annotation qualifier»=«annotation value»`.
+          If the annotation value is null, the parameter value follows from the protocol documentation.
+          &lt;dl&gt;Supported qualifiers and protocols:
+          &lt;dt&gt;`sideEffects` &lt;dd&gt;notifications about side effects to be triggered by the client (documentation)
+          &lt;dt&gt;`draft` &lt;dd&gt;collaborative draft; the parameter value is explained in the documentation
```
ralfhandl, line 1496: Which documentation is meant here? Please add links to the documentation.

vocabularies/Common.xml
```diff
@@ -1508,14 +1511,15 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
         <Annotation Term="Core.LongDescription">
-          <String>Messages sent over the channel follow a protocol whose documentation is listed below per qualifier of this annotation.
-          To consume a channel, the client opens a web socket connection at the [`WebSocketBaseURL`](#WebSocketBaseURL)
-          followed by a URL parameter `«annotation qualifier»=«annotation value»`.
-          If the annotation value is null, the parameter value follows from the protocol documentation.
-          &lt;dl&gt;Supported qualifiers and protocols:
-          &lt;dt&gt;`sideEffects` &lt;dd&gt;notifications about side effects to be triggered by the client (documentation)
-          &lt;dt&gt;`draft` &lt;dd&gt;collaborative draft; the parameter value is explained in the documentation
-          &lt;/dl&gt;</String>
+          <String>Messages sent over the channel follow the ABAP Push Channel protocol.
```
ralfhandl, line 1514:
```suggestion
          <String>Messages sent over the channel follow the [ABAP Push Channel protocol](https://community.sap.com/t5/application-development-blog-posts/specification-of-the-push-channel-protocol-pcp/ba-p/13137541).
```

ralfhandl (changes requested):
+1, please add link to ABAP Push Channel Protocol

HeikoTheissen (approved):

ralfhandl (approved):

PR #287, opened 2023-09-27
Hierarchy maintenance actions

HeikoTheissen:


HeikoTheissen (commented):
The `ChangeNextSibling` action has a nullable parameter `NextSibling`. Can a null value actually be passed in RAP?

PR #288, opened 2023-10-04
Alternative #2

HeikoTheissen:


ralfhandl:
Better move the AI Recommendations out of Common as they seem rather LoB-specific and not "common".

PR #290, opened 2023-10-24
Re-introduce preorder rank

HeikoTheissen:


vocabularies/Hierarchy.xml
```diff
@@ -120,6 +120,13 @@ the following collections of hierarchy nodes are distinguished:
             <String>This equals the number of ancestors in the sub-hierarchy, if the request involves a sub-hierarchy.</String>
           </Annotation>
         </Property>
+        <Property Name="PreorderRank" Type="Edm.Int64" Nullable="true">
+          <Annotation Term="Core.Description" String="Preorder rank of a node" />
+          <Annotation Term="Core.LongDescription">
+            <String>The preorder rank of a node is the index of the node in the sequence of nodes
+            created from preorder traversal of the unlimited hierarchy. The first node in preorder traversal has rank 0.</String>
```
ThomasChadzelek, line 127: I think that "limited hierarchy" would be more useful here. Not sure if you like to call it "LimitedPreorderRank" then ;-) From my perspective, we don't need both. But the preorder rank within the limited hierarchy is important because it corresponds to the $skip index you need when paging over that limited hierarchy. And that is the property which is helpful to us!

uhlmannm (approved):

ThomasChadzelek (approved):

uhlmannm (approved):

PR #293, opened 2023-11-10
Auto-expand transformation

HeikoTheissen:


vocabularies/Analytics.json
```diff
@@ -161,6 +161,102 @@
...
+            "@Core.OptionalParameter": {}
+          },
+          {
+            "$Name": "ResultEntriesCount",
+            "$Type": "Edm.Bool",
+            "@Core.Description": "Whether to return the total number of entries in the result independent of Skip/Top",
+            "@Core.OptionalParameter": { "DefaultValue": "false" }
+          }
+        ],
+        "$ReturnType": { "$Collection": true, "$Type": "Edm.EntityType" }
```
uhlmannm, line 234: I guess the EntityType is not enough. We will need some of the information that we also use in a hierarchical query. Most importantly, the [LimitedDescendantCount](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L109:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22) but also the [DrillState](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L112:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22) and the [DistanceFromRoot](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L121:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22).

vocabularies/Analytics.json
```diff
@@ -161,6 +161,102 @@
...
+      }
+    ],
+    "AutoExpandLevel": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "P": { "$Collection": true, "@Core.Description": "A non-empty set of property names constituting a level" }
```
uhlmannm, line 240: Does the description fit? We would need the the information of `group` and `groupLevels` as described in [sap.ui.model.odata.v4.ODataListBinding#setAggregation](https://ui5.sap.com/#/api/sap.ui.model.odata.v4.ODataListBinding%23methods/setAggregation) with `group` denoting all properties used for grouping the result data on leaf level and `groupLevels` denoting the properties used for visual grouping.

vocabularies/Analytics.json
HeikoTheissen: So we need an instance annotation whose type contains a subset of the properties of `Hierarchy.RecursiveHierarchyType`. Which subset exactly?

vocabularies/Analytics.json
HeikoTheissen: `Levels` is a collection of collections (technically, a collection of records with one collection-valued property). The set of grouping properties for the leaf level is the union of all nested collections, and the set of properties for visual grouping is the first nested collection. But you also need the grouping properties for the intermediate levels, which shall be auto-expanded.

vocabularies/Analytics.json
uhlmannm: @ThomasChadzelek I had found the [LimitedDescendantCount](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L109:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22), the [DrillState](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L112:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22) and the [DistanceFromRoot](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L121:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22). Is that it? Or already to much?

vocabularies/Analytics.json
uhlmannm: So I have something like `[['Country','Region'],['Segment']]` meaning that the leaf level is grouped by `Country`, `Region` and `Segment` and that there are two levels in the visual grouping, first the `Country` and then the `Region`. Is that understanding correct?

HeikoTheissen: No. Leaves are grouped by `Country`, `Region` and `Segment`. There is one more level grouped by `Country` and `Region` (this is called "visual grouping", right?).

vocabularies/Analytics.json
HeikoTheissen: Is it sufficient if these instance annotations are returned by the auto-expand function? If one node is expanded, they must be computed by the OData model itself.

vocabularies/Analytics.json
uhlmannm: To achieve what I want, I would need `[['Country'],['Region'],['Segment']]`?
The flexibility that [sap.ui.model.odata.v4.ODataListBinding#setAggregation](https://ui5.sap.com/#/api/sap.ui.model.odata.v4.ODataListBinding%23methods/setAggregation) provides with `group` and `groupLevels` is required for this function as well. I.e. we need to be able to specify an arbitrary number of groupable properties that are used for visual grouping. And the number of (properties in) `groupLevels` determines the number of visual groups. And then there can be an arbitrary number of additional properties in `group` that would be used additionally for the leaf level.

vocabularies/Analytics.json
```diff
@@ -161,6 +164,114 @@
...
+            "@Core.LongDescription": "Each element in the collection defines the property names for one level.\n            A property must not referenced by more than one level.\n            The first element in the collection defines the property names of the coarsest level,\n            the following elements define the property names of consecutively finer-grained aggregation levels.\n            All referenced properties must be groupable."
+          },
+          {
+            "$Name": "ExpandLevel",
+            "$Type": "Edm.Int64",
+            "@Core.Description": "Number of levels to be expanded, counting from left",
+            "@Core.LongDescription": "The function result comprises the leveled hierarchy with the expanded levels and the following level as leaves in preorder."
+          },
+          {
+            "$Name": "Aggregation",
```
uhlmannm, line 197: [sap.ui.model.odata.v4.ODataListBinding#setAggregation](https://ui5.sap.com/#/api/sap.ui.model.odata.v4.ODataListBinding%23methods/setAggregation) foresees to specify the `unit` with the related unit or currency property for a property in `aggregate`. Is this required here too or would the backend know that related unit/currency property and to the right thing automatically (depending on whether the unit/currency is used for grouping or not)?

vocabularies/Analytics.json
uhlmannm: I think yes. In the other cases, the model is creating requests to manually expand the visual groups. The backend would not even be aware that we are creating a visual grouping on the screen.

vocabularies/Analytics.json
HeikoTheissen: If the unit is a custom aggregate (which produces `null` if there is no unique value), then it must be passed in the `Aggregation` parameter. If it is used for grouping, it must be passed in the `Levels` parameter. How could the backend know which alternative the client desires?

vocabularies/Analytics.json
ThomasChadzelek: > @ThomasChadzelek I had found the [LimitedDescendantCount](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L109:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22), the [DrillState](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L112:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22) and the [DistanceFromRoot](https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/Hierarchy.xml#L121:~:text=%3CComplexType%20Name=%22-,RecursiveHierarchyType,-%22). Is that it? Or already too much?

What about LimitedRank? Our new favorite ;-)

DistanceFromRoot might be redundant because the hierarchy is leveled, not recursive?

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,114 @@
...
+            <String>Each element in the collection defines the property names for one level.
+            A property must not referenced by more than one level.
+            The first element in the collection defines the property names of the coarsest level,
+            the following elements define the property names of consecutively finer-grained aggregation levels.
+            The function result comprises the leveled hierarchy with these levels in preorder.
+            All referenced properties must be groupable.</String>
+          </Annotation>
+        </Parameter>
+        <Parameter Name="LeafLevel" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.Description" String="A possibly empty set of property names that constitute, together with the property names from `Levels`, the leaf level of the leveled hierarchy" />
```
ThomasChadzelek, line 201: What does "property names from `Levels`" mean exactly? To me, Levels looks like an array of objects which contain an array of strings, right?

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,114 @@
...
+          <Annotation Term="Core.Description" String="Properties to aggregate for all result entries on all levels" />
+          <Annotation Term="Core.LongDescription">
+            <String>All properties in this collection must be custom aggregates.</String>
+          </Annotation>
+        </Parameter>
+        <Parameter Name="SiblingOrder" Type="Collection(Analytics.AutoExpandSiblingOrder)" Nullable="false">
+          <Annotation Term="Core.Description" String="Sort specification to apply to all direct descendants of a given entry in the resulting leveled hierarchy" />
+        </Parameter>
+        <Parameter Name="BeforeAggregationFilter" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Expression valid for `filter` transformation to restrict the input set before any further procressing" />
```
ThomasChadzelek, line 217: Can't you achieve the same thing with a filter() transformation beforehand, in the same $apply? For TopLevels, we do likewise with orderby() or search() or filter().

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,114 @@
...
+            <Record/>
+          </Annotation>
+        </Parameter>
+        <Parameter Name="AggregatedValuesLeafFilter" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Expression valid for `filter` transformation to restrict the input set at the most detailed grouping level with conditions on aggregated values" />
+          <Annotation Term="Core.OptionalParameter">
+            <Record/>
+          </Annotation>
+        </Parameter>
+        <Parameter Name="Skip" Type="Edm.Int64" Nullable="false">
```
ThomasChadzelek, line 228: What is the difference compared to $skip, $top, $count? Can't they be used on the result of this function?

vocabularies/Hierarchy.xml
```diff
@@ -124,6 +142,12 @@ the following collections of hierarchy nodes are distinguished:
             <String>This equals the number of ancestors in the sub-hierarchy, if the request involves a sub-hierarchy.</String>
           </Annotation>
         </Property>
+        <Property Name="ChildCount" Type="Edm.Int64" Nullable="true">
```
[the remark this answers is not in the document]
HeikoTheissen: The three are inherited by `RecursiveHierarchyType` from `HierarchyType`, but with a changed definition, that's why they are repeated here. I moved them to the top (moving `ChildCount` and `DescendantCount` down).

vocabularies/Analytics.xml
HeikoTheissen: Yes, but then the backend would have to take all these transformations together and process them as one "auto-expand" transformation. (At least with the Analytical Engine, the filter is part of the auto-expand operation.)

If the backend agrees, we can change this.

vocabularies/Analytics.xml
HeikoTheissen: See https://github.com/SAP/odata-vocabularies/pull/293#discussion_r1475903114

vocabularies/Hierarchy.xml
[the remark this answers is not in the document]
ThomasChadzelek: OK, I see. Thanks!

vocabularies/Analytics.xml
ThomasChadzelek: If the backend disagrees, do we need support for search as well? Or is that too fuzzy for analytics?

vocabularies/Analytics.xml
ThomasChadzelek: OK, I see. Thanks!

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,114 @@
...
+            </Record>
+            <Record>
+              <PropertyValue Property="Value" String="desc" />
+              <Annotation Term="Core.Description" String="Sort in descending order" />
+            </Record>
+          </Collection>
+        </Annotation>
+      </TypeDefinition>
+
+      <Term Name="LevelInformation" Type="Hierarchy.HierarchyType" Nullable="false" AppliesTo="EntityType">
```
HeikoTheissen, line 285: Needs additional property for `ResultEntriesCount`

HeikoTheissen:
Open issue: How to instruct the transformation to create one row for an expanded group header above its descendants and another row for the subtotals below its descendants.

vocabularies/Analytics.json
```diff
@@ -161,6 +164,102 @@
...
       }
+    },
+    "AutoExpand": [
+      {
+        "$Kind": "Function",
+        "$EntitySetPath": "InputSet",
+        "$IsBound": true,
+        "@Common.Experimental": true,
+        "@Core.Description": "`$apply` transformation that expands an unnamed leveled hierarchy with custom aggregation of certain properties",
+        "@Core.LongDescription": "Example transformation sequence:\n```\n$apply=filter(Industry eq 'IT')\n/groupby((Country,CountryName,Region,RegionName),\n         filter($these/aggregate(Amount) gt 0 and\n                $these/aggregate(Currency) ne null))\n/concat(\n  Analytics.AutoExpand(\n    Levels=[{\"P\":[\"Country\",\"CountryName\"]},\n            {\"P\":[\"Region\",\"RegionName\"]}],\n    LeafLevel=[\"Segment\",\"Industry\"],\n    Aggregation=[\"Amount\",\"Currency\"],\n    SiblingOrder=[{\"Property\":\"Amount\",\n                   \"Order\":\"desc\"}],\n    ExpandLevels=[{\"Entry\":[\"US\",\"USA\"],\"Levels\":0},\n                  {\"Entry\":[\"DE\",\"Germany\",\"BW\",\"Baden-Württemberg\"],\"Levels\":2}])\n    /concat(aggregate($count) as ResultEntriesCount,\n            skip(20)/top(10)),\n  aggregate(Amount,Currency))\n```\n",
```
uhlmannm, line 175: In UI5 we start with the leaves count that is missing in the example. It would be a term `groupby((Country,CountryName,Region,RegionName))/aggregate($count as LeavesCount)`  in the outer concat followed by the grand total term `aggregate(Amount,Currency)` that is currently last in the example. The third term in the outer concat would be `Analytics.AutoExpand(...)/concat(aggregate($count) as ResultEntriesCount,skip(20)/top(10))`.

vocabularies/Analytics.json
```diff
@@ -161,6 +164,102 @@
...
+            "$Name": "Levels",
+            "$Collection": true,
+            "$Type": "Analytics.AutoExpandLevel",
+            "@Core.Description": "Collection of aggregation levels forming the visible part of a leveled hierarchy",
+            "@Core.LongDescription": "Each element in the collection defines the properties that constitute one level.\n            A property must not referenced by more than one level.\n            The first element in the collection defines the property names of the coarsest level,\n            the following elements define the property names of consecutively finer-grained aggregation levels.\n            The function result comprises the leveled hierarchy with these levels in preorder.\n            All referenced properties must be groupable."
+          },
+          {
+            "$Name": "LeafLevel",
+            "$Collection": true,
+            "@Core.Description": "A possibly empty set of property names that constitute, together with the property names from `Levels`, the leaf level of the leveled hierarchy",
```
uhlmannm, line 193: > A possibly empty set of property names

I wonder how that would work. If `LeafLevel` is empty then the last level in `Levels` is not expanded, cannot be expanded and could as well be written into `LeafLevel`?

HeikoTheissen: Isn't `LeafLevel` simply the last entry in `Levels`, then?

uhlmannm: No. There can be several properties used for grouping on the leaf level that are not used for visual grouping / group headers.
The interesting part is that we have three things:
1. properties used for visual grouping
2. properties used additionally for grouping on the leaf level
3. the number of visual groups initially expanded.

I think we had decided somewhen in the past that 3. is not necessary for the AutoExpand function as one can simply as collapsed visual groups to 2. But I wonder whether that decision is still good when introducing ExpandLevels.

HeikoTheissen: The leaf level is just the deepest grouping level, therefore we only need the parameter `Levels`, not `LeafLevel`.

In the current definition, nodes not mentioned in `ExpandLevels` are completely expanded. If you want the default to be "expand N levels", we need an additional parameter for that N.

uhlmannm: Yes, we can do it like that. But the question is how to denote the leaf level. Is it always the last of the Levels? And hence this last level is not and cannot be expanded?

HeikoTheissen: Yes.

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,123 @@
...
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="`$apply` transformation that expands an unnamed leveled hierarchy with custom aggregation of certain properties" />
+        <Annotation Term="Core.LongDescription">
+          <String>Example transformation sequence:
+filter on columns `Industry`, `Amount` and `Currency`,
+order by `Amount` descending,
+expand 2 levels, with two exceptions
+```
+$apply=filter(Industry in ['IT','AI'])
+/groupby((Country,CountryName,Region,RegionName,Segment,Industry),
```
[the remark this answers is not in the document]
uhlmannm: Here it should not matter whether these properties are added to groupby or not. It seems to me a bit more elegant to leave them out here.

vocabularies/Analytics.json
uhlmannm: Should it be stated explicitly in the Levels description?

HeikoTheissen: For me that is a tautology: The node on the deepest level cannot be expanded.

uhlmannm: If you look at the FE examples, then on the leaf level there are individual (unaggregated) records. If that is the target state, one could also argue that the last level can still be expanded and below this last level are the unaggregated records.

HeikoTheissen: https://github.com/SAP/odata-vocabularies/blob/8c7b80bf767e9b13b706b1aeb24029aae39e9ece/vocabularies/Analytics.xml#L224

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,109 @@
...
+            <Record/>
+          </Annotation>
+        </Parameter>
+        <ReturnType Type="Collection(Edm.EntityType)">
+          <Annotation Term="Core.Description" String="Output set including the instance annotation [`LevelInformation`](#LevelInformation)" />
+        </ReturnType>
+      </Function>
+      <ComplexType Name="AutoExpandLevel">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="P" Type="Collection(Edm.String)" Nullable="false">
```
HeikoTheissen, line 255: Have property `D` for dimensions and `A` for additional (dependent) properties.

vocabularies/Analytics.json
```diff
@@ -211,6 +211,13 @@
             "$Type": "Analytics.AutoExpandEntry",
             "@Core.Description": "Entries with exceptional expansion",
             "@Core.OptionalParameter": {}
+          },
+          {
+            "$Name": "SubtotalsAtBottom",
+            "$Type": "Edm.Bool",
```
uhlmannm, line 217: @simlin Are two value sufficient? Or could we have three (on top, at bottom, both)?

vocabularies/Analytics.json
```diff
@@ -211,6 +211,13 @@
...
+            "@Core.Description": "Whether to have expanded subtotal rows duplicated: before and after their descendants",
```
uhlmannm, line 218: Whether to duplicate the group headers so that they are included before and after their descendants

ralfhandl:
Why "auto" if it is part of the request? The annotation of the same name tells what the server will do without needing input from the client.

HeikoTheissen:
> Why "auto" if it is part of the request? The annotation of the same name tells what the server will do without needing input from the client.

"Automatic expansion" as opposed to "interactive expansion" where the user must click the expand button on every group header.

ralfhandl:
OData is not about user interaction with a UI, it is about interaction of the UI/client with the server, so "auto" feels wrong here, especially because the word "AutoExpand" has a different meaning in the Core vocabulary.

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,129 @@
         </Annotation>
       </Annotations>
 
+      <Function Name="AutoExpand" IsBound="true" EntitySetPath="InputSet">
```
HeikoTheissen, line 182: Suggested renaming:
```suggestion
      <Function Name="MultiLevelExpand" IsBound="true" EntitySetPath="InputSet">
```
Or simply "Expand".

vocabularies/Analytics.json
```diff
@@ -161,6 +164,114 @@
...
+          "@Core.Description": "Output set including the instance annotation [`LevelInformation`](#LevelInformation)"
+        }
+      }
+    ],
+    "MultiLevelExpandLevel": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "@Core.Description": "Property names constituting a level in an [unnamed leveled hierarchy](#MultiLevelExpand)",
+      "@Core.LongDescription": "Properties in `D` must be used to identify entries in [`ExpandEntries/Entry`](#MultiLevelExpandEntry),\n          otherwise they have the same effect as properties in `A`.",
+      "D": {
```
ralfhandl, line 235: "D" and "A" are unusually short names. Why not "DimensionProperties" and "AdditionalDimensionProperties"?

HeikoTheissen: They are so technical that I do not think longer names make anything clearer. And the properties in A must also appear in D, therefore "AdditionalDimensionProperties" would be misleading.

ralfhandl (dismissed):

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,129 @@
...
+$apply=filter(Industry in ('IT','AI'))
+/groupby((Country,Region,Segment,Industry),
+         filter($these/aggregate(Amount) gt 0 and
+                $these/aggregate(Currency) ne null))
+/concat(
+  groupby((Country,Region,Segment,Industry))
+    /aggregate($count as LeavesCount),
+  aggregate(Amount,Currency),
+  Analytics.MultiLevelExpand(
+    Levels=[{"DimensionProperties":["Country"],"AdditionalProperties":["CountryName"]},
```
ThomasChadzelek, line 200: This looks very chatty to me, but never mind.

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,129 @@
...
+</String>
+        </Annotation>
+        <Parameter Name="InputSet" Type="Collection(Edm.EntityType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Entity set to be processed" />
+        </Parameter>
+        <Parameter Name="Levels" Type="Collection(Analytics.MultiLevelExpandLevel)" Nullable="false">
+          <Annotation Term="Core.Description" String="Collection of aggregation levels forming a leveled hierarchy" />
+          <Annotation Term="Core.LongDescription">
+            <String>Each element in the collection defines the properties that constitute one level.
+            A property must not referenced by more than one level.
```
ThomasChadzelek, line 220: "A property must not **be** referenced by more than one level."

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,129 @@
...
+        <Property Name="DimensionProperties" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.Description" String="A non-empty set of property names that define a combination of dimension values" />
+        </Property>
+        <Property Name="AdditionalProperties" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.Description" String="A possibly empty set of names of additional properties of the dimensions that occur in `DimensionProperties`" />
+        </Property>
+      </ComplexType>
+      <ComplexType Name="MultiLevelExpandSiblingOrder">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Sibling order in an [unnamed leveled hierarchy](#MultiLevelExpand)" />
```
ThomasChadzelek, line 281: Hmm, this is not at all specific to hierarchies, multi-level, or unnamed.

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,129 @@
...
+          <Annotation Term="Core.Description" String="Property by which to sort" />
+        </Property>
+        <Property Name="Descending" Type="Edm.Boolean" Nullable="true">
+          <Annotation Term="Core.Description" String="Sort direction, ascending if not specified otherwise" />
+        </Property>
+      </ComplexType>
+      <ComplexType Name="MultiLevelExpandEntry">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Expansion state of an entry in an [unnamed leveled hierarchy](#MultiLevelExpand)" />
+        <Property Name="Entry" Type="Collection(Edm.String)" Nullable="false">
```
ThomasChadzelek, line 292: This means that a value for a dimension is always of type Edm.String? There cannot be a boolean dimension etc.?

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,129 @@
...
+      </Function>
+      <ComplexType Name="MultiLevelExpandLevel">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Property names constituting a level in an [unnamed leveled hierarchy](#MultiLevelExpand)" />
+        <Annotation Term="Core.LongDescription">
+          <String>`DimensionProperties` must be used to identify entries in [`ExpandEntries/Entry`](#MultiLevelExpandEntry),
+          otherwise they have the same effect as `AdditionalProperties`.</String>
+        </Annotation>
+        <Property Name="DimensionProperties" Type="Collection(Edm.String)" Nullable="false">
+          <Annotation Term="Core.Description" String="A non-empty set of property names that define a combination of dimension values" />
```
ThomasChadzelek, line 273: A set is usually unordered, as opposed to a list. But when I look at "Entry" below, I assume order matters and it must be the same here and there, right?

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,129 @@
...
+          <Annotation Term="Core.Description" String="Properties to aggregate for all result entries on all levels" />
+          <Annotation Term="Core.LongDescription">
+            <String>All properties in this collection must be custom aggregates.</String>
+          </Annotation>
+        </Parameter>
+        <Parameter Name="SiblingOrder" Type="Collection(Analytics.MultiLevelExpandSiblingOrder)" Nullable="false">
+          <Annotation Term="Core.Description" String="Sort specification to apply to all direct descendants of a given entry in the resulting leveled hierarchy" />
+        </Parameter>
+        <Parameter Name="ShowLevels" Type="Edm.Int64" Nullable="false">
+          <Annotation Term="Core.Description" String="Number of levels to be shown in the initial expansion (absent means all levels)" />
```
ThomasChadzelek, line 238: Allowed values range from 0 to length of "Levels", I assume? Should we define this contract more explicit? Or would it be OK to provide even larger values?

vocabularies/Analytics.xml
HeikoTheissen: I don't understand. The term "sibling" is defined only in a hierarchy.

vocabularies/Analytics.xml
HeikoTheissen: OData does not allow `Collection(Edm.PrimitiveType)`, therefore all node IDs must be converted to strings.

ThomasChadzelek: Is there a guideline how this transformation needs to happen? Would we use the same string as inside a JSON body or a URI literal or...?

vocabularies/Analytics.xml
ThomasChadzelek: OK, "sibling". But apart from that, it would apply to any ordinary collection as well.

vocabularies/Analytics.xml
HeikoTheissen: > Is there a guideline how this transformation needs to happen?

By casting the node ID to a string, in the same way as in the parameter `Hierarchy.TopLevels/ExpandLevels/NodeID`. I think that's the same as the string inside a JSON body.

vocabularies/Analytics.xml
HeikoTheissen: Do what?

ThomasChadzelek: As I tried to say: let's treat this as "case closed", but I don't know how to tell that to this tool.

vocabularies/Analytics.xml
ThomasChadzelek: My understanding is that the node ID is always a string, but not each dimension would be a string. "casting" can mean many different things in different programming languages. I don't know how exact we want or need to be here. From my perspective, we can keep fingers crossed and see what happens :-)

uhlmannm (approved):

uhlmannm (approved):

vocabularies/Analytics.json
```diff
@@ -161,6 +164,116 @@
...
+          {
+            "$Name": "SiblingOrder",
+            "$Collection": true,
+            "$Type": "Analytics.MultiLevelExpandSiblingOrder",
+            "@Core.Description": "Sort specification to apply to all direct descendants of a given entry in the resulting leveled hierarchy"
+          },
+          {
+            "$Name": "ShowLevels",
+            "$Type": "Edm.Int64",
+            "@Core.Description": "Number N of levels to be shown in the initial expansion",
```
agoerler, line 205: do we start counting with 0 or with 1? 1, I assume.

vocabularies/Analytics.json
```diff
@@ -161,6 +164,116 @@
...
+            "$Name": "ExpandEntries",
+            "$Collection": true,
+            "$Type": "Analytics.MultiLevelExpandEntry",
+            "@Core.Description": "Entries with exceptional expansion",
+            "@Core.OptionalParameter": {}
+          },
+          {
+            "$Name": "SubtotalsAtBottom",
+            "$Type": "Edm.Bool",
+            "@Core.Description": "Whether to duplicate the group headers so that they appear before and after their descendants",
```
agoerler, line 219: "AtBottom" vs "before and after"

vocabularies/Analytics.md
```diff
@@ -8,38 +8,116 @@ Terms for annotating analytical resources
...
+    /aggregate($count as LeavesCount),
+  aggregate(Amount,Currency),
+  Analytics.MultiLevelExpand(
+    Levels=[{"DimensionProperties":["Country"],"AdditionalProperties":["CountryName"]},
+            {"DimensionProperties":["Region"],"AdditionalProperties":["RegionName"]},
+            {"DimensionProperties":["Segment","Industry"],"AdditionalProperties":[]}],
+    Aggregation=["Amount","Currency"],
+    SiblingOrder=[{"Property":"Amount","Descending":true}],
+    ShowLevels=2,
+    ExpandEntries=[{"Entry":["US"],"Levels":0},
```
agoerler, line 50: can I omit "Levels" to indicate "Levels":null ?

vocabularies/Analytics.md
```diff
@@ -8,38 +8,116 @@ Terms for annotating analytical resources
...
+
+<a name="MultiLevelExpand"></a>
+### [MultiLevelExpand](Analytics.xml#L182) *([Experimental](Common.md#Experimental))*
+
+`$apply` transformation that expands an unnamed leveled hierarchy with custom aggregation of certain properties
+
+Example transformation sequence:
+filter on columns `Industry`, `Amount` and `Currency`,
+order by `Amount` descending,
+show 2 levels, with two exceptions
```
agoerler, line 33: "show 3 levels" ?

as we start counting with 0

agoerler (commented):
some feedback

vocabularies/Analytics.json
```diff
@@ -161,6 +164,116 @@
...
+            "$Collection": true,
+            "$Type": "Edm.EntityType",
+            "@Core.Description": "Entity set to be processed"
+          },
+          {
+            "$Name": "Levels",
+            "$Collection": true,
+            "$Type": "Analytics.MultiLevelExpandLevel",
+            "@Core.Description": "Collection of aggregation levels forming a leveled hierarchy",
+            "@Core.LongDescription": "Each element in the collection defines the properties that constitute one level.\n            A property must not be referenced by more than one level.\n            The first element in the collection defines the property names of the coarsest level,\n            the following elements define the property names of consecutively finer-grained aggregation levels.\n            The function result is the leveled hierarchy with these levels in preorder,\n            entries on the finest-grained level cannot be expanded further.\n            All referenced properties must be groupable."
```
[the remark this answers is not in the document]
HeikoTheissen: No, because `CountryName` is groupable and can appear here, but is not a dimension. It is a dimension text, there are also dimension attributes.

vocabularies/Analytics.json
HeikoTheissen: There is nothing to be shown at level 0. `ShowLevels` = 0 gives an empty result.

vocabularies/Analytics.json
HeikoTheissen: The name reflects what you see in the UI. The description makes clear the the row appears duplicated in the result set.

vocabularies/Analytics.md
HeikoTheissen: No, because Gateway insists on all properties in a complex type, unless they have a default value. But null cannot be defined as default value, see https://github.com/oasis-tcs/odata-specs/issues/392.

vocabularies/Analytics.md
HeikoTheissen: `ShowLevels` = 2 means that we display 2 levels: `Country` and `Region`.

uhlmannm (dismissed):

vocabularies/Analytics.md
HeikoTheissen: > Hence I would assume that here `ShowLevels = 0` would also indicate to show the root only, i.e. only the global aggregate.

But the result returned by `Analytics.MultiLevelExpand` does not contain a root entry. The global aggregate is added to the result _outside_ of the function call, see line 198:
https://github.com/SAP/odata-vocabularies/blob/0ae829984c1f5e7c882bbb9718700ff97a23ce66/vocabularies/Analytics.xml#L195-L199

uhlmannm (dismissed):

vocabularies/Analytics.xml
```diff
@@ -176,6 +179,138 @@
...
+            <String>All properties in this collection must be custom aggregates.</String>
+          </Annotation>
+        </Parameter>
+        <Parameter Name="SiblingOrder" Type="Collection(Analytics.MultiLevelExpandSiblingOrder)" Nullable="false">
+          <Annotation Term="Core.Description" String="Sort specification to apply to all direct descendants of a given entry (so-called siblings) in the resulting leveled hierarchy" />
+        </Parameter>
+        <Parameter Name="ShowLevels" Type="Edm.Int64" Nullable="false">
+          <Annotation Term="Core.Description" String="Number N of levels to be shown in the initial expansion" />
+          <Annotation Term="Core.LongDescription">
+            <String>The initial expansion shows the first N levels as defined in `Levels` (0 ≤ N ≤ length of `Levels`).
```
HeikoTheissen, line 242: Or should that be "1 ≤ N ≤ length of `Levels`"?

vocabularies/Analytics.md
agoerler: One could argue that `Analytics.MultiLevelExpand` shouldn't make an assumption on how it's used. But if the root level is never included the wording is OK.

vocabularies/Analytics.xml
agoerler: I would like to suggest that we start counting the levels form 1 if we don't include the root level yet. If we later decide that there's a use case to include the root level we could use `ShowLevels : 0` to specify root only.

vocabularies/Analytics.json
```diff
@@ -161,6 +164,116 @@
...
+      "@Core.Description": "Expansion state of an entry in an [unnamed leveled hierarchy](#MultiLevelExpand)",
+      "Entry": {
+        "$Collection": true,
+        "@Core.Description": "An entry on a given [level](#MultiLevelExpandLevel) is identified by a list of values for the `DimensionProperties` that constitute all levels up to and including the given one",
+        "@Core.LongDescription": "The values are cast to strings as in the OData `cast` function."
+      },
+      "Levels": {
+        "$Type": "Edm.Int64",
+        "$Nullable": true,
+        "@Core.Description": "Number of levels to be expanded, null means all levels, 0 means collapsed"
```
agoerler, line 268: this is aligned with TopLevels.ExpandLevels

but 

`ShowLevels` is not aligned with `TopLevels.Levels`

, as far as I understand

vocabularies/Analytics.md
```diff
@@ -8,38 +8,117 @@ Terms for annotating analytical resources
...
```
agoerler, line 33:
```suggestion
show two levels, with two exceptions
```

to consitently use words for numbers

vocabularies/Hierarchy.json
```diff
@@ -33,8 +33,43 @@
...
+        "@Core.Description": "Drill state of an entry",
+        "@Validation.AllowedValues": [
+          {
+            "Value": "expanded",
+            "@Core.Description": "The entry precedes entries from finer-grained aggregation levels"
+          },
+          {
+            "Value": "subtotal",
+            "@Core.Description": "The entry follows entries from finer-grained aggregation levels and contains subtotals"
+          },
```
agoerler, line 56: I don't think it's ideal that `RecursiveHierarchy` but the value `subtotal` is not allowed in the `RecursiveHierarchy`.
Also the wording of the drill states is quite different here (although the "visual interpretation" on the UI is actually the same).

vocabularies/Hierarchy.json
```diff
@@ -33,8 +33,43 @@
...
+          {
+            "Value": "collapsed",
+            "@Core.Description": "The entry belongs to the coarsest-grained non-expanded aggregation level, but is not a leaf"
+          }
+        ]
+      },
+      "DistanceFromRoot": {
+        "$Type": "Edm.Int64",
+        "$Nullable": true,
+        "@Core.Description": "Position of the current aggregation level in the list of all aggregation levels"
```
agoerler, line 67: Would be nice if we could align this with the `RecursiveHierarchy`

vocabularies/Hierarchy.json
```diff
@@ -33,8 +33,43 @@
...
+            "@Core.Description": "The entry precedes entries from finer-grained aggregation levels"
+          },
+          {
+            "Value": "subtotal",
+            "@Core.Description": "The entry follows entries from finer-grained aggregation levels and contains subtotals"
+          },
+          { "Value": "leaf", "@Core.Description": "The entry belongs to the finest-grained aggregation level" },
+          {
+            "Value": "collapsed",
+            "@Core.Description": "The entry belongs to the coarsest-grained non-expanded aggregation level, but is not a leaf"
```
agoerler, line 60: > to the coarsest-grained non-expanded aggregation level
or to the finest-grained non-expanded aggregation level ? Or are these the same levels?

I think only one level is non-expanded.

```suggestion
            "@Core.Description": "The entry belongs to a non-expanded aggregation level, but is not a leaf"
```

agoerler (commented):
I think this proposal exactly captures the needs of the UI for an "auto expand" function.

I thinks it's good that we try to identify commonalities between the recursive hierarchies and the leveled hierarchies. It also makes sense to have some "inheritance" between the hierarchy types. 

Maybe we would have an "abstract" `HierarchyType` with subtypes for leveled and recursive hierarchies.

I understand that the leveled hierarchies don't contain a root level with a grand total as we assume the grand total is requested by means of `concat` anyhow. But I think this might introduce one-off issue if compared to `TopLevels` but I might be wrong. I think we might want to introduce a "grand-total root-level" later and hence the counting of levels should allow for this.

I also think that the parameter names could be harmonized in `TopLevels` vs `MultiLevelExpand`:

|   | TopLevels | MultiLevelExpand |
| --- | --- | --- |
| number of levels to expand | Levels | ShowLevels |
| nodes to expand | ExpandLevels | ExpandEntries |
| dimension, which define the hierarchy | n/a | Levels |

Overall, I think this proposal is OK and fits the needs. But I think it leaves room for more harmonization between `TopLevels` and `MultiLevelExpand`, which might make things easier for the implementors.

vocabularies/Analytics.md
HeikoTheissen: I deliberately used "2 levels" because this corresponds to `ShowLevels=2` in the request, and "two exceptions" because the corresponding part of the request does not contain the integer `2`.

vocabularies/Hierarchy.json
HeikoTheissen: I chose a different description because the definitions of leveled and recursive hierarchies differ.

vocabularies/Analytics.json
```diff
@@ -161,6 +164,116 @@
...
+      }
+    ],
+    "MultiLevelExpandLevel": {
+      "$Kind": "ComplexType",
+      "@Common.Experimental": true,
+      "@Core.Description": "Property names constituting a level in an [unnamed leveled hierarchy](#MultiLevelExpand)",
+      "@Core.LongDescription": "`DimensionProperties` must be used to identify entries in [`ExpandEntries/Entry`](#MultiLevelExpandEntry),\n          otherwise they have the same effect as `AdditionalProperties`.",
+      "DimensionProperties": {
+        "$Collection": true,
+        "@Core.Description": "A non-empty list of property names that define a combination of dimension values"
```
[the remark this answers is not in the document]
HeikoTheissen: The fact that they are used for grouping is already explained in
https://github.com/SAP/odata-vocabularies/blob/20ee02904214f06343186de6aafbad381a31c98b/vocabularies/Analytics.xml#L222-L224

The point I want to make here is the distinction between "dimensions" and "dimension attributes" (or "dimension texts").

vocabularies/Analytics.json
HeikoTheissen: It is aligned in so far as `Levels=2` means two levels of output in both cases:
- In the recursive case, this is the root (`Sales`) plus one level (`US` and `EMEA`).
- In the leveled case, this is two grouping levels without a root.

uhlmannm (approved):

agoerler (approved):

PR #294, opened 2023-11-15
Interval label

BerSie:
Provides ability to define labels for an interval


ralfhandl (dismissed):

ralfhandl:
Do you want to mark the new property as "experimental", or are you sure that you will implement it exactly in that way?

ralfhandl (approved):

PR #295, opened 2023-11-17
Compute rank for a newly created or copied hierarchy node

HeikoTheissen:
A function that takes a node and behaves like `TopLevels` on the subset consisting of just that node.

### DIFF
--- a/vocabularies/Hierarchy.xml
+++ b/vocabularies/Hierarchy.xml
@@ -265,6 +265,42 @@ The template actions themselves cannot be invoked.</String>
         </Property>
       </ComplexType>
 
+      <Function Name="EvaluateHierarchy" IsBound="true" EntitySetPath="Node">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Evaluates a node in a recursive hierarchy and adds the [`RecursiveHierarchy`](#RecursiveHierarchy) annotation" />
+        <Annotation Term="Core.LongDescription">
+          <String>Can be used to determine, for example, the `LimitedRank` of a newly created or copied node
+in the JSON batch request:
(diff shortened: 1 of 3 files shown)


vocabularies/Hierarchy.xml
```diff
@@ -265,6 +265,42 @@ The template actions themselves cannot be invoked.</String>
...
+```json
+{"requests": [{
+  "id": "1",
+  "method": "post",
+  "url": "HierarchyDirectory(1)/Nodes('A')/CopyAction"
+}, {
+  "id": "2",
+  "dependsOn": ["1"],
+  "method": "get",
+  "url": "$1/Hierarchy.EvaluateHierarchy(
```
ThomasChadzelek, line 283: Don't we need more information for a LimitedRank? What about the Levels parameter? And can we really replace the InputSet by a single node?

ThomasChadzelek: One entity is perfectly fine, but do we have all information at hand to compute e.g. the LimitedRank? Wouldn't that depend on Levels as well? And doesn't it also depend on the input set (you could filter it beforehand, at least in future)?

HeikoTheissen:
This is not flexible enough. Use `@Core.ContentID` instead:
```json
{"requests": [{
  "id": "1",
  "method": "post",
  "url": "HierarchyDirectory(1)/Nodes('A')/CopyAction",
  "@Core.ContentID": "copiedNode"
}, {
  "id": "2",
  "dependsOn": ["1"],
  "method": "get",
  "url": "HierarchyDirectory(1)/Nodes?$apply=Hierarchy.TopLevels(...)/filter(ID eq $copiedNode/ID)"
}]}
```

PR #297, opened 2023-11-24
ValueLists can show values immediately

BerSie:
ValuesLists with a very small number of fixed values (annotated with Common.ValueListWithFixedValues) can decide that its values can be immediately shown (without any user interaction). Examples would be radio buttons for single-valued properties or check boxes for multi-values properties.


ralfhandl (approved):

ralfhandl (changes requested):
Please tag the new term as "experimental"

BerSie:
> Please tag the new term as "experimental"

Sorry, this is really something that I forget regularly.

ralfhandl (dismissed):

vocabularies/Common.xml
```diff
@@ -707,6 +707,16 @@ The minimum structure is
 
       <Term Name="ValueListWithFixedValues" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property Parameter">
         <Annotation Term="Core.Description" String="If specified as true, there's only one value list mapping and its value list consists of a small number of fixed values" />
+        <Annotation Term="Validation.ApplicableTerms">
+          <Collection>
+            <String>ValueListShowValues"</String>
```
HeikoTheissen, line 712:
```suggestion
            <String>Common.ValueListShowValues</String>
```

vocabularies/Common.xml
```diff
@@ -707,6 +707,16 @@ The minimum structure is
...
+          </Collection>
+        </Annotation>
+      </Term>
+
+      <Term Name="ValueListShowValues" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Annotation">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="A value list with a very small number of fixed values, can decide to show all values immediately" />
       </Term>
```
HeikoTheissen, line 720: Is it planned to apply this term to other annotations later? Otherwise, I would prefer a new term `Common.ValueListWithImmediateValues` as an alternative to `Common.ValueListWithFixedValues`.

HeikoTheissen (changes requested):

vocabularies/Common.xml
BerSie: The reason I did it that way was that the value list should have fixed (means a small number of) values. Only for these value lists I would allow it.

BerSie: Thank you for the proposed change!

ralfhandl (dismissed):

vocabularies/Common.xml
```diff
@@ -707,6 +707,16 @@ The minimum structure is
...
+        <Annotation Term="Validation.ApplicableTerms">
+          <Collection>
+            <String>Common.ValueListShowValues</String>
+          </Collection>
+        </Annotation>
+      </Term>
+
+      <Term Name="ValueListShowValues" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Annotation">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="A value list with a very small number of fixed values, can decide to show all values immediately" />
```
HeikoTheissen, line 719:
```suggestion
        </Term>
        <Term Name="ValueListWithImmediateValues" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Property Parameter">
          <Annotation Term="Core.Description" String="If specified as true, there's only one value list mapping and its value list consists of a small number of fixed values that can be shown immediately" />
```

HeikoTheissen: Alternatively, can't you use `Common.ValueListMapping/FetchValues = 1` for that?

BerSie: Hi Heiko, 
thank you very much for your proposals - really appreciate that! First I would like to comment on the second one (using FetchValues), which I immediately liked, but then I thought about apps/customers that used that already, and afterwards set the _Common.ValueListWithFixedValues_ annotation. Today they see in Fiori Elements a dropdown listbox. But after FE would use that to show e.g. a group of radio buttons in this constellation they would see radio buttons (independent if the amount of values really fits this visualization) - I think they will regard this as a regression. 
Wrt the other proposal I personally would make this term also only available as annotation of _Common.ValueListWithFixedValues_. I see the new term as a kind of "specialization" of this existing term - it's a clear subset.
But as said these are my personal opinions - maybe @ralfhandl or @stefan0816 want to comment on that.

BerSie: Thanks, Ralf! So, if Heiko doesn't veto, we stick to the original term definition, but we might discuss the name of the new term: We have "ValueListShowValues", "ValueListWithImmediateValues", and "ValueListShowValuesImmediately" - the last one seems to be the most expressive, but it's also the longest. So, what's your choice?

ralfhandl: I like expressive and prefer `ValueListShowValuesImmediately`.

HeikoTheissen (approved):

ralfhandl (approved):

PR #306, opened 2024-02-06
Common.DocumentationRef

HeikoTheissen:


vocabularies/Common.xml
```diff
@@ -74,6 +74,17 @@
         <Annotation Term="Core.IsLanguageDependent" />
       </Term>
 
+      <Term Name="DocumentationRef" Type="Edm.String" Nullable="false" AppliesTo="EntityType Property Parameter ReturnType">
+        <Annotation Term="Core.IsURL" />
+        <Annotation Term="Core.Description" String="A URI referencing language-dependent documentation for the annotated model element" />
+        <Annotation Term="Core.Example">
+          <Record Type="Core.PrimitiveExampleValue">
+            <PropertyValue Property="Description" String="URN scheme understood by SAP Companion" />
+            <PropertyValue Property="Value" String="urn:sap:companion:DataElement:CURRENCY:G1Y_000" />
```
HeikoTheissen, line 83: Please update this example

uhlmannm (dismissed):

vocabularies/Common.json
```diff
@@ -59,6 +59,17 @@
       "@Core.Description": "A short, human-readable text suitable for tool tips in UIs",
       "@Core.IsLanguageDependent": true
     },
+    "DocumentationRef": {
+      "$Kind": "Term",
+      "$AppliesTo": ["EntityType", "Property", "Parameter", "ReturnType"],
```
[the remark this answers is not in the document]
uhlmannm: Function, Action and FunctionImport are not relevant. ComplexType could be used. TypeDefinition also seems unneccessary.

vocabularies/Common.json
```diff
@@ -59,6 +59,17 @@
...
+      "@Core.IsURL": true,
+      "@Core.Description": "A URI referencing language-dependent documentation for the annotated model element",
```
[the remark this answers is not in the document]
uhlmannm: Which means that you could tag anything with the DocumentationRef? The UI would definitely only evaluate it at a few places. And GW would not send it everywhere. But from an annotation point of view that could still be okay. Just external viewers could expect too much.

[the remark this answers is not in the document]
ralfhandl: I would not restrict it to what ABAP/GW/... currently supports as this is in Common and not in an ABAP vocabulary.

vocabularies/Common.xml
```diff
@@ -74,6 +74,18 @@
         <Annotation Term="Core.IsLanguageDependent" />
       </Term>
 
+      <Term Name="DocumentationRef" Type="Edm.String" Nullable="false" AppliesTo="EntityType Property Parameter ReturnType">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.IsURL" />
+        <Annotation Term="Core.Description" String="A URI referencing language-dependent documentation for the annotated model element" />
+        <Annotation Term="Core.Example">
+          <Record Type="Core.PrimitiveExampleValue">
+            <PropertyValue Property="Description" String="URN scheme understood by SAP Companion" />
```
BerSie, line 83: This product has in the last two or three years at least three product name changes - not sure if I would mention it here. Is there a difference to "normal" URLs?

uhlmannm: Yes. This is an example from one of our test services.
> `<Annotation Term="SAP__common.DocumentationRef" String="urn:sap-com:documentation:key?=type=DE&id=SPRAS"/>`

So it is in the end pretty ABAP specific. If we would want to enhance it for other backends, also Companion would need to learn about that.

BerSie: > Yes. This is an example from one of our test services.
> 
> > `<Annotation Term="SAP__common.DocumentationRef" String="urn:sap-com:documentation:key?=type=DE&id=SPRAS"/>`
> 
> So it is in the end pretty ABAP specific. If we would want to enhance it for other backends, also Companion would need to learn about that.

OK - thanks, @uhlmannm . The question is if this urn schema was also used for xray, for FLP In-App Help, and Web Assistant, which is now used for SAP Companion - so either it's only relevant for this specific tool, or it's relevant for the in-app help framework from SAP. Usually we want to define here a "timeless" abstraction ...

HeikoTheissen: This is just an _example_.

BerSie: Thanks, @HeikoTheissen!
Nevertheless, it's only yesterday that I heard that there is the next name under way. I assume that in two years nobody knows what companion is/was. But don't mind - it was just a comment. For me the term itself is completely fine.

BerSie (dismissed):

vocabularies/Common.xml
uhlmannm: But in the end they evaluate the schema. Should we write rather "URN scheme containing the information to identify the help text"?

uhlmannm (dismissed):

vocabularies/Common.xml
```diff
@@ -74,6 +74,21 @@
         <Annotation Term="Core.IsLanguageDependent" />
       </Term>
 
+      <Term Name="DocumentationRef" Type="Edm.String" Nullable="false">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.IsURL" />
```
ralfhandl, line 79:
```suggestion
```

ralfhandl (changes requested):
Is not a URL

ralfhandl (approved):

BerSie (approved):
under the given circumstances it's OK for me, but I will never understand why we make nearly everything in our system accessible via (OData) URL, but our documentation (static texts) can't be directly addressed by URLs.

PR #311, opened 2024-03-12
Common.ValueListParameterConstants

BerSie:
Allows to provide multiple fixed values / constants as value list parameter


MarcelWaechter (approved):
Looks good to me. But would be good if Ralf or Heiko have a second look into it.
Thanks :)

vocabularies/Common.xml
```diff
@@ -795,6 +795,15 @@ The minimum structure is
           <Annotation Term="Core.Description" String="Initial value, e.g. empty string, is a valid and significant value" />
         </Property>
       </ComplexType>
+      <ComplexType Name="ValueListParameterConstants" BaseType="Common.ValueListParameter">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="Constants" Type="Collection(Edm.PrimitiveType)">
+          <Annotation Term="Core.Description" String="List of constant values that are used to filter the value list with `eq` comparisons connected by `or` operators" />
```
HeikoTheissen, line 801:
```suggestion
          <Annotation Term="Core.Description" String="List of constant values that are used to filter the value list with `eq` comparisons connected by `or` operators, using the same representation as property default values, see [CSDL XML, 7.2.7 Default Value](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_DefaultValue)" />
```
Use the same explanation as for the single constant?

vocabularies/Common.xml
```diff
@@ -795,6 +795,15 @@ The minimum structure is
           <Annotation Term="Core.Description" String="Initial value, e.g. empty string, is a valid and significant value" />
         </Property>
       </ComplexType>
+      <ComplexType Name="ValueListParameterConstants" BaseType="Common.ValueListParameter">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="Constants" Type="Collection(Edm.PrimitiveType)">
```
HeikoTheissen, line 800: Do we allow an empty list of constants? If yes, does
```json
"Parameters": [{
  "@type": "#Common.ValueListConstant",
  "ValueListProperty": "Country",
  "Constant": "DE"
}, {
  "@type": "#Common.ValueListConstants",
  "ValueListProperty": "City",
  "Constants": []
}]
```
mean `Country eq 'DE'`?

BerSie: I would say yes - if we should allow it? What do you or the others think?

HeikoTheissen (approved):

vocabularies/Common.xml
```diff
@@ -795,6 +795,18 @@ The minimum structure is
...
+          <Annotation Term="Core.Description" String="List of constant values that are used to filter the value list with `eq` comparisons connected by `or` operators, using the same representation as property default values, see [CSDL XML, 7.2.7 Default Value](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_DefaultValue)" />
+          <Annotation Term="Core.LongDescription">
+            <String>An empty list means a vacuous filter condition</String>
+          </Annotation>
+        </Property>
+        <Property Name="InitialValueIsSignificant" Type="Edm.Boolean" DefaultValue="false" Nullable="false">
+          <Annotation Term="Core.Description" String="Initial value, e.g. empty string, is a valid and significant value" />
+        </Property>
```
ralfhandl, line 808: Do we really need this? Why should `<String></String>` not be significant if I explictly wrote it?

What about the same property on `ValueListParameterConstant` that is still experimental: can we remove it?

PR #313, opened 2024-03-12
ValueListParameterConstants: initial values are by default significant, no flag needed

ralfhandl:


vocabularies/UI.md
ralfhandl, line 1: Apparently the MD file was run through Prettier after build

BerSie (approved):

HeikoTheissen (commented):
Remove the initial values flag from the singular version as well?

ralfhandl:
> Remove the initial values flag from the singular version as well?

I'd like to do that and will do this in a separate pull request, depending on feedback from the dev team that requested the experimental property.

PR #317, opened 2024-05-08
New term UI.InputMask

BerSie:
New term UI.InputMask was added along with a simple example. I'm still waiting for more sophisticated examples from the application developers requesting this feature.
Idea is to have a defaultset of rules for mask symbols like "*", "C", "9", but give the chance to define own rules for their masks.

### DIFF
--- a/examples/UI.InputMask-sample.xml
+++ b/examples/UI.InputMask-sample.xml
@@ -19,15 +19,27 @@
           <PropertyRef Name="CustomerID" />
         </Key>
         <Property Name="CustomerID" Type="Edm.Guid" Nullable="false"/>
-        <Property Name="DUNS" Type="Edm.String" Nullable="true" MaxLength="9"/>
+        <Property Name="SwiftCode" Type="Edm.String" Nullable="true" MaxLength="11"/>
         <!-- ... and a lot more -->
       </EntityType>
 
-      <Annotations Target="Examples.Customer_Type/DUNS">
+      <Annotations Target="Examples.Customer_Type/SwiftCode">
         <Annotation Term="UI.InputMask">
           <Record>
-            <PropertyValue Property="Mask" String="99-999-9999" />
-            <!-- Data Universal Numbering System number issued by Dun & Bradstreet - might also be used without dashes "999999999"-->
+            <PropertyValue Property="Mask" String="CCCCCCAAAAA" />
+            <PropertyValue Property="Placeholder" String="." />
+            <PropertyValue Property="Rules">
+              <Collection>
+                <Record>
+                  <PropertyValue Property="MaskSymbol" String="C" />
+                  <PropertyValue Property="RegExp" String="[A-Z]" />
+                </Record>
+                <Record>
+                  <PropertyValue Property="MaskSymbol" String="A" />
+                  <PropertyValue Property="RegExp" String="[A-Z0-9]" />
+                </Record>
+              </Collection>
+            </PropertyValue>
           </Record>
         </Annotation>
       </Annotations>
--- a/examples/UI.InputMask-sample.xml
+++ b/examples/UI.InputMask-sample.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
+  <edmx:Reference Uri="https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.xml">
+    <edmx:Include Namespace="Org.OData.Core.V1" Alias="Core" />
+  </edmx:Reference>
+  <edmx:Reference Uri="https://sap.github.io/odata-vocabularies/vocabularies/UI.xml">
+    <edmx:Include Namespace="com.sap.vocabularies.UI.v1" Alias="UI" />
+  </edmx:Reference>
+
+  <edmx:DataServices>
+    <Schema Namespace="Examples" xmlns="http://docs.oasis-open.org/odata/ns/edm" Alias="this">
+
+      <EntityContainer Name="container">
+        <EntitySet Name="Customers" EntityType="this.Customer_Type" />
+      </EntityContainer>
+
+      <EntityType Name="Customer_Type">
+        <Key>
+          <PropertyRef Name="CustomerID" />
+        </Key>
+        <Property Name="CustomerID" Type="Edm.Guid" Nullable="false"/>
+        <Property Name="DUNS" Type="Edm.String" Nullable="true" MaxLength="9"/>
+        <!-- ... and a lot more -->
+      </EntityType>
+
+      <Annotations Target="Examples.Customer_Type/DUNS">
+        <Annotation Term="UI.InputMask">
+          <Record>
+            <PropertyValue Property="Mask" String="999999999" />
+            <!-- Data Universal Numbering System number issued by Dun & Bradstreet - nine-digit number might also be used with dashes for better readability "99-999-9999"-->
+          </Record>
+        </Annotation>
+      </Annotations>
+
+    </Schema>
+  </edmx:DataServices>
+</edmx:Edmx>
--- a/examples/UI.InputMask-sample.xml
+++ b/examples/UI.InputMask-sample.xml
@@ -26,8 +26,8 @@
       <Annotations Target="Examples.Customer_Type/DUNS">
         <Annotation Term="UI.InputMask">
           <Record>
-            <PropertyValue Property="Mask" String="999999999" />
-            <!-- Data Universal Numbering System number issued by Dun & Bradstreet - nine-digit number might also be used with dashes for better readability "99-999-9999"-->
+            <PropertyValue Property="Mask" String="99-999-9999" />
+            <!-- Data Universal Numbering System number issued by Dun & Bradstreet - might also be used without dashes "999999999"-->
           </Record>
         </Annotation>
       </Annotations>
--- a/examples/UI.InputMask-sample.xml
+++ b/examples/UI.InputMask-sample.xml
@@ -19,10 +19,19 @@
           <PropertyRef Name="CustomerID" />
         </Key>
         <Property Name="CustomerID" Type="Edm.Guid" Nullable="false"/>
+        <Property Name="DUNS" Type="Edm.String" Nullable="true" MaxLength="9"/>
         <Property Name="SwiftCode" Type="Edm.String" Nullable="true" MaxLength="11"/>
         <!-- ... and a lot more -->
       </EntityType>
 
+      <Annotations Target="Examples.Customer_Type/DUNS">
+        <Annotation Term="UI.InputMask">
+          <Record>
+            <PropertyValue Property="Mask" String="99-999-9999" />
+            <!-- Data Universal Numbering System number issued by Dun & Bradstreet - might also be used without dashes "999999999"-->
+          </Record>
+        </Annotation>
+      </Annotations>
       <Annotations Target="Examples.Customer_Type/SwiftCode">
         <Annotation Term="UI.InputMask">
           <Record>
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1417,7 +1417,7 @@ So, masks can be especially useful for input fields that have a fixed pattern, e
           <Annotation Term="Core.Description" String="A single character symbol to be shown where the user can type a character" />
         </Property>
         <Property Name="InputMaskRules" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
-          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask." />
+          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask" />
           <Annotation Term="Core.LongDescription">
             <String>The following rules are defined as default and don't need to be listed here: * = ., C = [a-zA-Z], 9 = [0-9]</String>
           </Annotation>
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1402,6 +1402,7 @@ The trend is
       </Term>
 
       <Term Name="InputMask" Type="UI.InputMaskType" Nullable="false" AppliesTo="Property Parameter">
+        <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode" />
         <Annotation Term="Core.LongDescription">
           <String>Input masks improve readability and help to enter data correctly. 
@@ -1410,6 +1411,7 @@ So, masks can be especially useful for input fields that have a fixed pattern, e
         </Annotation>
       </Term>
       <ComplexType Name="InputMaskType">
+        <Annotation Term="Common.Experimental" />
         <Property Name="Mask" Type="Edm.String" Nullable="false">
           <Annotation Term="Core.Description" String="The mask to be applied to the property or the parameter" />
         </Property>
@@ -1424,6 +1426,7 @@ So, masks can be especially useful for input fields that have a fixed pattern, e
         </Property>
       </ComplexType>
       <ComplexType Name="InputMaskRuleType">
+        <Annotation Term="Common.Experimental" />
         <Property Name="MaskSymbol" Type="Edm.String" MaxLength="1" Nullable="false">
           <Annotation Term="Core.Description" String="A symbol in the mask that stands for a regular expression which must be matched by the user input in every position where this symbol occurs" />
         </Property>
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1401,6 +1401,37 @@ The trend is
         <Annotation Term="Core.IsLanguageDependent" />
       </Term>
 
+      <Term Name="InputMask" Type="UI.InputMaskType" Nullable="false" AppliesTo="Property Parameter">
+        <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode." />
+        <Annotation Term="Core.LongDescription">
+          <String>Input masks not only improve readability, but also help to enter data correctly. 
+So, masks can be especially useful for input fields that have a fixed pattern, e.g. international bank account numbers or social security numbers. 
+[Here](../examples/UI.InputMask-sample.xml) you can find an example for this annotation</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="InputMaskType">
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property" />
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" DefaultValue="_">
+          <Annotation Term="Core.Description" String="A single character symbol that is used as a substitute." />
+        </Property>
+        <Property Name="InputMaskRule" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask." />
+          <Annotation Term="Core.LongDescription">
+            <String>The following rules are defined as default and don't need to be listed here: * = [.], C = [a-zA-Z], 9 = [0-9]</String>
+          </Annotation>
+        </Property>
+      </ComplexType>
+      <ComplexType Name="InputMaskRuleType">
+        <Property Name="MaskSymbol" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The character in the mask that should be validated." />
+        </Property>
+        <Property Name="Rule" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The rule that defines the valid values for the mask symbol as regular expression." />
+        </Property>
+      </ComplexType>
+
       <Term Name="TextArrangement" Type="UI.TextArrangementType" Nullable="false" AppliesTo="Annotation EntityType">
         <Annotation Term="Core.Description" String="Describes the arrangement of a code or ID value and its text" />
         <Annotation Term="Core.LongDescription">
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1404,19 +1404,19 @@ The trend is
       <Term Name="InputMask" Type="UI.InputMaskType" Nullable="false" AppliesTo="Property Parameter">
         <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode." />
         <Annotation Term="Core.LongDescription">
-          <String>Input masks not only improve readability, but also help to enter data correctly. 
-So, masks can be especially useful for input fields that have a fixed pattern, e.g. international bank account numbers or social security numbers. 
+          <String>Input masks improve readability and help to enter data correctly. 
+So, masks can be especially useful for input fields that have a fixed pattern, e.g. DUNS numbers or similar. 
 [Here](../examples/UI.InputMask-sample.xml) you can find an example for this annotation</String>
         </Annotation>
       </Term>
       <ComplexType Name="InputMaskType">
         <Property Name="Mask" Type="Edm.String" Nullable="false">
-          <Annotation Term="Core.Description" String="The mask to be applied to the property" />
+          <Annotation Term="Core.Description" String="The mask to be applied to the property or the parameter" />
         </Property>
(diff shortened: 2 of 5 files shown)


vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+So, masks can be especially useful for input fields that have a fixed pattern, e.g. international bank account numbers or social security numbers. 
+[Here](../examples/UI.InputMask-sample.xml) you can find an example for this annotation</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="InputMaskType">
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property" />
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" DefaultValue="_">
+          <Annotation Term="Core.Description" String="A single character symbol that is used as a substitute." />
```
HeikoTheissen, line 1417: A substitute for what? Can you give an example?

BerSie: This is the symbol that's shown where there is no user input, eg. DE__ ____ ____ ____ ____ __ for a german IBAN. Maybe substitute is not really the right choice for the explanation - do you have a better one?
Thanks!

HeikoTheissen: But the UI mask for IBAN would be `DE99 9999 9999 9999 9999 99` since the variable characters are digits.

For a Swiss IBAN, `CH99 9999 9999 9999 9999 99` would not work, because the leading `C` can mean _any_ letter, right?

The example is still unclear to me.

HeikoTheissen, line 1417:
```suggestion
          <Annotation Term="Core.Description" String="A single character symbol to be shown where the user can type a character" />
```

BerSie: I wrote something wrt the placeholder that's shown e.g. in a UI control, and you are writing about (the definition of) a mask. 
An IBAN in general is not well suited for a mask as e.g. the length differs based on country. If you want to prepare your UI for a swiss IBAN, you can define another mask and different rules.

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        <Property Name="InputMaskRule" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask." />
+          <Annotation Term="Core.LongDescription">
+            <String>The following rules are defined as default and don't need to be listed here: * = [.], C = [a-zA-Z], 9 = [0-9]</String>
+          </Annotation>
+        </Property>
+      </ComplexType>
+      <ComplexType Name="InputMaskRuleType">
+        <Property Name="MaskSymbol" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The character in the mask that should be validated." />
```
HeikoTheissen, line 1428:
```suggestion
        <Property Name="MaskSymbol" Type="Edm.String" MaxLength="1" Nullable="false">
          <Annotation Term="Core.Description" String="A symbol in the mask that stands for a regular expression which must be matched by the user input in every position where this symbol occurs" />
```

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        </Property>
+        <Property Name="Rule" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The rule that defines the valid values for the mask symbol as regular expression." />
```
HeikoTheissen, line 1431:
```suggestion
        <Property Name="RegExp" Type="Edm.String" Nullable="false">
          <Annotation Term="Core.Description" String="The regular expression for which this symbol stands" />
```

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property" />
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" DefaultValue="_">
+          <Annotation Term="Core.Description" String="A single character symbol that is used as a substitute." />
+        </Property>
+        <Property Name="InputMaskRule" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask." />
+          <Annotation Term="Core.LongDescription">
+            <String>The following rules are defined as default and don't need to be listed here: * = [.], C = [a-zA-Z], 9 = [0-9]</String>
```
HeikoTheissen, line 1422:
```suggestion
            <String>The following rules are defined as default and don't need to be listed here: * = ., C = [a-zA-Z], 9 = [0-9]</String>
```
The regular expression `[.]` is matched only by a full stop.
Shall rule `C` not match ä or é?

HeikoTheissen (changes requested):

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        </Annotation>
+      </Term>
+      <ComplexType Name="InputMaskType">
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property" />
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" DefaultValue="_">
+          <Annotation Term="Core.Description" String="A single character symbol to be shown where the user can type a character" />
+        </Property>
+        <Property Name="InputMaskRule" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
```
HeikoTheissen, line 1419: How are these rules enforced?
* Is typing restricted to the allowed characters (immediate feedback)? Then users would need a visual indication which rule they have to obey at a given position (for example, different placeholders per rule).
* Or are rules checked only after the user has finished typing?

HeikoTheissen:
I suggest to let the mask consist of the placeholders themselves. Every character in the mask can be
* a character with an associated regular expression (which is `.` if the user can type anything). This character is then displayed as a placeholder until the user types something
* or a character without associated regular expression. This character is then display-only, and the cursor jumps over it (for example, the spaces in an IBAN or the hyphens in a DUNS number).

examples/UI.InputMask-sample.xml
```diff
@@ -0,0 +1,37 @@
...
+        <Property Name="CustomerID" Type="Edm.Guid" Nullable="false"/>
+        <Property Name="DUNS" Type="Edm.String" Nullable="true" MaxLength="9"/>
+        <!-- ... and a lot more -->
+      </EntityType>
+
+      <Annotations Target="Examples.Customer_Type/DUNS">
+        <Annotation Term="UI.InputMask">
+          <Record>
+            <PropertyValue Property="Mask" String="999999999" />
+            <!-- Data Universal Numbering System number issued by Dun & Bradstreet - nine-digit number might also be used with dashes for better readability "99-999-9999"-->
```
ralfhandl, line 30: Why don't we use the dashed version in the example?

vocabularies/UI.xml
BerSie: renaming the property is fine with me, but the documentation I don't like, esp. "this symbol". How about "The regular expression for the mask symbol"?

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
         <Annotation Term="Core.IsLanguageDependent" />
       </Term>
 
+      <Term Name="InputMask" Type="UI.InputMaskType" Nullable="false" AppliesTo="Property Parameter">
```
ralfhandl, line 1404: "InputMask" threw me off because we already have `Common.Masked` and `Common.MaskedValue` where "masking" means hiding.

examples/UI.InputMask-sample.xml
BerSie: Because it's said that the "modern version" is the one without dashes (see https://en.wikipedia.org/wiki/Data_Universal_Numbering_System). But of course we could change the example.

vocabularies/UI.xml
HeikoTheissen: Good point, suggest calling it `InputPattern` instead.

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode." />
+        <Annotation Term="Core.LongDescription">
+          <String>Input masks not only improve readability, but also help to enter data correctly.
```
ralfhandl, line 1407:
```suggestion
          <String>Input masks improve readability and help to enter data correctly. 
```

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
         <Annotation Term="Core.IsLanguageDependent" />
       </Term>
 
+      <Term Name="InputMask" Type="UI.InputMaskType" Nullable="false" AppliesTo="Property Parameter">
+        <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode." />
+        <Annotation Term="Core.LongDescription">
+          <String>Input masks not only improve readability, but also help to enter data correctly. 
+So, masks can be especially useful for input fields that have a fixed pattern, e.g. international bank account numbers or social security numbers. 
+[Here](../examples/UI.InputMask-sample.xml) you can find an example for this annotation</String>
```
ralfhandl, line 1409: Another reason to use the dashed variant of a DUNS number in the example.

vocabularies/UI.xml
BerSie: Typing is restricted to allowed characters. A UI.Placeholder might help here :-) See: https://sapui5.hana.ondemand.com/#/entity/sap.m.MaskInput/sample/sap.m.sample.MaskInput

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        </Annotation>
+      </Term>
+      <ComplexType Name="InputMaskType">
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property" />
```
ralfhandl, line 1414:
```suggestion
          <Annotation Term="Core.Description" String="The mask to be applied to the string value" />
```

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" DefaultValue="_">
```
ralfhandl, line 1416: Please add example using `Placeholder`

vocabularies/UI.xml
ralfhandl, line 1419: Please add an example using at least two input mask rule entries.

Should the property have a plural name?

ralfhandl (changes requested):
See comments

vocabularies/UI.xml
BerSie: https://www.google.com/search?q=inputmask

vocabularies/UI.xml
BerSie: yes, plural is good. As written in my first sentence - I'm still waiting for real-world examples. If I have that, I can provide more examples as wished, but currently I'm looking for examples for a requirement that was raised by others, without bringing up valid examples.

vocabularies/UI.xml
BerSie: mentioning both sound better for me

vocabularies/UI.xml
BerSie: I will change the example

vocabularies/UI.xml
BerSie: will change that

BerSie:
> I suggest to let the mask consist of the placeholders themselves. Every character in the mask can be associated with a regular expression (which is `.` if the user can type anything). This character is then displayed as a placeholder until the user types something. Characters without associated regular expression are display-only, and the cursor jumps over them (for example, the spaces in an IBAN or the hyphens in a DUNS number).
> 
> ```json
> "IBAN@UI.InputPattern": {
>   "Pattern": "DE__ ____ ____ ____ ____ __",
>   "Rules": [
>     {"Symbol": "_", "RegExp": "[0-9]"}
>   ]
> },
> "DUNSNumber@UI.InputPattern": {
>   "Pattern": "__-___-____",
>   "Rules": [
>     {"Symbol": "_", "RegExp": "[0-9]"}
>   ]
> },
> "AlternativeDUNSNumber@UI.InputPattern": {
>   "$Path": "DUNSNumber@UI.InputPattern"
> }
> ```

no, no - what about input fields where the first two positions are letters and the rest are numbers - there is only one placeholder symbol (in the UI). We have to differentiate between the defined mask and the validations and what is used in the UI as a placeholder.

examples/UI.InputMask-sample.xml
BerSie: as said below - will change that

HeikoTheissen:
> there is only one placeholder symbol (in the UI)

How does the user know what they are supposed to type at a given placeholder symbol, then? How are the rules enforced?

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode." />
+        <Annotation Term="Core.LongDescription">
+          <String>Input masks improve readability and help to enter data correctly. 
+So, masks can be especially useful for input fields that have a fixed pattern, e.g. DUNS numbers or similar. 
+[Here](../examples/UI.InputMask-sample.xml) you can find an example for this annotation</String>
+        </Annotation>
+      </Term>
+      <ComplexType Name="InputMaskType">
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property or the parameter" />
```
HeikoTheissen, line 1414:
```suggestion
          <Annotation Term="Core.Description" String="The mask to be applied to the property or the parameter" />
          <Annotation Term="Core.LongDescription">
            <String>Symbols in the mask for which there is an `InputMaskRule` are initially displayed using the `Placeholder`, and what the user types in their position must match the associated regular expression. Other symbols are read-only and the cursor automatically moves over them as the user types.</String>
          </Annotation>
```

BerSie:
> > there is only one placeholder symbol (in the UI)
> 
> How does the user know what they are supposed to type at a given placeholder symbol, then? How are the rules enforced?

Use the UI.Placeholder annotation to give hints about the expected input. The rules are enforced by the control based on the defined mask and the mask rules.

vocabularies/UI.xml
BerSie: This sounds even for me, who created the term, far too complex - maybe we can write something similar in the example as an explanation. Maybe we can ask a UA colleague to make a proposal.

HeikoTheissen: We must specify the expected UI behavior somewhere.

vocabularies/UI.xml
HeikoTheissen: "The character in the mask that should be validated" is misleading. It is not the character in the mask that is validated but the character which the user typed.

BerSie: true

vocabularies/UI.xml
BerSie: yes - definitely.

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" DefaultValue="_">
+          <Annotation Term="Core.Description" String="A single character symbol to be shown where the user can type a character" />
+        </Property>
+        <Property Name="InputMaskRules" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
```
ralfhandl, line 1419: Thanks, now please add an example using this

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+            <String>The following rules are defined as default and don't need to be listed here: * = ., C = [a-zA-Z], 9 = [0-9]</String>
+          </Annotation>
+        </Property>
+      </ComplexType>
+      <ComplexType Name="InputMaskRuleType">
+        <Property Name="MaskSymbol" Type="Edm.String" MaxLength="1" Nullable="false">
+          <Annotation Term="Core.Description" String="A symbol in the mask that stands for a regular expression which must be matched by the user input in every position where this symbol occurs" />
+        </Property>
+        <Property Name="RegExp" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Regular expression that defines the valid values for the mask symbol." />
```
ralfhandl, line 1431:
```suggestion
          <Annotation Term="Core.Description" String="Regular expression that defines the valid values for the mask symbol" />
```

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
         <Annotation Term="Core.IsLanguageDependent" />
       </Term>
 
+      <Term Name="InputMask" Type="UI.InputMaskType" Nullable="false" AppliesTo="Property Parameter">
+        <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode." />
```
ralfhandl, line 1405:
```suggestion
        <Annotation Term="Core.Description" String="Properties or parameters annotated with this term will get a mask in edit mode" />
```

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+      </Term>
+      <ComplexType Name="InputMaskType">
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property or the parameter" />
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" DefaultValue="_">
+          <Annotation Term="Core.Description" String="A single character symbol to be shown where the user can type a character" />
+        </Property>
+        <Property Name="InputMaskRules" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask." />
```
ralfhandl, line 1420:
```suggestion
          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask" />
```

ralfhandl:
Almost there... 😄

vocabularies/UI.xml
ralfhandl: I like the example and wonder whether we can keep it Germany-centric or whether we should generalize it according to https://en.wikipedia.org/wiki/International_Bank_Account_Number

> The IBAN consists of up to 34 alphanumeric characters, as follows:
>
> - country code using [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) – two letters,
> - [check digits](https://en.wikipedia.org/wiki/Check_digit) – two digits, and
> - **Basic Bank Account Number (BBAN)** – up to 30 alphanumeric characters that are country-specific.[[2]](https://en.wikipedia.org/wiki/International_Bank_Account_Number#cite_note-IBANRegistry-2)

BerSie: https://experience.sap.com/internal/fiori-design-web/generic-mask-input/#do-not-use-mask-input-if

BerSie: > Is the control then expected to convert lower case letters to upper case?

yes, it will adapt the input to the mask.

HeikoTheissen, line 1419:
```suggestion
        <Property Name="Rules" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
```
No need to repeat `InputMask`.

HeikoTheissen: Does it parse the regular expression to know that?

HeikoTheissen: > The IBAN consists of up to 34 alphanumeric characters, as follows:

An input mask makes sense only if the length of the input is fixed, not "up to N characters".

BerSie: > An input mask makes sense only if the length of the input is fixed, not "up to N characters".

exactly, that's basically what the Fiori Design Guideline says about it.

BerSie: > Does it parse the regular expression to know that?

I have to correct myself - apparently it prevents the input of lower case letters: https://sapui5.hana.ondemand.com/#/entity/sap.m.MaskInput/sample/sap.m.sample.MaskInput. Honestly, I personally would expect a correction from lower to upper case.

vocabularies/UI.xml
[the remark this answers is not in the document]
BerSie: done

vocabularies/UI.xml
HeikoTheissen: > I personally would expect a correction from lower to upper case.

But this is not expressed in the `InputMask` annotation.

HeikoTheissen: Alternative example:
```json
"SwiftCode@UI.Placeholder": "Has the format `Bank Country Location Branch`",
"SwiftCode@UI.InputMask": {
  "Mask": "CCCC CC AA AAA",
  "Placeholder": ".",
  "Rules": [
    { "MaskSymbol": "C", "RegExp": "[A-Z]" },
    { "MaskSymbol": "A", "RegExp": "[A-Z0-9]" }
  ]
}
```

BerSie: > > I personally would expect a correction from lower to upper case.
> 
> But this is not expressed in the `InputMask` annotation.

I think that's something that the client control has to decide, how it deals with that annotation and the defined rules. Neither I would exclude such a behavior nor do I want to enforce it. It should do something meaningful that the end user expects (something that UX Designer would specify usually).

BerSie: > Alternative example:

this one looks OK to me - the app could fill the last 3 characters if the user doesn't provide it.

HeikoTheissen: > the client control has to decide, how it deals with that annotation and the defined rules

If we expect the client to understand the regular expressions, how does this differ from a [`Validation.Pattern`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Validation.V1.md#Pattern) annotation? The latter is more complex, because our regular expressions only describe one character at a time. But does that justify introducing this separate term?

vocabularies/UI.xml
```diff
@@ -1401,6 +1401,37 @@ The trend is
...
+        <Property Name="Mask" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="The mask to be applied to the property or the parameter" />
+        </Property>
+        <Property Name="Placeholder" Type="Edm.String" MaxLength="1" Nullable="false" DefaultValue="_">
+          <Annotation Term="Core.Description" String="A single character symbol to be shown where the user can type a character" />
+        </Property>
+        <Property Name="Rules" Type="Collection(UI.InputMaskRuleType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Rules that define valid values for one symbol in the mask" />
+          <Annotation Term="Core.LongDescription">
+            <String>The following rules are defined as default and don't need to be listed here: * = ., C = [a-zA-Z], 9 = [0-9]</String>
```
HeikoTheissen, line 1422: Are these rules hardwired into the UI5 control, or can they be overwritten by this annotation? If they are hardwired, we must rephase this as "MUST NOT be listed here".

vocabularies/UI.xml
BerSie: Hmm difficult, I think a general idea (at least on my side) was, to ease the definition of this annotation by having lots of defaults that come out-of-the-box.

HeikoTheissen: But this does not help if these defaults can be overwritten, see https://github.com/SAP/odata-vocabularies/pull/317#discussion_r1593961731.

vocabularies/UI.xml
BerSie: No, it's not hard-wired - the placeholder "_" is a default there.But the rules have to be defined on the control (default is "*" and [a-zA-Z0-9].

HeikoTheissen (approved):

vocabularies/UI.xml
BerSie: I wanted to ease it for the (app) developer :-) - not the client code.

ralfhandl (approved):
@BerSie Feel free to merge

BerSie:
> @BerSie Feel free to merge

Sorry, @ralfhandl , @HeikoTheissen ... the most important thing I forgot again ... it's only experimental

HeikoTheissen (approved):

ralfhandl (approved):

PR #319, opened 2024-06-10
Pdf text direction layout

fb0d3nheimer:
This pull requests introduces a new annotation which indicates if PDF supports setting the text direction-layout for a document.  This is required for supporting RTL (right-to-left)-languages.


vocabularies/PDF.xml
```diff
@@ -109,6 +109,12 @@
             <String>If this is true, treeview output is supported for hierarchical data</String>
           </Annotation>
         </Property>
+       <Property Name="TextDirectionLayout" Type="Edm.Boolean">
```
HeikoTheissen, line 112:
```suggestion
       <Property Name="TextDirectionLayout" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
```

HeikoTheissen (changes requested):

HeikoTheissen:
Shouldn't the other Booleans also get a `DefaultValue="false"`?

fb0d3nheimer:
> Shouldn't the other Booleans also get a `DefaultValue="false"`?

Hi Heiko, you mean Treeview, IANATimezoneFormat, etc? This is not scope of this pull request, but I can fix it in another one.

ralfhandl (approved):

PR #323, opened 2024-06-20
Rename property InputMask.Placeholder

BerSie:
Feedback from developers and app developers indicates that the property name might be misleading (maybe also due the term Placeholder, which is in the vocabulary right above the term InputMask). Proposal was to use PlaceholderSymbol. Looking forward to your feedback.


HeikoTheissen:
I suggest to ask a native speaker.

ralfhandl (approved):
Luckily this is still experimental 😄

nlunets (approved):
It makes to avoid confusion :)

ralfhandl:
@BerSie @nlunets Please merge at your convenience.

PR #324, opened 2024-06-25
Merge cells in table columns

BerSie:
Create an opportunity to define columns in a table where the cells are merged if the adjacent values are the same (in html rowSpan would be used). I decided to go as a starting point with an HTML5 collection of propertyPath that can be applied to UI.LineItem. Imo this new annotation makes only sense for UI.LineItem, and I didn't want to change the term itself.
Other possibilities would be applying the annotation to properties or adding something to the UI.PresentationVariant.


vocabularies/HTML5.xml
```diff
@@ -88,6 +88,14 @@
         </Annotation>
       </TypeDefinition>
 
+      <Term Name="MergeDuplicateValuesInColumns" Type="Collection(Edm.PropertyPath)" Nullable="false">
```
ralfhandl, line 91: I find the "in columns" misleading as this is row-span, so vertically merge adjacent cells with the same value in one columns.

ralfhandl (changes requested):

vocabularies/HTML5.xml
BerSie: Do you have a better name - I changed it definitely 5 times before pushing :-) Currently I try to say that it merges duplicate values in a column - adjacent is missing - but simply adding it, makes the name even longer.

HeikoTheissen (changes requested):
You annotate a `UI.LineItem` with a collection of `Edm.PropertyPath`s. But the line item contains
```
"UI.DataField": {"Value": {"$Path": ...}}
```
and can even combine more than one property.

I suggest to annotate the `UI.DataField` record instead with a Boolean term `HTML5.PreferRowSpan`.

BerSie:
> I suggest to annotate the `UI.DataField` record instead with a Boolean term:
> 
> ```
> <Record Type="UI.DataField">
>   <Annotation Term="HTML5.PreferRowSpan" Bool="true" />
> ```

That sounds indeed interesting - thank you @HeikoTheissen ! The only thing that I'm not 100% convinced of is the name  (esp. the "Prefer"). I will prepare a new commit.

vocabularies/HTML5.xml
```diff
@@ -88,11 +88,13 @@
         </Annotation>
       </TypeDefinition>
 
-      <Term Name="MergeDuplicateValuesInColumns" Type="Collection(Edm.PropertyPath)" Nullable="false">
+      <Term Name="RowSpanForDuplicateValues" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Record">
         <Annotation Term="Common.Experimental" />
-        <Annotation Term="Core.Description" String="For these properties adjacent column cells will be merged if the values are the same." />
+        <Annotation Term="Core.Description" String="The annotated UI.DataField might use a rowSpan for adjacent duplicate values if used in a table." />
```
HeikoTheissen, line 93:
```suggestion
        <Annotation Term="Core.RequiresType" String="UI.DataFieldAbstract" />
        <Annotation Term="Core.Description" String="The annotated data field might use a rowSpan for adjacent duplicate values if used in a table." />
```

HeikoTheissen (changes requested):

HeikoTheissen (approved):

HeikoTheissen (changes requested):
Add reference to UI vocabulary.

ralfhandl (commented):
I like the new term name and where it applies!

HeikoTheissen (approved):

ralfhandl (approved):

PR #329, opened 2024-07-11
Allow manual start of build job

ralfhandl:


BerSie (approved):

HeikoTheissen:
Why is this necessary? The build job runs after every push.

ralfhandl:
> Why is this necessary? The build job runs after every push.

We just had a conflict with `main` after running the build, and I thought running the build again might solve the problem.

HeikoTheissen (approved):

PR #336, opened 2024-07-25
UI.DateTimeStyle

HeikoTheissen:


ralfhandl (approved):

BerSie:
Here are my thoughts: 
the style: this is usually something I would expect to be set by the framework according to UI guidelines and available space. If you have a date in the table with 3 columns, you can of course use the _long_ version, but what if the user adds three more columns?
the pattern: if we open this up to be set per property/parameter, consistency is lost. I'm not sure about the scope of this pattern definition - be it the app, or the user, or all users at a customer site. For this scope I would expect one pattern definition per style (if style is introduced, because we are neither able to define UI guidelines for that nor able to calculate the available space).

HeikoTheissen:
**Feedback from Albatross 2024-08-27**

If a property is thus annotated, it would be shown in the same style everywhere (in the header, in a table, etc.). Is this really intended?

BerSie:
> **Feedback from Albatross 2024-08-27**
> 
> If a property is thus annotated, it would be shown in the same style everywhere (in the header, in a table, etc.). Is this really intended?

For me the visualization should be everywhere consistent, besides the end user immediately understands the difference and why there is this difference (that's something I also tried to express with my statements above).

HeikoTheissen:
One given use case, which is generally considered valid, is a timestamp that must be displayed including its fractional seconds. Unfortunately,
* the "style" does not support this (not even if "full")
* the "pattern" allows far more than that, which endangers UI consistency.

Perhaps rather a tag "ShowFractionalSeconds"?

BerSie:
Hi @HeikoTheissen , we discussed this topic again in our team and consensus was that the information for a semantic style should come from an annotation, whereas if a dedicated and very specific pattern is needed, it has to be solved individually (by either having a appdescriptor setting or using FE building blocks). 
This might now even have consequences on the annotation's name :-).

BerSie (approved):
Thank you, @HeikoTheissen, looks good to me.

BerSie (approved):

ralfhandl (approved):

HeikoTheissen:
Will be merged after Albatross meeting, 2024-09-03T15:00:00+02:00.

PR #337, opened 2024-07-25
Add term "Meta"

hjboth:
Common.Meta shall collect "meta" information about a schema such as a creator identification string, the version number of the creator or a copy right notice.

### DIFF
--- a/vocabularies/Common.xml
+++ b/vocabularies/Common.xml
@@ -1480,6 +1480,15 @@ If the request succeeds the response will contain a JSON body of `Content-Type: 
         <Annotation Term="Core.Description" String="Base URL for WebSocket connections" />
       </Term>
 
+      <Term Name="Meta" Type="Common.MetaType" Nullable="false" AppliesTo="Schema">
+        <Annotation Term="Core.Description" String="Meta attributes of the Schema" />
+      </Term>
+      <ComplexType Name="MetaType">
+        <Property Name="Creator" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Constant and version independent identification of the schema creator" />
+        </Property>
+      </ComplexType>
+
     </Schema>
   </edmx:DataServices>
 </edmx:Edmx>
--- a/vocabularies/Common.xml
+++ b/vocabularies/Common.xml
@@ -1481,9 +1481,11 @@ If the request succeeds the response will contain a JSON body of `Content-Type: 
       </Term>
 
       <Term Name="Meta" Type="Common.MetaType" Nullable="false" AppliesTo="Schema">
+        <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="Meta attributes of the Schema" />
       </Term>
       <ComplexType Name="MetaType">
+        <Annotation Term="Common.Experimental" />
         <Property Name="Creator" Type="Edm.String" Nullable="false">
           <Annotation Term="Core.Description" String="Constant and version independent identification of the schema creator" />
         </Property>
--- a/vocabularies/Common.xml
+++ b/vocabularies/Common.xml
@@ -1487,7 +1487,7 @@ If the request succeeds the response will contain a JSON body of `Content-Type: 
       <ComplexType Name="MetaType">
         <Annotation Term="Common.Experimental" />
         <Property Name="Creator" Type="Edm.String" Nullable="false">
-          <Annotation Term="Core.Description" String="Constant and version independent identification of the schema creator" />
+          <Annotation Term="Core.Description" String="Constant and version-independent identification of the schema creator" />
         </Property>
       </ComplexType>
 
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -1327,7 +1327,7 @@
     "MetaType": {
       "$Kind": "ComplexType",
       "@Common.Experimental": true,
-      "Creator": { "@Core.Description": "Constant and version independent identification of the schema creator" }
+      "Creator": { "@Core.Description": "Constant and version-independent identification of the schema creator" }
     }
   }
 }
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -1316,6 +1316,18 @@
       "@Common.Experimental": true,
       "@Core.IsURL": true,
       "@Core.Description": "Base URL for WebSocket connections"
+    },
+    "Meta": {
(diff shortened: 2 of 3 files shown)


HeikoTheissen (commented):
The comment mentions version number and copyright, which are not present in the current iteration of `MetaType`.

hjboth:
> The comment mentions version number and copyright, which are not present in the current iteration of `MetaType`.

yes that's right, do you prefer to remove these suggestions?

HeikoTheissen:
> yes that's right, do you prefer to remove these suggestions?

The question is whether you want to _add_ properties for version number and copyright to `MetaType`.

hjboth:
> The question is whether you want to _add_ properties for version number and copyright to `MetaType`.

I see. Your initial comment wasn't phrased as a question but a statement. To answer: We don't know what the future will bring, that's why we designed it as record/complex type. 

The root cause for this definition is that it is very hard to support $metadata related issues without knowing which party was responsible for the creation of that API.

We have some customer incidents that report about unclear APIs which are from unclear origin. Magnifying the top 20 lines of a $metadata response png to compare attribute permutations etc. just to find out that the API hasn't been created by CAP isn't effective.

Therefore we need to "watermark" the API and maybe there are further informations necessary in the future like version information or other info that helps to identify the origin of that API.

Since version info is volatile it shouldn't be part of the `creator` property which is assumed to be long term constant. Frequent changes in static text blocks are unfortunate in automated test environments (to avoid unnecessary diffs).

HeikoTheissen:
Summary:
* We start with just the `creator` property, other properties may be added later.
* The `Meta` annotation will be added by CAP, other frameworks may follow later.

Correct?

vocabularies/Common.xml
```diff
@@ -1480,6 +1480,17 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
       </Term>
 
+      <Term Name="Meta" Type="Common.MetaType" Nullable="false" AppliesTo="Schema">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="Meta attributes of the Schema" />
+      </Term>
+      <ComplexType Name="MetaType">
+        <Annotation Term="Common.Experimental" />
+        <Property Name="Creator" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Constant and version-independent identification of the schema creator" />
```
HeikoTheissen, line 1490: How is uniqueness guaranteed? Should we insist that this is a URI?

hjboth: I thought a qualifier would be enough, like: 

      <Annotation Term="Common.Meta" Qualifier="CAP">
        <Record Type="Common.MetaType">
          <PropertyValue Property="Creator" String="SAP CAP Core Data Services"/>
        </Record>
      </Annotation>

hjboth:
> Summary:
> 
> * We start with just the `Creator` property, other properties may be added later. (For the version, there is [`Core.SchemaVersion`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#SchemaVersion).)

No the schema version doesn't indicate the software version of the creator

> * The `Meta` annotation will be added by CAP, other frameworks may follow later.
> 
> Correct?

that's the plan, we still need a sign off by our CPO hence it's experimental.

vocabularies/Common.xml
HeikoTheissen: With that approach, you can have several annotations:
```xml
<Annotation Term="Common.Meta" Qualifier="CAP">
  <Record Type="Common.MetaType">
    <PropertyValue Property="Creator" String="SAP CAP Core Data Services"/>
  </Record>
</Annotation>
<Annotation Term="Common.Meta" Qualifier="RAP">
  <Record Type="Common.MetaType">
    <PropertyValue Property="Creator" String="SAP RAP Core Data Services"/>
  </Record>
</Annotation>
```
Is this intended?

hjboth: yes, if some application wants to eternalize itself, why not...

Admittedly, personally I would prefer a single attribute right in `edmx:DataServices`: low footprint, tamperproof, etc... but I think that's not easily achievable...

hjboth: @HeikoTheissen feel free to come up with a better approach, any ideas are appreciated

HeikoTheissen, line 1490:
```suggestion
          <Annotation Term="Core.IsURL" />
          <Annotation Term="Core.Description" String="Constant and version-independent URI identifying the schema creator" />
```

HeikoTheissen (changes requested):

vocabularies/Common.xml
HeikoTheissen: Identity the creator by a URI instead of a qualifier, see https://github.com/SAP/odata-vocabularies/pull/337#discussion_r1692703125.

hjboth: I don't really see what a difference an URI makes...

HeikoTheissen:
Approved, but let me point out an alternative:
```xml
<Annotation Term="Core.Links">
  <Collection>
    <Record>
      <PropertyValue Property="rel" String="author" />
      <PropertyValue Property="href" String="https://pages.github.tools.sap/cap/docs/about/" />
    </Record>
  </Collection>
</Annotation>
```

hjboth:
> Approved, but let me point out an alternative:
> 
> ```
> <Annotation Term="Core.Links">
>   <Collection>
>     <Record>
>       <PropertyValue Property="rel" String="author" />
>       <PropertyValue Property="href" String="https://pages.github.tools.sap/cap/docs/about/" />
>     </Record>
>   </Collection>
> </Annotation>
> ```

yes indeed, that could be an option if we use the official URI `https://cap.cloud.sap/docs/about/` instead ;) I didn't think of capire...

hjboth:
... and btw I appreciate your considerations 👍

hjboth:
would it make sense to offer both options in parallel like `CreatorURI : URI` and `Creator : String` ?

HeikoTheissen:
> ```cds
> annotation Meta {
>   Creator: String;
>   CreatorURI: URI;
> };
> ```

Technically, a URI is also an `Edm.String`, therefore the property `Creator` can already hold a URI.

hjboth:
After more consideration I like `@Core.Links` better as it is an existing concept. We'll use that one instead of an home brew alternative.

PR #340, opened 2024-08-29
Use `Core.ContentID` in actions

HeikoTheissen:
SAP-specific solution to https://github.com/oasis-tcs/odata-specs/issues/1969


vocabularies/Common.xml
```diff
@@ -1480,6 +1480,63 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+
+Exactly one of `ParameterValue` and `ParameterEntity` and
+exactly one of `ReturnedValue` and `ReturnedEntity` must be given.</String>
+        </Annotation>
+        <Property Name="ParameterValue" Type="Edm.PropertyPath" Nullable="true">
+          <Annotation Term="Core.Description" String="Path to a value in a parameter that may be annotated with `Core.ContentID`" />
+        </Property>
+        <Property Name="ParameterEntity" Type="Edm.NavigationPropertyPath" Nullable="true">
+          <Annotation Term="Core.Description" String="Path to an entity in a parameter that may be annotated with `Core.ContentID`" />
+        </Property>
```
ralfhandl, line 1532: We really should use one property of type AnyPropertyPath here.

Maybe put this in a separate vocabulary? The use case is not that common anyway.

vocabularies/Common.xml
```diff
@@ -1480,6 +1480,63 @@ If the request succeeds the response will contain a JSON body of `Content-Type:
...
+  "dependsOn": [ "1" ],
+  "method": "post",
+  "url": "$I1/Discounts",
+  "body": {
+    "percent": 10
+  }
+}]}
+```
+In the response to the action invocation the sales order item for the sugar is annotated
+with `"@Core.ContentID": "I1"`. The subsequent POST request can reference this item without knowing its key.
```
ralfhandl, line 1522: Please insert the corresponding annotation example here that reflects the prose text.

ralfhandl (changes requested):

vocabularies/Common.xml
HeikoTheissen: The example is already linked:
https://github.com/SAP/odata-vocabularies/blob/8649020a200b208d8df9c04360b0bca770f04ade/vocabularies/Common.xml#L1497

vocabularies/Common.xml
HeikoTheissen: Naming suggestions for the new vocabulary? `com.sap.vocabularies.Batch.v1`?

vocabularies/Common.xml
ralfhandl: That is a batch request example. I need an _annotation_ example.

ralfhandl: Ok, where is the link to the example? And I'd appreciate having that snippet directly next to the batch request example.

ralfhandl (approved):

examples/Batch.CorrespondingContentIDs-sample.xml
```diff
@@ -0,0 +1,45 @@
...
+      </EntityType>
+      <Action Name="CreateSalesOrder" IsBound="true">
+        <Parameter Name="_it" Type="self.SalesQuotation" Nullable="false" />
+        <Parameter Name="items" Type="Collection(self.SalesQuotationItemToBeCopied)" />
+        <ReturnType Type="self.SalesOrder" Nullable="false" />
+        <Annotation Term="Batch.CorrespondingContentIDs">
+          <Collection>
+            <Record>
+              <Annotation Term="Core.Description" String="Correspondence between sales quotation item and sales order item" />
+              <PropertyValue Property="ParameterValue" PropertyPath="items" />
```
BerSie, line 34: where does this "items" come from? It's a PropertyPath, so I would expect somewhere a property "items" - the only thing I see in the example is a parameter.

examples/Batch.CorrespondingContentIDs-sample.xml
```diff
@@ -0,0 +1,45 @@
...
+              <PropertyValue Property="ReturnedValue" NavigationPropertyPath="$ReturnType/Items" />
```
BerSie, line 35: cool: $ReturnType  - I have never seen this notation

examples/Batch.CorrespondingContentIDs-sample.xml
HeikoTheissen: A property path relative to an action starts with with a parameter name or with `$ReturnType`.

PR #343, opened 2024-09-13
Undeprecate `Common.SecondaryKey`

HeikoTheissen:


vocabularies/Common.xml
```diff
@@ -218,15 +218,12 @@
       </Term>
 
       <Term Name="SecondaryKey" AppliesTo="EntityType" Type="Collection(Edm.PropertyPath)" Nullable="false">
```
BerSie, line 220: One of our colleagues mentioned that it would make sense in course of this de-deprecation to call the term SecondaryKeys as it's a collection (analogous to AlternateKeys from the Core vocabulary.

vocabularies/Common.xml
```diff
@@ -218,15 +218,12 @@
...
-            <Record>
-              <PropertyValue Property="Kind" EnumMember="Core.RevisionKind/Deprecated" />
-              <PropertyValue Property="Description" String="Use term `AlternateKeys` from the OASIS Core vocabulary instead" />
-            </Record>
-          </Collection>
+        <Annotation Term="Core.Description" String="The listed properties form a secondary key" />
+        <Annotation Term="Core.LongDescription">
+          <String>Multiple secondary keys are possible using different qualifiers.
+          Unlike [`Core.AlternateKeys`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#AlternateKeys),
+          secondary keys cannot be used to address an entity in a resource path.</String>
```
BerSie, line 225: I would phrase it a bit weaker ... in a sense that this addressing is not required (but I wouldn't mind if we go with your proposal)

BerSie (commented):
... and we may set it to experimental ...

vocabularies/Common.xml
ralfhandl: No, this is a _single_ secondary key with _one or more key properties_ - we do have lots of multi-part keys at SAP.

Multiple secondary keys are differentiated via qualifiers.

vocabularies/Common.xml
ralfhandl, line 225:
```suggestion
          secondary keys need not support addressing an entity in a resource path.</String>
```

ralfhandl (changes requested):

vocabularies/Common.xml
BerSie: Thanks @ralfhandl - now, I see - AlternateKeys is a collection of AlternateKey, which has a collection of properties. Not sure, if we really need it, but I think it's even better to address a specific secondary (or alternate) key using a qualifier. And your proposal for re-phrasing sounds good to me.

ralfhandl (approved):

vocabularies/Common.xml
```diff
@@ -232,15 +232,12 @@
...
-            <Record>
-              <PropertyValue Property="Kind" EnumMember="Core.RevisionKind/Deprecated" />
-              <PropertyValue Property="Description" String="Use term `AlternateKeys` from the OASIS Core vocabulary instead" />
-            </Record>
-          </Collection>
+        <Annotation Term="Core.Description" String="The listed properties form a secondary key" />
+        <Annotation Term="Core.LongDescription">
+          <String>Multiple secondary keys are possible using different qualifiers.
+          Unlike [`Core.AlternateKeys`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#AlternateKeys),
+          secondary keys need not support addressing an entity in a resource path.</String>
```
HeikoTheissen, line 239: What shall the client make of this? How does it know whether it can use a secondary key in a resource path? That's why I wrote "cannot".

HeikoTheissen, line 239: What shall the client make of this? How does it know whether it can use a secondary key in a resource path? That's why I wrote "cannot".
```suggestion
          Unless also annotated as [`Core.AlternateKeys`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#AlternateKeys),
          secondary keys cannot be used to to address an entity in a resource path.</String>
```

ralfhandl: The client cannot know, and we can't either. Maybe the service supports secondary keys in resource paths tomorrow.

ralfhandl (approved):

PR #344, opened 2024-09-13
Grid

BerSie:
first version

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -702,6 +702,40 @@ The trend is
         </Property>
       </ComplexType>
 
+      <Term Name="Grid" Type="UI.GridDefinitionType" Nullable="false" AppliesTo="EntityType">
+        <Annotation Term="UI.ThingPerspective" />
+        <Annotation Term="Core.Description" String="Visualization of a data grid" />
+      </Term>
+      <ComplexType Name="GridDefinitionType">
+        <Property Name="Title" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.Description" String="Title of the grid" />
+          <Annotation Term="Core.IsLanguageDependent" />
+        </Property>
+        <Property Name="Description" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.Description" String="Short description" />
+          <Annotation Term="Core.IsLanguageDependent" />
+        </Property>
+        <Property Name="Columns" Type="Collection(UI.DataFieldGridAbstract)" Nullable="false">
+          <Annotation Term="Core.Description" String="Columns of the grid" />
+        </Property>
+        <Property Name="Rows" Type="Collection(UI.DataFieldGridAbstract)" Nullable="false">
+        </Property>
+        <Property Name="Actions" Type="Collection(UI.DataFieldForActionAbstract)" Nullable="false">
+          <Annotation Term="Core.Description" String="Available actions" />
+        </Property>
+      </ComplexType>
+      <ComplexType Name="DataFieldGridAbstract" BaseType="UI.DataFieldAbstract" Abstract="true">
+        <Annotation Term="Core.Description" String="Abstract type to restrict usage in Grid" />
+      </ComplexType>
+      <ComplexType Name="DataFieldMeasures" BaseType="UI.DataFieldGridAbstract">
+        <Annotation Term="Core.Description" String="A structured piece of data described by an annotation" />
+        <Property Name="Measures" Type="Collection(Edm.PropertyPath)" Nullable="false" />
+      </ComplexType>
+      <ComplexType Name="DataFieldDimension" BaseType="UI.DataFieldGridAbstract">
+        <Annotation Term="Core.Description" String="A structured piece of data described by an annotation" />
+        <Property Name="Dimensions" Type="Edm.PropertyPath" Nullable="false" />
+      </ComplexType>
+
       <Term Name="Chart" Type="UI.ChartDefinitionType" Nullable="false" AppliesTo="EntityType">
         <Annotation Term="UI.ThingPerspective" />
         <Annotation Term="Core.Description" String="Visualization of multiple data points" />
--- a/vocabularies/UI.json
+++ b/vocabularies/UI.json
@@ -601,6 +601,47 @@
         "@Core.Description": "Name of the Action on the Semantic Object. If not specified, let user choose which of the available actions to trigger."
       }
     },
+    "Grid": {
+      "$Kind": "Term",
+      "$Type": "UI.GridDefinitionType",
+      "$AppliesTo": ["EntityType"],
+      "@UI.ThingPerspective": true,
+      "@Core.Description": "Visualization of a data grid"
+    },
+    "GridDefinitionType": {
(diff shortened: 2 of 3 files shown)


vocabularies/UI.xml
```diff
@@ -702,6 +702,40 @@ The trend is
...
+          <Annotation Term="Core.Description" String="Available actions" />
+        </Property>
+      </ComplexType>
+      <ComplexType Name="DataFieldGridAbstract" BaseType="UI.DataFieldAbstract" Abstract="true">
+        <Annotation Term="Core.Description" String="Abstract type to restrict usage in Grid" />
+      </ComplexType>
+      <ComplexType Name="DataFieldMeasures" BaseType="UI.DataFieldGridAbstract">
+        <Annotation Term="Core.Description" String="A structured piece of data described by an annotation" />
+        <Property Name="Measures" Type="Collection(Edm.PropertyPath)" Nullable="false" />
+      </ComplexType>
```
HeikoTheissen, line 733: One `DataFieldMeasures` already contains several measures, and in addition to that you can have several `DataFieldMeasures`s in `UI.Grid/Columns`. Why these nested collections?

HeikoTheissen (commented):
Please also consider the annotations suggested in https://github.tools.sap/D024504/grid/issues/1

vocabularies/UI.xml
```diff
@@ -702,6 +702,40 @@ The trend is
         </Property>
       </ComplexType>
 
+      <Term Name="Grid" Type="UI.GridDefinitionType" Nullable="false" AppliesTo="EntityType">
```
ralfhandl, line 705:
```suggestion
      <Term Name="DataGrid" Type="UI.DataGridType" Nullable="false" AppliesTo="EntityType">
```
We usually use the term name with suffix `Type`

vocabularies/UI.xml
```diff
@@ -702,6 +702,40 @@ The trend is
...
+        <Annotation Term="UI.ThingPerspective" />
+        <Annotation Term="Core.Description" String="Visualization of a data grid" />
```
ralfhandl, line 707: If it is i a "data grid", why not use that as term name?

vocabularies/UI.xml
```diff
@@ -702,6 +702,40 @@ The trend is
...
+      </Term>
+      <ComplexType Name="GridDefinitionType">
```
ralfhandl, line 709:
```suggestion
      <ComplexType Name="DataGridType">
```

vocabularies/UI.xml
```diff
@@ -702,6 +702,40 @@ The trend is
...
+      <ComplexType Name="DataFieldGridAbstract" BaseType="UI.DataFieldAbstract" Abstract="true">
+        <Annotation Term="Core.Description" String="Abstract type to restrict usage in Grid" />
+      </ComplexType>
+      <ComplexType Name="DataFieldMeasures" BaseType="UI.DataFieldGridAbstract">
+        <Annotation Term="Core.Description" String="A structured piece of data described by an annotation" />
+        <Property Name="Measures" Type="Collection(Edm.PropertyPath)" Nullable="false" />
+      </ComplexType>
+      <ComplexType Name="DataFieldDimension" BaseType="UI.DataFieldGridAbstract">
+        <Annotation Term="Core.Description" String="A structured piece of data described by an annotation" />
+        <Property Name="Dimensions" Type="Edm.PropertyPath" Nullable="false" />
```
ralfhandl, line 736:
```suggestion
        <Property Name="Dimension" Type="Edm.PropertyPath" Nullable="false" />
```
Complex type name is singular, and type is also just one property path

ralfhandl (changes requested):

ralfhandl:
What actually is a "data grid", and why isn't it just another ChartType?

https://github.com/SAP/odata-vocabularies/blob/35c70803cbe5ce3f00f089a0eaf2a8fae27afc98/vocabularies/UI.xml#L793-L800

vocabularies/UI.xml
BerSie: That's not intented - thanks for the hint

vocabularies/UI.xml
BerSie: good idea

vocabularies/UI.xml
BerSie: True, I missed that, when I changed the type name from plural to singular

BerSie:
> What actually is a "data grid", and why isn't it just another ChartType?
> 
> https://github.com/SAP/odata-vocabularies/blob/35c70803cbe5ce3f00f089a0eaf2a8fae27afc98/vocabularies/UI.xml#L793-L800

I'm not an expert on that, but a grid has rows and columns which can be flexibly arranged using measures and dimensions (and there is something in addition we left out in the first version, which seems to be called structural properties, which are filter conditions on meaasures - if I got that correctly). So, simply a new chart type wouldn't suffice these requirements.

vocabularies/UI.xml
HeikoTheissen: But you haven't changed it yet.

BerSie: there can be only one datafieldmeasures either in columns or in rows. I'm struggling with the definition of the rows in columns as there either can dimensions and measures there,  but measures can either be in the rows or in columns and there are always all together, dimensions can be spread over columns and rows. And of course the measure collection doesn't have a fixed position :-).

BerSie: > But you haven't changed it yet.

Because I had to think about it :-) - I'm not as fast as you.

BerSie: > > But you haven't changed it yet.
> 
> Because I had to think about it :-) - I'm not as fast as you.

And if you have an idea how to solve the dilemma, don't hesitate to publish it...

ralfhandl: > there can be only one datafieldmeasures either in columns or in rows

Sounds complicated. Do you have examples, ideally with visuals of the corresponding grids?

HeikoTheissen:
> What actually is a "data grid", and why isn't it just another ChartType?

A grid cannot show its full power by making OData requests _at runtime_, it needs InA for that. The purpose of this pull request is only to define a grid through OData annotations (_design time_), not offer an OData service for the runtime.

This could be expressed by having an annotated entity type in the $metadata, but no entity container.

vocabularies/UI.xml
BerSie: I will send you (both) an email with a screen shot.

HeikoTheissen: > there can be only one datafieldmeasures either in columns or in rows. I'm struggling with the definition of the rows in columns as there either can dimensions and measures there, but measures can either be in the rows or in columns and there are always all together, dimensions can be spread over columns and rows. And of course the measure collection doesn't have a fixed position :-).

Let the `UI.Grid` have properties
* `Rows`, which can contain `UI.DataField`s with groupable values and `UI.ReferenceFacet`s pointing to `Analytics.StructureElements`
* `Columns`, which can contain the same things
* `Measures`, which can contain `UI.DataField`s with an aggregatable value
* a property `MeasuresAxis`, which is either `Rows` or `Columns`.

PR #348, opened 2024-10-01
Feat/dpp extension

sthesse:
Extending DPP Annotations by 3 additional records.


ralfhandl (dismissed):

vocabularies/PersonalData.xml
```diff
@@ -166,7 +166,7 @@ Terms for contact and address information are defined in the [Communication voca
               <PropertyValue Property="Value" String="LegalEntityID" />
               <Annotation Term="Core.Description" String="The unique identifier of a legal entity" />
               <Annotation Term="Core.LongDescription">
-                <String>A legal entity is a corporation, an association, or any other organization of legal capacity, which has statutory rights and responsibilities.</String>
+                <String>(To be deprecated) A legal entity is a corporation, an association, or any other organization of legal capacity, which has statutory rights and responsibilities.</String>
```
HeikoTheissen, line 169: Why not use the `Core.Revisions` annotation to deprecate this?

HeikoTheissen (dismissed):

BerSie (commented):
It would really help me to have an example for the FieldSemantics term. If I read the description it all makes sense to me, but I really struggle to understand how a resp. annotation will look like and how the consumer is expected to react on it.

HeikoTheissen (approved):

vocabularies/PersonalData.xml
```diff
@@ -162,12 +162,19 @@ Terms for contact and address information are defined in the [Communication voca
...
                 <String>A legal entity is a corporation, an association, or any other organization of legal capacity, which has statutory rights and responsibilities.</String>
               </Annotation>
+              <Annotation Term="Core.Revisions">
+                <Collection>
+                  <Record>
+                    <PropertyValue Property="Kind" EnumMember="Core.RevisionKind/Deprecated" />
+                    <PropertyValue Property="Description" String="Deprecated in favor of [`DataControllerID`](#DataControllerID)" />
+                  </Record>
+                </Collection>
+              </Annotation>
```
HeikoTheissen, line 177: Since that value was experimental, you need not deprecate it but can simply delete it (if you don't want to keep it for reference).

sthesse: I would like to keep it for reference to avoid breaking any unknown integrations using this value...

PR #354, opened 2024-10-29
Add Node 22

ralfhandl:
Node 22 is now current, Node 20 is active LTS, see https://github.com/nodejs/release?tab=readme-ov-file#release-schedule

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [20.x, 22.x]
+        node-version: [22.x]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [20.x]
+        node-version: [20.x, 22.x]
```
HeikoTheissen, line 11:
```suggestion
        node-version: [22.x]
```

HeikoTheissen (changes requested):
No need to run the build twice.

HeikoTheissen (dismissed):

HeikoTheissen:
The build runs even if we do not specify a Node version. And if we do not require a specific one before merging, that should be sufficient.

HeikoTheissen (approved):

PR #355, opened 2024-11-15
UI.TextArrangement on properties

HeikoTheissen:


BerSie:
why do we introduce this additional AppliesTo ? A TextArrangement cannot (should not) live without a text annotation.

PR #356, opened 2024-11-15
Add identifier to action groups

BerSie:
For the consumers there is a need to identify action groups uniquely. Similar to CollectionFacets there is nothing that would uniquely identify this group/collection. If it would be the concatenation of all entries, already moving or removing one entry would change the ID (although the semantics of the group doesn't change). If there are others idea to achieve this unique identification of an action group, pls. let me know - I would be really interested in.

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -1745,6 +1745,9 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
       <ComplexType Name="DataFieldForActionGroup" BaseType="UI.DataFieldAbstract">
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="Collection of OData actions and intent based navigations" />
+        <Property Name="ID" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.Description" String="Unique identifier of an action group. ID should be stable, as long as the perceived semantics of the action group is unchanged." />
+        </Property>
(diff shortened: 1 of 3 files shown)


vocabularies/UI.xml
```diff
@@ -1745,6 +1745,9 @@ It is NOT tied to a data value (in contrast to [DataFieldWithIntentBasedNavigati
       <ComplexType Name="DataFieldForActionGroup" BaseType="UI.DataFieldAbstract">
         <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="Collection of OData actions and intent based navigations" />
+        <Property Name="ID" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.Description" String="Unique identifier of an action group. ID should be stable, as long as the perceived semantics of the action group is unchanged." />
```
HeikoTheissen, line 1749: Unique within what? Globally? Within the annotated service document?

HeikoTheissen (changes requested):

vocabularies/UI.xml
BerSie: Unique is indeed weird in the context of annotations and metadata. The respective UI thing that's related to the annotation has to be uniquely identified. Identifier is enough here. I will adapt the description.

HeikoTheissen (approved):

ralfhandl (approved):

MarcelWaechter (approved):

PR #359, opened 2024-12-06
Add facets to operations to group parameters

BerSie:
As there seem to exist actions with large number of parameters, these should get a kind of structure now. I've seen that UI.FieldGroup applies already to Action / Function, but I cannot recognize how this could lead to a meaningful structure / order. Therefore I introduce facets as a structure element.

### DIFF
--- a/vocabularies/UI.xml
+++ b/vocabularies/UI.xml
@@ -994,6 +994,7 @@ The trend is
       </Term>
 
       <Term Name="OperationParameterFacets" Type="Collection(UI.ReferenceFacet)" Nullable="false" AppliesTo="Action Function ActionImport FunctionImport">
+        <Annotation Term="Common.Experimental" />
         <Annotation Term="Core.Description" String="Facets that reference UI.FieldGroup annotations to group action or function parameters" />
       </Term>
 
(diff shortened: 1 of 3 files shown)


vocabularies/UI.xml
```diff
@@ -993,6 +993,11 @@ The trend is
         <Annotation Term="Core.Description" String="Facets that reference UI.FieldGroup annotations to group filterable fields" />
       </Term>
 
+      <Term Name="OperationParameterFacets" Type="Collection(UI.ReferenceFacet)" Nullable="false" AppliesTo="Action Function ActionImport FunctionImport">
```
ralfhandl, line 996: Please add `ActionImport` to the `AppliesTo` list of `FieldGroup` here:

https://github.com/SAP/odata-vocabularies/blob/4b48175db142914aa723129920ca58bcebb08c0f/vocabularies/UI.xml#L157

ralfhandl (changes requested):
+1, one comment.

vocabularies/UI.xml
BerSie: done

ralfhandl (approved):

HeikoTheissen (approved):

PR #360, opened 2024-12-19
Common.ErrorResolution/ErrorLogUrl

HeikoTheissen:


uhlmannm (commented):
As Core.Links seems to suffice, we should not introduce another annotation.
"AdditionalNote" would not be required right now, if I got the mails right.
"ErrorLogUrl" is not necessary if Core.Links can be used instead.

PR #362, opened 2025-01-21
DataField representing PresentationVariant

HeikoTheissen:
A `UI.DataFieldForAnnotation` can be a `UI.Chart`, which is used to fill a table column with mini-charts. To define a sort order or a filter for these mini-charts, a `UI.PresentationVariant` or `UI.SelectionPresentationVariant` is needed, in analogy with `UI.Facet`:
https://github.com/SAP/odata-vocabularies/blob/85d8c0c5eda33dfd159e11557873455ef326fc29/vocabularies/UI.xml#L1044-L1046

### DIFF
--- a/vocabularies/UI.json
+++ b/vocabularies/UI.json
@@ -1507,7 +1507,9 @@
           "UI.Chart",
           "UI.ConnectedFields",
           "UI.DataPoint",
-          "UI.FieldGroup"
+          "UI.FieldGroup",
+          "UI.PresentationVariant",
+          "UI.SelectionPresentationVariant"
         ]
       }
     },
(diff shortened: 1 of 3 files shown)


BerSie (approved):

MN-public:
@HeikoTheissen When can we expect it to be merged?

HeikoTheissen:
> @HeikoTheissen When can we expect it to be merged?

Next Tuesday, after the Albatross meeting.

HeikoTheissen:
The merge has been postponed by a week, because we have not yet seen the use of such an annotation in a real Fiori app. @MN-public, can you point us to an example?

MN-public:
Hi Heiko, unfortunately, I do not have an overview of the existing apps to provide you an example. However, this is a new functionality added in the recent UI5 version and IMO it makes sense to provide the tool support for it even before the first public app is using it. For support in tools, such as annotation LSP, we rely on the vocabulary definitions.
We can continue the discussion using internal channels

PR #365, opened 2025-02-04
Prefer change requests via navigation paths

HeikoTheissen:
Shall we also give possible reasons for such a preference (performance, message handling)?


uhlmannm:
> Shall we also give possible reasons for such a preference (performance, message handling)?

I would not add the reasoning.

beckermarc (approved):

ThomasChadzelek (approved):

martin4all (approved):

PR #375, opened 2025-03-14
ValueListParameterIn, ValueListParameterInOut: Add operator parameter

PierreFritsch:
We have the requirement of pre-filtering a value help programmatically to preselect valid values for the end users: 

- *Valid From* shall be *on or after* some date
- *Valid To* shall be *before or on* some date

Screenshot 2025-03-14 at 12 00 59

Currently, the OData vocabulary for `ValueListParameterIn` and `ValueListParameterInOut` only supports the `eq` (Equal to) operator a.k.a. `SelectionRangeOptionType`.  Suggesting we extend it so that it supports at least `le` (Less than or equal to) and `ge` (Greater than or equal to).

While these two options would be sufficient for our specific use case, we noticed that the enumeration of selection options contains the operators `bt` (Between) and `nb` (Not between). Supporting all operators would mean that a second value would need to be added to the value list parameters.

Note that this is our first contribution to this repository and that we have a limited capacity for contributing, so feel free to edit/adjust the proposal. Thanks for your comprehension!

### DIFF
--- a/vocabularies/Common.xml
+++ b/vocabularies/Common.xml
@@ -795,8 +795,15 @@ The minimum structure is
         </Property>
       </ComplexType>
       <ComplexType Name="ValueListParameterIn" BaseType="Common.ValueListParameter">
+        <Property Name="Operator" Type="UI.SelectionRangeOptionType" DefaultValue="0" Nullable="false">
+          <Annotation Term="Core.Description" String="Operator applied to the property. Defaults to `EQ`." />
+        </Property>
         <Property Name="LocalDataProperty" Type="Edm.PropertyPath" Nullable="false">
-          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list with `eq` comparison" />
+          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list by applying the operator (Single value or lower interval boundary)" />
+          <Annotation Term="Core.LongDescription" String="In case the property path contains a collection-based navigation or structural property, the filter is a set of `eq` comparisons connected by `or` operators" />
+        </Property>
+        <Property Name="LocalDataPropertyHigh" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list by applying the operator (Upper interval boundary)" />
           <Annotation Term="Core.LongDescription" String="In case the property path contains a collection-based navigation or structural property, the filter is a set of `eq` comparisons connected by `or` operators" />
         </Property>
         <Property Name="InitialValueIsSignificant" Type="Edm.Boolean" DefaultValue="false" Nullable="false">
@@ -822,8 +829,14 @@ The minimum structure is
         </Property>
       </ComplexType>
       <ComplexType Name="ValueListParameterInOut" BaseType="Common.ValueListParameter">
+        <Property Name="Operator" Type="UI.SelectionRangeOptionType" DefaultValue="0" Nullable="false">
+          <Annotation Term="Core.Description" String="Operator applied to the property. Defaults to `EQ`." />
+        </Property>
         <Property Name="LocalDataProperty" Type="Edm.PropertyPath" Nullable="false">
-          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list with `startswith` comparison and filled from the picked value list item" />
+          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list by applying the operator, and filled from the picked value list item  (Single value or lower interval boundary)" />
+        </Property>
+        <Property Name="LocalDataPropertyHigh" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Path to the property that is used to filter the value list by applying the operator, and filled from the picked value list item (Upper interval boundary)" />
         </Property>
         <Property Name="InitialValueIsSignificant" Type="Edm.Boolean" DefaultValue="false" Nullable="false">
           <Annotation Term="Core.Description" String="Initial value, e.g. empty string, is a valid and significant value" />
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -698,9 +698,19 @@
     "ValueListParameterIn": {
       "$Kind": "ComplexType",
       "$BaseType": "Common.ValueListParameter",
+      "Operator": {
+        "$Type": "UI.SelectionRangeOptionType",
+        "$DefaultValue": 0,
+        "@Core.Description": "Operator applied to the property. Defaults to `EQ`."
+      },
       "LocalDataProperty": {
         "$Type": "Edm.PropertyPath",
-        "@Core.Description": "Path to property that is used to filter the value list with `eq` comparison",
+        "@Core.Description": "Path to property that is used to filter the value list by applying the operator (Single value or lower interval boundary)",
+        "@Core.LongDescription": "In case the property path contains a collection-based navigation or structural property, the filter is a set of `eq` comparisons connected by `or` operators"
+      },
+      "LocalDataPropertyHigh": {
+        "$Type": "Edm.PropertyPath",
+        "@Core.Description": "Path to property that is used to filter the value list by applying the operator (Upper interval boundary)",
         "@Core.LongDescription": "In case the property path contains a collection-based navigation or structural property, the filter is a set of `eq` comparisons connected by `or` operators"
       },
       "InitialValueIsSignificant": {
@@ -737,9 +747,18 @@
     "ValueListParameterInOut": {
       "$Kind": "ComplexType",
       "$BaseType": "Common.ValueListParameter",
+      "Operator": {
+        "$Type": "UI.SelectionRangeOptionType",
+        "$DefaultValue": 0,
+        "@Core.Description": "Operator applied to the property. Defaults to `EQ`."
+      },
       "LocalDataProperty": {
         "$Type": "Edm.PropertyPath",
-        "@Core.Description": "Path to property that is used to filter the value list with `startswith` comparison and filled from the picked value list item"
+        "@Core.Description": "Path to property that is used to filter the value list by applying the operator, and filled from the picked value list item  (Single value or lower interval boundary)"
+      },
+      "LocalDataPropertyHigh": {
+        "$Type": "Edm.PropertyPath",
+        "@Core.Description": "Path to the property that is used to filter the value list by applying the operator, and filled from the picked value list item (Upper interval boundary)"
       },
       "InitialValueIsSignificant": {
         "$Type": "Edm.Boolean",
--- a/vocabularies/Common.md
+++ b/vocabularies/Common.md
@@ -68,52 +68,52 @@ Term|Type|Description
 [ValueListForValidation](Common.xml#L746)|String|<a name="ValueListForValidation"></a>Contains the qualifier of the ValueList or ValueListMapping that should be used for validation<br>An empty string identifies the ValueList or ValueListMapping without a qualifier.
 [ValueListReferences](Common.xml#L751)|\[URL\]|<a name="ValueListReferences"></a>A list of URLs of CSDL documents containing value list mappings for this parameter or property
 [ValueListMapping](Common.xml#L756)|[ValueListMappingType](#ValueListMappingType)|<a name="ValueListMapping"></a>Specifies the mapping between data service properties and value list properties<br>The value list can be filtered based on user input. It can be used for type-ahead and classical pick lists. There may be many alternative mappings with different qualifiers.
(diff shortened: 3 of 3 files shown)


PierreFritsch:
CC @d027132 FYI (See also the internal feature request FIORITECHE1-8621.)

CC @D041453 @jan-philip-zieher FYI

vocabularies/Common.xml
```diff
@@ -795,8 +795,15 @@ The minimum structure is
         </Property>
       </ComplexType>
       <ComplexType Name="ValueListParameterIn" BaseType="Common.ValueListParameter">
+        <Property Name="Operator" Type="UI.SelectionRangeOptionType" DefaultValue="0" Nullable="false">
```
HeikoTheissen, line 798: Instead of extending the existing type, define a new subtype (suggested name `ValueListRangeIn`) of `ValueListParameterIn` and add the new properties to that subtype only.

vocabularies/Common.xml
```diff
@@ -795,8 +795,15 @@ The minimum structure is
...
+          <Annotation Term="Core.Description" String="Operator applied to the property. Defaults to `EQ`." />
+        </Property>
         <Property Name="LocalDataProperty" Type="Edm.PropertyPath" Nullable="false">
-          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list with `eq` comparison" />
+          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list by applying the operator (Single value or lower interval boundary)" />
+          <Annotation Term="Core.LongDescription" String="In case the property path contains a collection-based navigation or structural property, the filter is a set of `eq` comparisons connected by `or` operators" />
+        </Property>
+        <Property Name="LocalDataPropertyHigh" Type="Edm.PropertyPath" Nullable="false">
```
HeikoTheissen, line 805: Shouldn't this be nullable? For `Operator = CP` there is no upper boundary.

vocabularies/Common.xml
```diff
@@ -795,8 +795,15 @@ The minimum structure is
...
+          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list by applying the operator (Upper interval boundary)" />
           <Annotation Term="Core.LongDescription" String="In case the property path contains a collection-based navigation or structural property, the filter is a set of `eq` comparisons connected by `or` operators" />
```
HeikoTheissen, line 807: A collection of values for the upper boundary makes no sense.

vocabularies/Common.xml
```diff
@@ -822,8 +829,14 @@ The minimum structure is
         </Property>
       </ComplexType>
       <ComplexType Name="ValueListParameterInOut" BaseType="Common.ValueListParameter">
+        <Property Name="Operator" Type="UI.SelectionRangeOptionType" DefaultValue="0" Nullable="false">
```
HeikoTheissen, line 832: Define new subtype `ValueListRangeInOut` of `ValueListParameterInOut`.

HeikoTheissen (changes requested):
You suggest a very broad generalization of the value help concept but give only a narrow use case as justification:
* Valid From shall be on or after some date
* Valid To shall be before or on some date

Do you really need the general case? And don't you need the rule "valid from shall be on or before valid to"? (This seems natural to me, but is not covered by your suggestion.)

jan-philip-zieher:
> You suggest a very broad generalization of the value help concept but give only a narrow use case as justification:
> 
> * Valid From shall be on or after some date
> * Valid To shall be before or on some date
> 
> Do you really need the general case? And don't you need the rule "valid from shall be on or before valid to"? (This seems natural to me, but is not covered by your suggestion.)

@HeikoTheissen thanks for your comments, I pushed how I thought it could be incorporated, please review again.

vocabularies/Common.xml
jan-philip-zieher: what is meant by operator CP, I have only them https://sapui5.hana.ondemand.com/#/api/sap.ui.model.FilterOperator in mind

vocabularies/Common.xml
jan-philip-zieher: maybe we have here a misunderstanding ? this parameter only needs to be set if a developer wants to default the value to between, although I also don't have a meaningful use case for that, but when using the FE UI you are technically able to configure such scenario, therefore the vocabulary standard should not prohibit us to do so
image

vocabularies/Common.xml
PierreFritsch: CP = Contains Pattern

https://github.com/SAP/odata-vocabularies/blob/9c2dc33ba9913d8f1a2d41b3f8223ce204c598e7/vocabularies/UI.xml#L1337-1339

vocabularies/Common.xml
```diff
@@ -795,10 +795,17 @@ The minimum structure is
         </Property>
       </ComplexType>
       <ComplexType Name="ValueListParameterIn" BaseType="Common.ValueListParameter">
+        <Property Name="Operator" Type="Edm.PrimitiveType" DefaultValue="EQ" Nullable="true">
```
HeikoTheissen, line 798: Use a concrete type that lists all allowed values (define one if necessary).

vocabularies/Common.xml
```diff
@@ -795,10 +795,17 @@ The minimum structure is
...
+          <Annotation Term="Core.Description" String="Operator applied to the property. Defaults to `EQ`." />
+        </Property>
         <Property Name="LocalDataProperty" Type="Edm.PropertyPath" Nullable="false">
-          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list with `eq` comparison" />
+          <Annotation Term="Core.Description" String="Path to property that is used to filter the value list by applying the operator (Single value or lower interval boundary)" />
           <Annotation Term="Core.LongDescription" String="In case the property path contains a collection-based navigation or structural property, the filter is a set of `eq` comparisons connected by `or` operators" />
         </Property>
+        <Property Name="LocalDataPropertyBT" Type="Edm.PropertyPath" Nullable="true">
```
HeikoTheissen, line 805: As I said before, do not clutter the type with additional properties. Define a new subtype.

HeikoTheissen (changes requested):
First clarify if the requirement is needed in this generality.
- Does the UI support this?
- Are there alternatives (https://github.com/SAP/odata-vocabularies/pull/375#issuecomment-2724695989)?

vocabularies/Common.xml
```diff
@@ -803,6 +803,11 @@ The minimum structure is
           <Annotation Term="Core.Description" String="Initial value, e.g. empty string, is a valid and significant value" />
         </Property>
       </ComplexType>
+      <ComplexType Name="ValueListRangeIn" BaseType="Common.ValueListParameterIn">
+        <Property Name="Operator" Type="UI.ValueListRangeOperator" DefaultValue="EQ" Nullable="false">
```
PierreFritsch, line 807: Is this the proper way of passing an `Enum` value as `DefaultValue`? (Couldn't find other similar samples in the repo, only `true` or `false`.)

PierreFritsch:
> You suggest a very broad generalization of the value help concept but give only a narrow use case as justification:
> 
> * Valid From shall be on or after some date
> * Valid To shall be before or on some date
> 
> Do you really need the general case? And don't you need the rule "valid from shall be on or before valid to"? (This seems natural to me, but is not covered by your suggestion.)

Thanks for the feedback @HeikoTheissen. Reduced the change to our concrete use case.

vocabularies/Common.xml
PierreFritsch: => 2e24e35

vocabularies/Common.xml
PierreFritsch: Removed the BT use case and the upper boundary with 2e24e35

vocabularies/Common.xml
PierreFritsch: Removed the *Between* use case and the upper boundary with 2e24e35.

vocabularies/Common.xml
PierreFritsch: => https://github.com/SAP/odata-vocabularies/commit/2e24e353bd0f15f9e82d3adc4244e098f2f2cea1

vocabularies/Common.xml
PierreFritsch: => https://github.com/SAP/odata-vocabularies/commit/2e24e353bd0f15f9e82d3adc4244e098f2f2cea1

vocabularies/UI.xml
```diff
@@ -1360,6 +1360,18 @@ The trend is
         </Member>
       </EnumType>
 
+      <EnumType Name="ValueListRangeOperator">
```
HeikoTheissen, line 1363: Why is this in UI if it is only used in Common?

vocabularies/Common.xml
HeikoTheissen, line 807:
```suggestion
        <Property Name="Operator" Type="UI.ValueListRangeOperator" DefaultValue="UI.ValueListRangeOperator/EQ" Nullable="false">
```

vocabularies/Common.xml
```diff
@@ -834,6 +839,11 @@ The minimum structure is
           </Collection>
         </Annotation>
       </ComplexType>
+      <ComplexType Name="ValueListRangeInOut" BaseType="Common.ValueListParameterInOut">
+        <Property Name="Operator" Type="UI.ValueListRangeOperator" DefaultValue="EQ" Nullable="false">
```
HeikoTheissen, line 843:
```suggestion
        <Property Name="Operator" Type="UI.ValueListRangeOperator" DefaultValue="UI.ValueListRangeOperator/EQ" Nullable="false">
```

vocabularies/Common.xml
```diff
@@ -803,6 +803,11 @@ The minimum structure is
           <Annotation Term="Core.Description" String="Initial value, e.g. empty string, is a valid and significant value" />
         </Property>
       </ComplexType>
+      <ComplexType Name="ValueListRangeIn" BaseType="Common.ValueListParameterIn">
+        <Property Name="Operator" Type="UI.ValueListRangeOperator" DefaultValue="EQ" Nullable="false">
+          <Annotation Term="Core.Description" String="Operator applied to the property." />
```
HeikoTheissen, line 808:
```suggestion
          <Annotation Term="Core.Description" String="Comparison operator used to filter the value list by the property." />
```

vocabularies/Common.xml
```diff
@@ -834,6 +839,11 @@ The minimum structure is
           </Collection>
         </Annotation>
       </ComplexType>
+      <ComplexType Name="ValueListRangeInOut" BaseType="Common.ValueListParameterInOut">
+        <Property Name="Operator" Type="UI.ValueListRangeOperator" DefaultValue="EQ" Nullable="false">
+          <Annotation Term="Core.Description" String="Operator applied to the property." />
```
HeikoTheissen, line 844:
```suggestion
          <Annotation Term="Core.Description" String="Comparison operator used to filter the value list by the property." />
```

vocabularies/Common.xml
```diff
@@ -803,6 +803,11 @@ The minimum structure is
           <Annotation Term="Core.Description" String="Initial value, e.g. empty string, is a valid and significant value" />
         </Property>
       </ComplexType>
+      <ComplexType Name="ValueListRangeIn" BaseType="Common.ValueListParameterIn">
```
HeikoTheissen, line 806: The subtype is still called "range" but the functional scope is restricted to EQ, LE and GE. Are there plans to extend this in the future to CP and BT? If not, consider
- changing the type name
- whether a subtype is really needed.

One additional property whose default value reflects the current behavior does not necessitate a subtype. I asked for a subtype when it still had many more properties.

HeikoTheissen (changes requested):
The main question is still: Does the UI support the new functionality?

vocabularies/UI.xml
PierreFritsch: Initially placed it here, as it felt like a subset of `SelectionRangeOptionType` above to me. Moved it to `Common` (and renamed to `ValueListParameterOperator`) with 7ff504d7eecf8e1236e872df202ab8ed4d2f4a6f.

vocabularies/Common.xml
PierreFritsch: 6a4c5f0c2f8349aadee26ea7393f7646de2088e7

vocabularies/Common.xml
PierreFritsch: Thx for the hint! 395a11a54059513ee6d8bed10c5cabf68de56832

vocabularies/Common.xml
PierreFritsch: 6a4c5f0c2f8349aadee26ea7393f7646de2088e7

vocabularies/Common.xml
PierreFritsch: https://github.com/SAP/odata-vocabularies/pull/375/files#r2007450478

PR #383, opened 2025-04-04
ShareAction/IsDeltaUpdate

HeikoTheissen:


martin4all (approved):

docs/share-action.md
```diff
@@ -0,0 +1,28 @@
...
+
+```xml
+<ComplexType Name="DraftUserAccessType">
+  <Property Name="UserID" Type="Edm.String" Nullable="false" MaxLength="12"/>
+  <Property Name="UserAccessRole" Type="Edm.String" Nullable="false" MaxLength="1"/>
+</ComplexType>
+<Action Name="Share" IsBound="true">
+  <Parameter Name="_it" Type="com.sap.gateway.srvd.i_shareddemoaction_sd.v0001.SharedEntityType" Nullable="false"/>
+  <Parameter Name="ShareAll" Type="Edm.Boolean" Nullable="false"/>
+  <Parameter Name="IsDeltaUpdate" Type="Edm.Boolean" Nullable="false"/>
```
HeikoTheissen, line 25:
```suggestion
  <Parameter Name="ShareAll" Type="Edm.Boolean" Nullable="false" DefaultValue="false"/>
  <Parameter Name="IsDeltaUpdate" Type="Edm.Boolean" Nullable="false" DefaultValue="false"/>
```
The vocabulary says that these parameters can be omitted, therefore they need a default value.

docs/share-action.md
```diff
@@ -0,0 +1,28 @@
...
+  <Parameter Name="Users" Type="Collection(com.sap.gateway.srvd.i_shareddemoaction_sd.v0001.DraftUserAccessType)" Nullable="false"/>
```
HeikoTheissen, line 26: Why does the parameter order differ from the one in the text?

docs/share-action.md
```diff
@@ -0,0 +1,129 @@
...
+Rule 1: A change in sharing (i.e. exclusive vs. share all vs. collaborative) is only allowed by
+- the draft owner or
+- all (admin, UserRole=Owner) users in the user list.
+
+Rule 2: A call of the share action to change the user list is only allowed
+- in case of ShareAll = true
+  - by a user in the list
+  - by the current user if (a) it is a delta update and (b) users is empty or only contains the current user
+- in case of ShareAll = false
+  - by a user in the list
```
HeikoTheissen, line 129: Missing case ShareAll = null

martin4all: ShareAll = null is not supported (nullable=false). Nevertheless we distinguish between "not provided" (switch to exclusive draft) and "provided" (collaborative draft w/ and w/o explicit invide)

HeikoTheissen: "Not provided" is possible only if
- either the parameter is nullable (omitted nullable parameters are treated as null)
- or the parameter is marked as optional.

Do you prefer the second alternative?

docs/share-action.md
```diff
@@ -0,0 +1,137 @@
...
+    <Annotation Term="Core.OptionalParameter">
+      <Record/>
+    </Annotation>
+  </Parameter>
+</Action>
+```
+
+## Semantics
+
+ShareAll = true, IsDeltaUpdate = true, Users = $[]$ or $[C]$ (the current user)
```
HeikoTheissen, line 40: This list does not explain the effect of the `UserAccessRole` property in `Users`.

docs/share-action.md
martin4all: Parameter should be optional.

vocabularies/Common.xml
```diff
@@ -1125,10 +1125,11 @@ New drafts may also be created by POSTing an entity with property `IsActiveEntit
           <Annotation Term="Core.Description" String="Action that shares a draft document with other users and restricts access to the listed users in their specified roles" />
           <Annotation Term="Core.LongDescription">
             <String>The action is bound to the draft document root node and has the following signature:
-- `Users`: collection of structure with properties
+- `ShareAll` optional of type `Boolean`, `true` means sharing with all users that are authorized for the bound draft instance; the `Users` parameter is optional in this case, provide it to actively notify the listed users
```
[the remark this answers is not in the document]
martin4all: All possible combinations are mentioned. In case `ShareAll `is omitted, `Users ` can be omitted, too. All others are marked as inconsistent and rejected by the backend.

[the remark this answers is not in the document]
HeikoTheissen: > In case `ShareAll `is omitted, `Users ` can be omitted, too.

The vocabulary says otherwise:
> `ShareAll` = `true` means sharing with all users that are authorized for the bound draft instance; the `Users` parameter is optional in this case, provide it to actively notify the listed users

Is this not true?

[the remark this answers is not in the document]
martin4all: `ShareAll = true` switches the draft kind to "ShareAll". `Users` parameter is optional and only needed if you want to change the user list in the backend. In addition user which are added or removed from the list in the backend are notified. 
The conclusion if `ShareAll` is omitted then `Users `can be omitted is not correct. `Users `parameter can always be omitted.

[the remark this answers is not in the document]
HeikoTheissen: > `ShareAll = true` switches the draft kind to "ShareAll". `Users` parameter is optional and only needed if you want to change the user list in the backend.

But the wiki/document calls the combination ShareAll = true, IsDeltaUpdate = true, Users not provided "inconsistent".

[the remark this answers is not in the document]
martin4all: Yes, I thought it is ok to allow the request, but backend rejects it?

martin4all (approved):

PR #392, opened 2025-07-02
Introducing Relative Value List collections

MarcelWaechter:


vocabularies/Common.xml
```diff
@@ -685,6 +685,9 @@ The minimum structure is
         <Property Name="CollectionRoot" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Service root of the value list collection; not specified means local to the document containing the annotation" />
         </Property>
+        <Property Name="RelativeToTarget" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
```
HeikoTheissen, line 688: What's the difference between this and `"CollectionRoot": null`? What does "local to the document containing the annotation" mean?

Suggested alternative: A collection path relative to the annotation target is specified in a new property with type `Edm.NavigationPropertyPath`. `CollectionPath` becomes nullable then.

MarcelWaechter: We also had CollectionRoot: null in mind, but we were not sure if this is really understandable?
Also the idea of having a new property we (Bernhard and I) discussed but then what happens if both are given, so we need to describe it.

Actually, we are fine with all three approaches... I let you decide which one you like at most

HeikoTheissen: `"CollectionRoot": null` is already defined today, why is this not the same as the new desired behavior?

MarcelWaechter: If you say "CollectionRoot": null means the collection path is relative to the annotation target... even better, and we don't need to invent anything new, and will use this one for sure?

MarcelWaechter: Your wish is my command :)
Let me know if you are fine or if you prefer a better description

vocabularies/Common.xml
```diff
@@ -679,9 +679,12 @@ The minimum structure is
           <Annotation Term="Core.IsLanguageDependent" />
           <Annotation Term="Core.Description" String="Headline for value list, fallback is the label of the property or parameter" />
         </Property>
-        <Property Name="CollectionPath" Type="Edm.String" Nullable="false">
+        <Property Name="CollectionPath" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Resource path of an OData collection with possible values, relative to CollectionRoot" />
         </Property>
+        <Property Name="RelativeCollectionPath" Type="Edm.NavigationPropertyPath" Nullable="true">
+          <Annotation Term="Core.Description" String="Navigation property path of an OData collection with possible values, relative to the annotation target" />
+        </Property>
```
HeikoTheissen, line 687: Add a description to this complex type:
> Exactly one of `CollectionPath` and `RelativeCollectionPath` must be provided.

vocabularies/Common.xml
```diff
@@ -679,9 +679,12 @@ The minimum structure is
...
         <Property Name="CollectionRoot" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Service root of the value list collection; not specified means local to the document containing the annotation" />
```
HeikoTheissen, line 689:
```suggestion
          <Annotation Term="Core.Description" String="Service root of the value list collection; not specified means local to the document containing the annotation" />
          <Annotation Term="Core.LongDescription">
            <String>Must only be specified if `CollectionPath` is provided.</String>
          </Annotation>
```

HeikoTheissen (changes requested):

HeikoTheissen (approved):

PR #399, opened 2025-09-23
Align Common.ReferentialConstraint with RAP implementation

HeikoTheissen:
The annotation introduced in #397 can alternatively have an external target, provided that this does not change the interpretation of the property paths. It works if the external annotation is of the form `ns.EntityType/NavigationProperty`.
```xml
<EntityType Name="Warehouse">
  …
  <Property Name="ID" Nullable="false"/>
</EntityType>
<EntityType Name="Product">
  …
  <Property Name="WarehouseID" Nullable="false"/>
  <NavigationProperty Name="_Warehouse" Type="self.Warehouse" Nullable="true"/>
</EntityType>
<Annotations Target="self.Product/_Warehouse">
  <Annotation Term="Common.ReferentialConstraint">
    <Collection>
      <Record>
        <PropertyValue Property="Property" PropertyPath="WarehouseID"/>
        <PropertyValue Property="ReferencedProperty" PropertyPath="_Warehouse/ID"/>
      </Record>
    </Collection>
  </Annotation>
</Annotations>
```


BerSie (approved):
OK

PR #418, opened 2026-01-26
Introducing new term UI.AIGenerated

BerSie:
Based on [BLI](https://jira.tools.sap/browse/FIORITECHP1-31073) there is a need to make the end-user aware that certain values on the UI are generated by AI and has to be treated with special attention.


HeikoTheissen (commented):
Is this a metadata annotation or an instance annotation? In the former case, properties are either always AI-generated or never. But isn't it more likely that a value of a given property is sometimes AI-generated and sometimes not?

And is "AI-generatedness" really a feature of the property? Shouldn't this be an annotation of a `UI.DataFieldAbstract`, like `UI.Hidden` or `UI.Placeholder`?

PR #424, opened 2026-03-03
Adding Auditing annotation

schiwekM:
Adds support for an AuditorScopes annotation to define which roles/scopes have auditing access for an entity set or all entity sets within a service.

```cds
@Auditing.DefaultAuditorScopes : [
  'CATALOG_AUDITOR'
]
service CatalogService {
   entity Customers             as projection on db.Customers;
   @Auditing.AuditorScopes : [
      'CUSTOMER_AUDITOR'
   ]
   entity Orders         as select from db.Orders;
}
```


HeikoTheissen (changes requested):
This sounds like a completely new topic, which deserves its own vocabulary with namespace `com.sap.vocabularies.Auditing.v1`.

schiwekM:
Sounds good - I moved it to its own vocabulary.

vocabularies/Auditing.xml
```diff
@@ -0,0 +1,51 @@
...
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="List of scopes which have by default auditing access for entity sets in the annotated service. Annotating @Auditing.AuditorScopes to an entity set overrides this default." />
+      </Term>
+
+      <Term Name="AuditorScopes" Type="Auditing.AuditorScopesType" Nullable="false" AppliesTo="EntitySet">
+        <Annotation Term="Common.Experimental" />
+        <Annotation Term="Core.Description" String="List of scopes which have auditing access for the annotated entity set. Auditing access could for example mean accessing blocked or archived entities." />
+      </Term>
+
+      <TypeDefinition Name="AuditorScopesType" UnderlyingType="Collection(Edm.String)">
```
HeikoTheissen: That's forbidden, `Collection` cannot appear in a `TypeDefinition`. It can only appear in a `Term` or `Property` definition.

HeikoTheissen (changes requested):
Also add the new vocabulary to the `vocabularies/README.md`.

vocabularies/Auditing.xml
schiwekM: Good to know - I removed the TypeDefinition and use the Collection directly now.

HeikoTheissen (commented):
The vocabulary looks good now, but which OData client will consume this information?

HeikoTheissen (approved):

PR #427, opened 2026-04-28
Introduce Support.DebugInfo

uhlmannm:
Introduce Support.DebugInfo

### DIFF
--- a/vocabularies/Support.xml
+++ b/vocabularies/Support.xml
@@ -32,7 +32,7 @@
       </Annotation>
       <Annotation Term="Common.Experimental" />
 
-      <Term Name="DebugInfo" Type="String" Nullable="false" AppliesTo="Schema">
+      <Term Name="DebugInfo" Type="String" Nullable="false">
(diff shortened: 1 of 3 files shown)


vocabularies/Support.xml
```diff
@@ -32,6 +32,9 @@
       </Annotation>
       <Annotation Term="Common.Experimental" />
 
+      <Term Name="DebugInfo" Type="String" Nullable="false" AppliesTo="Schema">
```
HeikoTheissen:
```suggestion
      <Term Name="DebugInfo" Type="String" Nullable="false">
```
This appears in JSON batch payloads, which matches none of the standard `AppliesTo` values.

HeikoTheissen (changes requested):
The definition (if not the term name) should make clear that this is (for now) used in JSON batch payloads.

HeikoTheissen (commented):
The definition (if not the term name) should make clear that this is (for now) used in JSON batch payloads.

HeikoTheissen (approved):

PR #428, opened 2026-05-15
Separate Created, Updated & Deleted Source Entities for SideEffects

BerSie:
adding specific properties for created and deleted source entities (NavigationPropertyPath). Consumers get the chance to define and react specifically on inserting and removing associated instances. Inistially the proposal contained also an SourceEntityUpdated - I omitted it, as the update is already covered by SourceEntities and I hope that can be handled in a compatible way.


GwendalMoreau:
@gwendalmoreau

MarcelWaechter:
Hi Bernhard,
the idea of also having SourceEntityUpdated was to specify side effects that should only be triggered if a property of this entity was updated, but not deleted or inserted. SourceEntity is triggered in all three cases.
But, I'm not aware of any use case, and I'd say we should be good with just the Inserted and Deleted for now.
I also added Renzo to it, so he and Heiko also might have a look.

Thanks

MarcelWaechter (dismissed):

RaizoC (changes requested):
I would consequently also add SourceEntitiesUpdated as it then also would be consistent to the possible and available trigger conditions in RAP that also offer both "update" as well as "field" aka property. I am not sure if the SourceProperties * is identical but remember some differences between this one and the entity one with regard to navigation properties. If this holds true, SourceEntitiesUpdated would even add functionality, otherwise it at least adds more convenience.

RaizoC (dismissed):

RaizoC (dismissed):

BerSie:
@HeikoTheissen : have you checked this PR?

HeikoTheissen (commented):
Property `SourceEntities` can be set to null to mean the annotated entity. Do we need a similar mechanism for the new properties?

BerSie:
> Property `SourceEntities` can be set to null to mean the annotated entity. Do we need a similar mechanism for the new properties?

Imo this is not necessary - doesn't make sense for create / delete, and for update it's handled by SourceEntities already. But what do the others think, @RaizoC , @MarcelWaechter ?

RaizoC:
> > Property `SourceEntities` can be set to null to mean the annotated entity. Do we need a similar mechanism for the new properties?
> 
> Imo this is not necessary - doesn't make sense for create / delete, and for update it's handled by SourceEntities already. But what do the others think, @RaizoC , @MarcelWaechter ?

I also asked myself the same, but it depends how this is being evaluated. I think it only is evaluated anyhow on the instance itself, so not sure it anyhow would only apply for "update". I also don't think we need it, but we also can add it later on ;-).

RaizoC (approved):

PR #430, opened 2026-05-21
Run actions with Node 24

HeikoTheissen:
Actions running on node 20 are deprecated.

The auto-refresh mechanism is abolished, because `build` is now a required check (like in oasis-tcs/odata-vocabularies).


BerSie (approved):
lgtm

PR #435, opened 2026-06-05
Add AI vocabulary with hint term

vyshnavigadamsetti:
## Summary

Introduces a new `AI` vocabulary (`com.sap.vocabularies.AI.v1`) with a single term `AI.hint`.

`@AI.hint` provides a free-text hint for AI consumers (e.g., large language models or AI agents) on how to use or interpret an annotated OData element. It is intentionally kept separate from human-readable descriptions (`Core.Description`) so that end-user-facing documentation and AI-targeted guidance can evolve independently.

## Cross-format alignment

This is part of a coordinated cross-format rollout of AI-ready metadata across all SAP API formats:

| Format | Annotation |
|---|---|
| OpenAPI | `x-sap-ai-hint` |
| AsyncAPI | `x-sap-ai-hint` |
| JSON Schema | `x-sap-ai-hint` |
| CSN Interop | `@AI.hint` ([PR #169](https://github.com/SAP/csn-interop-specification/pull/169)) |
| OData | `@AI.hint` (this PR) |
| ABAP CDS | `@AI.hint` (CDS Annotation Change Request in progress) |

Reference guideline: https://github.tools.sap/CPA/api-guidelines/pull/361

## Changes

- `vocabularies/AI.xml` — new vocabulary (source of truth)
- `vocabularies/AI.md` — generated
- `vocabularies/AI.json` — generated

PR #440, opened 2026-06-09
`AI.hint` (non-forked version of #435)

HeikoTheissen:
This is a non-forked copy of #435


HeikoTheissen (commented):
Why can't this be expressed as a qualified description?
```xml
<Annotation Term="Core.Description" Qualifier="AI_hint">...</Annotation>
```
I would expect an LLM to read the human-readable descriptions as well, so why not just express this as another description?

vocabularies/AI.xml
```diff
@@ -0,0 +1,65 @@
...
+
+* Business context: what business concept or domain object the element represents
+* Usage guidance: when and how to use this entity, property, or service
+* Disambiguation: how to distinguish from similar entities or operations
+* Value constraints: valid values, formats, coding standards (e.g., ISO codes, enums)
+* Authorization: required scopes or roles for write operations (service level)
+* Routing guidance: when NOT to use this element and what to use instead
+
+The content of AI.hint is not intended for display to end users and MUST be filtered
+when publishing metadata externally (e.g., to public API catalogs).
```
HeikoTheissen, line 56: What's the rationale behind this? And does consumption of the OData service (for example, by a Fiori app) already count as external? If yes, then the `AI.hint` must not be included in the $metadata, because every OData consumer gets this. In other words: Why should `AI.hint` then be an OData annotation at all?

Which consumers shall receive this information? Can this alternatively be published only through an MCP service that wraps the OData service?

Fannon: @HeikoTheissen That's a good point. Only publish this through the MCP Service that wraps OData would be a reasonable option. 

We also have the possibility to distinguish between internal and public metadata, also for OData. But you're right - if you put it directly into the OData metadata, it would be served by $metadata by default. If we would want to keep it internal, some filtering out would have to happen there. We consider similar filtering out of such metadata for other protocols / metadata sources as well. 

One approach is also to add it in an Overlay, patch it in - then the final OData EDMX would need to be able to hold this annotation. 

Also whether this is public or internal is purely a matter of (current) policy - it might change, or depend on case-by-case decisions.

HeikoTheissen: > Only publish this through the MCP Service that wraps OData would be a reasonable option.

How do you rate this option? If you decide for it, we can close this PR.

> One approach is also to add it in an Overlay, patch it in - then the final OData EDMX would need to be able to hold this annotation.

Which component do you want to do the work of "patching in" the additional annotations?

vyshnavigadamsetti:
Closing in favor of #450 which reworks this using the  vocabulary namespace () instead of a separate  vocabulary.

PR #442, opened 2026-06-09
Clean up nodejs.yml workflow

HeikoTheissen:
Removed references to `matrix`, which is not used.
Removed environment variable `CI`, because it is [set automatically](https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/).


BerSie (approved):
lgtm

PR #448, opened 2026-07-09
`$expand=@Common.ServiceDefault`

HeikoTheissen:
SAP-specific extension of the idea described in oasis-tcs/odata-vocabularies#295.


HeikoTheissen:
No longer needed, since [OData V4.02](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#SystemQueryOptionexpand) now allows empty parameters `$select=&$expand=`.

PR #449, opened 2026-07-22
Remove experimental flag from Common.IntervalType/Label

saurbhsingh:
## What

Removes the `Common.Experimental` annotation from the `Label` property of `Common.IntervalType`, making it a stable, non-experimental property.

## Why

The `Label` property has stabilized and should no longer be marked experimental. This follows the same pattern as recent PRs removing experimental flags (e.g. #446).

## Changes

- `Common.xml` — source: removed the `<Annotation Term="Common.Experimental" />` line on `IntervalType/Label`
- `Common.json` / `Common.md` — regenerated via `npm run build`

The large `Common.md` diff is almost entirely `Common.xml#L...` line-number reference shifts caused by removing one line from the source; the only genuine content change is dropping the `(Experimental)` tag on `Label`.

### DIFF
--- a/vocabularies/Common.json
+++ b/vocabularies/Common.json
@@ -473,7 +473,6 @@
       "$Kind": "ComplexType",
       "Label": {
         "$Nullable": true,
-        "@Common.Experimental": true,
         "@Core.Description": "A short, human-readable text suitable for labels and captions in UIs",
         "@Core.IsLanguageDependent": true
       },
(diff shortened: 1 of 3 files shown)


saurbhsingh:
@HeikoTheissen  - The Interval is now consumed by Fiori Elements, so we want to remove the experiment flag from the label. Please do let me know if you need any more information.

HeikoTheissen (approved):

PR #450, opened 2026-07-24
@Consumption.aiHint — AI hint term in Consumption vocabulary

vyshnavigadamsetti:
## Summary

Adds `@Consumption.aiHint` to a new `Consumption` vocabulary, providing a free-text hint for AI consumers (LLMs, agents) on how to use or interpret an annotated OData element.

### Changes

- New vocabulary `com.sap.vocabularies.Consumption.v1` with term `aiHint`
- Term is marked `@Common.Experimental`
- Applies to: EntityType, EntitySet, NavigationProperty, Property, EntityContainer, TypeDefinition

### Cross-format alignment

This aligns the OData annotation name with the other SAP metadata formats:

| Format | Annotation / Property |
|---|---|
| **OData** | `@Consumption.aiHint` ← this PR |
| **CSN Interop / ABAP CDS** | `@Consumption.aiHint` ([SAP/csn-interop-specification#186](https://github.com/SAP/csn-interop-specification/pull/186)) |
| OpenAPI / AsyncAPI / JSON Schema | `x-sap-ai-hint` |

### Replaces

This replaces #440 (`@AI.hint` in a separate AI vocabulary). Using the `Consumption` namespace aligns with the CSN Interop decision (PR #186) and avoids a separate vocabulary just for one term.

The key design argument for a dedicated annotation (vs. a qualified `Core.Description`): AI-targeted guidance and human-readable descriptions must evolve independently, and the annotation value MUST NOT be surfaced to end users. A qualified description does not carry that semantic contract.

### DIFF
--- a/vocabularies/Consumption.json
+++ b/vocabularies/Consumption.json
@@ -1,0 +1,31 @@
+{
+  "$Version": "4.0",
+  "$Reference": {
+    "https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.json": {
+      "$Include": [{ "$Namespace": "Org.OData.Core.V1", "$Alias": "Core" }]
+    },
+    "https://sap.github.io/odata-vocabularies/vocabularies/Common.json": {
+      "$Include": [{ "$Namespace": "com.sap.vocabularies.Common.v1", "$Alias": "Common" }]
+    }
+  },
+  "com.sap.vocabularies.Consumption.v1": {
+    "$Alias": "Consumption",
+    "@Core.Description": "Terms for consumption metadata",
+    "@Core.Description#Published": "2026-07-24 © Copyright 2026 SAP SE. All rights reserved",
+    "@Core.Links": [
+      { "rel": "alternate", "href": "https://sap.github.io/odata-vocabularies/vocabularies/Consumption.xml" },
+      { "rel": "latest-version", "href": "https://sap.github.io/odata-vocabularies/vocabularies/Consumption.json" },
+      {
+        "rel": "describedby",
+        "href": "https://github.com/sap/odata-vocabularies/blob/main/vocabularies/Consumption.md"
+      }
+    ],
+    "aiHint": {
+      "$Kind": "Term",
+      "$AppliesTo": ["EntityType", "EntitySet", "NavigationProperty", "Property", "EntityContainer", "TypeDefinition"],
+      "@Common.Experimental": true,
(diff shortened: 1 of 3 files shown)


HeikoTheissen (changes requested):
Many consumption-related annotations are already in the "Common" vocabulary. I am therefore against a new "Consumption" vocabulary.

Why don't you want the vocabulary name "AI"? Your term is AI-specific.

But, most importantly, I still doubt that we need an OData annotation at all for this (see https://github.com/SAP/odata-vocabularies/pull/440#discussion_r3381006908).

vyshnavigadamsetti:
> Many consumption-related annotations are already in the "Common" vocabulary. I am therefore against a new "Consumption" vocabulary.
> 
> Why don't you want the vocabulary name "AI"? Your term is AI-specific.
> 
> But, most importantly, I still doubt that we need an OData annotation at all for this (see [#440 (comment)](https://github.com/SAP/odata-vocabularies/pull/440#discussion_r3381006908)).

in this case if we already have Consumption as part of of 'Common'. I can extend the "Common Vocabulary" once we reach an agreement, instead of creating a new Consumption vocabulary .
For using a new vocabulary for AI, in the CDS annotation we conculded that for this aiHint it is enough to use the existing Consumption domain and have aiHint as part of it . Hence to keep it consistent we wanted to do similar thing here.

@Fannon  could you please also provide your POV here .

Fannon:
@HeikoTheissen I think we should keep this in sync how ABAP CDS defines it though - otherwise we'd have to translate the annotation names. My impression was that we usually keep this 1:1. 

As in ABAP CDS the Consumption vocabulary is already established, their recommendation was to go with that instead of opening up a completely new AI vocabulary (what was our first suggestion). 

In the end, I would be fine with either of the 3 options, it's more important that we make a decision soon, as the AI topics are moving fast and everything we don't standardize now quickly enough will become messy with interim solutions

Fannon:
@vyshnavigadamsetti - I think the Consumption vocabulary is fine, if this is what we get through ABAP CDS Review quickest, we can pick it. For simplicity I would then strongly be in favor that we don't do it differently for OData

HeikoTheissen:
> I think we should keep this in sync how ABAP CDS defines it though - otherwise we'd have to translate the annotation names. My impression was that we usually keep this 1:1.

No. Every RAP CDS annotation is explicitly mapped to an OData annotation. Nothing gets easier if they use the same namespace.

And you have not addressed my main concern: I think we do not need an OData annotation at all for this, because the information shall _not_ be exposed to all consumers. But it would be if we use the standard CDS-OData translation mechanism.

We had already started discussing this in #440, but the discussion was not continued, and it seems to have been lost when you closed that PR.

PR #453, opened 2026-07-28
AllowedEditingStatusFilters

HeikoTheissen:


saurbhsingh:
There are a bunch of other editing status filters as well. For completeness do we need to consider the others as well?

Screenshot 2026-07-29 at 6 31 52 PM


Or this is specifically for unmanaged queries with draft where only these two options are supported.



Also, I was not part of the discussion but would it make sense to indicate that this is a "unmanaged query with draft" so that the UI can adjust editing status filters and another features that might not be supported?

PR #454, opened 2026-08-03
TechnicalInfoLinks applies to container

HeikoTheissen:
As agreed with @MarcelWaechter and Olaf Tennie.


MarcelWaechter (approved):
Thanks

PR #457, opened 2026-08-26
MultiLevelExpand: Simpler way to express raw data

HeikoTheissen:


vocabularies/Analytics.xml
```diff
@@ -278,7 +277,11 @@ $apply=filter(Industry in ('IT','AI'))
           otherwise they have the same effect as `AdditionalProperties`.</String>
         </Annotation>
         <Property Name="DimensionProperties" Type="Collection(Edm.String)" Nullable="false">
-          <Annotation Term="Core.Description" String="A non-empty list of property names that define a combination of dimension values" />
+          <Annotation Term="Core.Description" String="A list of property names that define a combination of dimension values" />
+          <Annotation Term="Core.LongDescription">
+            <String>An empty list can occur only on the finest-grained level where it means all non-aggregated properties.
```
ThomasChadzelek: "non-aggregated" or "key"?

What about "A property must not be referenced by more than one level." - do we need to say that the empty array means only properties not used so far?

vocabularies/Analytics.xml
```diff
@@ -223,8 +224,6 @@ $apply=filter(Industry in ('IT','AI'))
             the following elements define the property names of consecutively finer-grained aggregation levels.
             The function result is the leveled hierarchy with these levels in preorder,
             entries on the finest-grained level cannot be expanded further.
-            If the finest-grained level includes all key properties, entries on that level additionally contain
```
ThomasChadzelek, line 226: You could say "If the finest-grained level uses an empty list, entries on that level additionally contain all other non-aggregated properties even if they are not referenced in `LevelProperties`."

PetrNovak-SAP:
Q; Do we need the _AdditionalProperties_?
A; The _AdditionalProperties_ parameter is meant for dimension texts and attributes. It is in principle redundant in combination with _$select_, if we implictly add all dimension properties from _$select_ to given group level. However GW does not know property roles and analytical releations. The _AdditionalProperties_ makes the parameterization more self contained. We could ask GW, whether they would be willing to collect the information themselves or retrieve it from us, the service provider...

Q: Is the _Levels=2_ sufficient to control the expand level?
A: It depends, whether the client needs the _UnlimitedDescentantsCount_ or any other information related to the expanded state (e.g. count of leaf level records), with or without the raw data level. I think that we shall be prepared for both.

Q: Shall the _DimensionProperties_ on lower levels avoid properties from higher levels?
A: This requirement is fine just for explicit lists of properties in the _DimensionProperties_ to avoid redundance in the parametrization. But in the result, a group record on certain level must include all dimension properties of higher levels. Otherwise you could not be able to reconstruct the record key. For simplicity I suggest that each level in the result include all properties from higher levels. 

Q: How do we express, which properties shall be returned on the not-grouped level?
A: The _DimensionProperties_, _AdditionalProperties_ and _Aggregation_ is not suitable, e.g. for not groupable measures without custom aggregate, or not groupable paths. If we use _$select_ and _$expand_ like in the "identity scenario", the not-grouped level records in the function result must include all properties of the input set.

ThomasChadzelek:
> implictly add all dimension properties from _$select_ to given group level

That sounds strange to me. I would expect that the leaf level may represent single records and therefore contains all properties in the first place, and then in a next step $select chooses some of them. What you describe sounds kind of the other way 'round?

ThomasChadzelek:
> UnlimitedDescentantsCount

Got it. Currently I see no need for this with UI5's v4.ODataModel

ThomasChadzelek:
> each level in the result include all properties from higher levels

I fully agree. My comment was just meant as nit picking the spec text to make it bullet proof ;-)

ThomasChadzelek:
> If we use _$select_ and _$expand_ like in the "identity scenario", the not-grouped level records in the function result must include all properties of the input set.

I agree (if we are on the single record level).

PetrNovak-SAP:
> > implictly add all dimension properties from _$select_ to given group level
> 
> That sounds strange to me. I would expect that the leaf level may represent single records and therefore contains all properties in the first place, and then in a next step $select chooses some of them. What you describe sounds kind of the other way 'round?

I might have mistunderstood your point about the need to the _AdditionalProperties_. 

What I was thinking about was that this (without _AdditionalProperties_):
_$apply=Analytics.MultiLevelExpand(LevelProperties=[["Country"],["Region"]])&$select=Country,CountryName,Region,RegionName_

... could have implicitly the same result as the following:
_$apply=Analytics.MultiLevelExpand(LevelProperties=[
   {"DimensionProperties":["Country"],"AdditionalProperties":["CountryName"]},
   {"DimensionProperties":["Region"],"AdditionalProperties":["RegionName"]}})_

But, it is not quite strait to implement and understand, what level gives which props.

ThomasChadzelek:
> I might have mistunderstood your point about the need to the _AdditionalProperties_.

My question was "Bräuchte ich Stand heute auf der Einzelsatzebene eigentlich AdditionalProperties oder erübrigen die sich analog zur Diskussion bzgl. identity?" I think it has been clarified above. If the leaf level represents single records and LevelProperties=[] is used (or IncludeRawDataLevel) then this works like `identity` and we can use $select (or even $expand) afterwards.

PetrNovak-SAP:
> > UnlimitedDescentantsCount
> 
> Got it. Currently I see no need for this with UI5's v4.ODataModel

OK. I would expect that too. But I think, soon or later UI/user might wish to know the count of leaf level records per group. And I think I have heared from Matthias, that Markus wished to support, the leaf level on the UI is technically a grouped level. Hence the leaf level must be specified. Theoretically, if BW implement the function, you are allways on a group level and sometimes even, you must technically not read not-grouped data (the fact table level).

ThomasChadzelek (approved):

HeikoTheissen:
OK if we merge this, @PetrNovak-SAP?

PR #459, opened 2026-09-04
`$expand` after `concat`

HeikoTheissen:


ThomasChadzelek:
Do we need the information that there has to be an `identity` transformation somewhere within that `concat`? I guess w/o it, $expand cannot work and $select can only reduce the properties contained in the result in a way which is not helpful.

ThomasChadzelek (approved):

HeikoTheissen:
> Do we need the information that there has to be an `identity` transformation somewhere within that `concat`? I guess w/o it, $expand cannot work and $select can only reduce the properties contained in the result in a way which is not helpful.

What exactly is the new functionality that `Common.ExpandAfterConcatSupported` shall advertise? Is
```
Sales?$apply=concat(
  groupby((Product/ID, Product/Name), aggregate(…)),
  groupby((Customer/ID, Customer/Name), aggregate(…))
)?$expand=Product,Customer
```
- already supported
- newly supported?

